PR #132, opened 2022-02-14
Decode block header only in Index::index_blockfiles

casey:


casey:
A `bitcoin::Block` consists of a `BlockHeader` followed by a `Vec<Transaction>`. Since we don't look at transactions during the initial pass over the blockfiles, this PR changes the code to only deserialize the block header, and not the transaction data.

PR #136, opened 2022-02-17
Test mining and spending transactions in the same block

terror:


tests/find.rs
```diff
@@ -114,3 +114,22 @@ fn index_multiple_blockfiles() -> Result {
     })
     .run()
 }
+#[test]
```
terror, line 117:
```suggestion

#[test]
```

PR #141, opened 2022-02-19
Resume instead of restarting indexing

casey:


casey:
@terror This PR adds a new table that stores which the block hashes we've indexed for each height. This lets us resume where we left off, and also lets us detect reorgs, although I didn't write any code to actually handle reorgs.

I want to add some tests, but wanted to get this code up, since it's useful if you try to index the main chain.

PR #160, opened 2022-02-21
Update to redb 0.0.5

cberner:


casey:
Very nice. I think this is a way better API for reverse iterators. The previous API was a little bit confusing, since you passed in a range from A to B and got back an iterator from B to A.

cberner:
Thanks, ya I've been revising all the APIs to more closely match `std::collections::BtreeMap`

PR #177, opened 2022-03-03
Add parallel LMDB database implementation

casey:
Refactor redb into a separate module, so we can add a parallel lmdb implementation. I don't think there's a strong reason to merge this as is, it's more of a starting point to add an lmdb database implementation. We don't need a trait, we just need to add an `lmdb_database.rs` source file containing `Database` and `WriteTransaction` with the same public functions.


casey:
@terror I refactored things so it should be straightforward to add an lmdb database implementation, just turn on the `lmdb` feature and start implementing the todos in lmdb_database.rs

casey:
I didn't see anything that jumped out at me as being clearly wrong. Do things like `cargo run info` work?

casey:
@terror I figured it out! Lmdb (like redb) only allows one writer at a time. Calling `Environment::create_db` must count as a write, so if there's already a write transaction, `Environment::create_db` will block. The solution here is to create the databases before opening the write transaction.

casey:
I added some more changes which make the empty_index test pass. Putting `open_db` after `begin_ro_txn` failed, and the reverse worked, which surprised me, since I assumed the problem was with doing this with write transactions.

PR #220, opened 2022-06-27
Remove data_hash from metadata

alok:
Nft has that information (implicitly, since it can be derived from nft.data).


casey:
Closing in favor of #221.

PR #229, opened 2022-07-08
Test deploy on vagrant

terror:


casey:
If we fix the test, (which is just breaking due to a new clippy lint) I think this can be merged, since it's an improvement.

PR #252, opened 2022-07-18
Redb bug

casey:
Try to trigger this redb bug on CI


casey:
Seems to be fixed on redb 0.3.0

PR #254, opened 2022-07-18
Clarify duplicate transaction rule in BIP

casey:


casey:
@terror First commit wraps at 79 for readability, second and third are the actual change.

PR #258, opened 2022-07-20
Update dependencies

casey:


casey:
This updates redb to 0.3.0, which fixes the bug I encountered.

Also remove the explicit dependency on secp256k1. It was causing a weird situation where we `cargo update` would fail because it couldn't select dependencies due to conflicting requirements.

terror (approved):

PR #261, opened 2022-07-21
Allow passing ordinals in degree notation

casey:


casey:
This commit is pretty heinous. It allows passing ordinals with degree notation, and the calculations are pretty complicated.

PR #270, opened 2022-07-29
Remove unused functionality

casey:
- The paper wallet functions were for the ordinals workshop, and nobody
  should be using them for anything serious.
- The benchmark was against an old version of ord that read blocks from
  bitcoind's blocks directory, so it doesn't even work.
- There were a bunch of unused dependencies.
- bitcoin provides bitcoin_hashes as a reexport, so we don't need to
  import it ourselves


terror (approved):
LGTM!

PR #292, opened 2022-08-05
List transaction outputs

terror:


casey:
I think this is ready to go. I did a few things:

- Bitcoin datadir needed `Bitcoin` added to it on Linux
- The map vs string issue was because getrawtransactions returns hex when verbose is false
- Update to redb 0.5.0, which automatically aborts transactions on drop, so we won't get that leaked write transaction warning (see https://github.com/cberner/redb/issues/320)

casey:
I added and then removed a wrapper, AbortOnDrop, but post 0.5.0 we don't need it so I removed it.

PR #301, opened 2022-08-10
Use Bitcoin Core from Homebrew on CI

casey:


casey:
Also run bin/forbid on every platform, since we can get ripgrep from homebrew.

casey:
Looks like the GitHub runners get Homebrew updates on a delay. This will probably work in a day or two.

casey:
I actually think this is a bad idea now, it takes 1m30s to install bitcoind via homebrew, which is much slower than getting it from bitcoincore.org.

PR #343, opened 2022-08-18
Use an ordered list for homepage

casey:


casey:
@terror Been wanting to do this for a while. Makes the markup clearer, looks a little better, and is probably better for screen readers.

PR #344, opened 2022-08-18
Remove GitHub pages directory

casey:
Now that we're serving everything from the block explorer, we can nuke the original sources.


casey:
The deck isn't online, but I have the source for that elsewhere, and it's not super useful anyways.

terror (approved):

PR #376, opened 2022-08-25
Only bin/graph current sync

casey:


justfile
```diff
@@ -87,5 +87,9 @@ update-modern-normalize:
     https://raw.githubusercontent.com/sindresorhus/modern-normalize/main/modern-normalize.css \
     > static/modern-normalize.css
 
-graph:
-  ./bin/graph
+download-log host="ordinals.com":
+  ssh root@{{host}} 'journalctl -u ord > ord.log'
+  rsync --progress root@{{host}}:ord.log ord.log
+
+graph host="ordinals.com":
```
terror, line 94: Shouldn't this depend on download-log ?

casey: It could, although download-log is actually pretty slow. It takes around 3 minutes to download the current log from ordinals.com, so you probably don't want to re-download it every time you view the graph.

terror: Ah okay. Other than that does `host` need to be passed in to `graph`? It isn't used anywhere currently

casey: Good catch! Removed it.

PR #380, opened 2022-08-30
Guard against invalid percentiles

casey:


casey:
@terror Check this out when you have the chance. I wasn't validating percentiles, so this fixes that. Wanted to get it in pronto so I could deploy before bitdevs.

Issue #387, opened 2022-09-01
Add titles to clock hands


casey:
SVG supports a `<title>` element, which looks like it shows tooltips on hover:

https://stackoverflow.com/questions/10643426/how-to-add-a-tooltip-to-an-svg-graphic

It would be cool to label the different parts of the clock with titles, so users can hover over them and see what they mean.

A good first step would be to add titles to the chapter ring marks, so when you hover over them, they say "Each mark on the chapter ring is a halving."

Issue #386, opened 2022-09-01
Show block height on clock


casey:
The clock at ordinals.com/clock is based on the block height, so I think I it should include the block height, as text, on the face, either above or below the middle.

PR #447, opened 2022-09-01
added block height to ordinal clock

raphjaph:
This starts work on #386 and #387.

The SVG had to be scaled up (2x for now) because the smallest readable font size is `10px` even with `font-weight="1"`. I haven't made any choices regarding `font-family` but it should probably be some sort of `sans-serif` or `monospace`.

An idea/proposal on how to do <title> explanations. Added some css so people hover over the marker long enough for the explanation to pop up. 

Is is called Halving or Halvening?

### DIFF
--- a/src/subcommand/server/templates/clock.rs
+++ b/src/subcommand/server/templates/clock.rs
@@ -2,6 +2,7 @@ use super::*;
 
 #[derive(Display)]
 pub(crate) struct ClockSvg {
+  height: u64,
   hour: f64,
   minute: f64,
   second: f64,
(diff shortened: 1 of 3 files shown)


casey:
Want to split this into two PRs? That'll make them easy to review and merge independently. One for block height, and one for tooltips? I like the idea of elements changing color when they're hovered over.

I think for the font, `sans-serif` is a good first start, on the small side, I was thinking in light grey.

I'm not sure I understand the font size issue. Can't the font size just be doubled or halved, instead of scaling the SVG?

raphjaph:
ok I'll close this PR and open two new ones on feature branches

PR #475, opened 2022-09-04
Remove nix cruft

jurraca:
So, I had added a flake on the fly for development and building using Nix, and haven't kept up with the project as I should have, and it doesn't use an explicit Rust version, nor does it build correctly. After playing with this, a few reasons to change the current setup: 
- flakes are experimental, and probably don't have much of an audience
- no need to make building more complicated than just using cargo, or supporting multiple build methods
- a `shell.nix` is sufficient because it addresses all Nix users irrespective of version/setup: just `nix-shell` in the project directory and you're set up for development, which is the whole point. 
- in this case it uses rust 1.63 as required via mozilla's [rust overlays](https://github.com/mozilla/nixpkgs-mozilla#using-in-nix-expressions), which is the most foolproof way to use Rust with Nix. 
Happy to nuke the Nix stuff if you think it's annoying, but the only maintenance here will be bumping rust versions as needed.


casey:
Nice! I don't think I understand what a `shell.nix` is. Is it just supposed to provide all the dependencies needed to work on a project?

Some questions:

- Is it possible to use stable rust, and not require any particular version of rust? This seems better than having to bump the version manually, and `just` should always build on stable rust. Also, is it possible to include stable rust in a way which is the same as a NixOS installation will get it, so that it doesn't require an additional download?
- If it's not a good idea to use whatever's stable, can we refer to the rust by version, i.e. 1.63 instead of "stable" with a date?
- Is gcc required? Or is llvm/clang good enough?
- What's the convention for what to call these things? moz_overlay_shell is confusing, maybe `just-shell`, `just-overlay`, or `just-overlay-shell`?
- Users might not want logging on, so I'd remove `RUST_LOG`

jurraca:
> Is it just supposed to provide all the dependencies needed to work on a project?
yep, exactly
> gcc
sorry, old habits die hard
> What's the convention for what to call these things?
yea, `just-shell` is fine, or `ord shell`, it's really just metadata, you won't be calling the shell by name. However, I did add `just` to the shell since it's available in `nixpkgs`. 
> `just` should always build on stable rust
trust you here obvs, but I don't follow what just has to do with building, I thought it just ran commands?

casey:
> yea, just-shell is fine, or ord shell, it's really just metadata, you won't be calling the shell by name. However, I did add just to the shell since it's available in nixpkgs.

Whoops I'm silly, I was confused and thought this issue was in github.com/casey/just 😅

shell.nix
```diff
@@ -0,0 +1,16 @@
...
+    name = "ord-shell";
+    buildInputs = [
+      # to use the latest nightly:
+      nixpkgs.latest.rustChannels.stable.rust
+      clang
+      just
+      # to use the project's rust-toolchain file:
+      # (nixpkgs.rustChannelOf { rustToolchain = ./rust-toolchain; }).rust
+    ];
+  }
```
casey, line 16: Is clang needed? I assume it is, probably due to a dependency somewhere, but wanted to make sure.

How about removing the comment about latest nightly, and also about the rust-toolchain file, since we don't have one?

```suggestion
let
    moz_overlay = import (builtins.fetchTarball https://github.com/mozilla/nixpkgs-mozilla/archive/master.tar.gz);
  nixpkgs = import <nixpkgs> { overlays = [ moz_overlay ]; };
in
  with nixpkgs;
  stdenv.mkDerivation {
    name = "ord-shell";
    buildInputs = [
      nixpkgs.latest.rustChannels.stable.rust
      clang
      just
    ];
  }
```

jurraca: thanks, that was sloppy. i'm a little unclear on what rust depends on, but yea it will get pulled in via the rust dep.

casey:
Sweet, looks good. Is this ready to merge?

jurraca:
yep! thanks for the feedback.

casey:
Awesome, this is great! I really like how simple it makes getting started.

PR #476, opened 2022-09-05
added tooltips to clock

raphjaph:
Let me know what you think of the text.

When hovering the ring marks change color to the type of ordinal they represent.


raphjaph:
I called Difficulty Adjustment + Halving a Conjunction in the SVG but it should probably be called a Cycle (i.e. 1st Cycle, 2nd Cycle, ...). This keeps the naming consistent with the degree notation.

raphjaph:
I know you don't like the <style></style> stuff on the bottom but there is no way to add the color change on :hover without it.

PR #478, opened 2022-09-05
Invalid degree bugfix

casey:


src/ordinal.rs
```diff
@@ -107,9 +107,17 @@ impl Ordinal {
 
     let cycle_start_epoch = cycle_number * CYCLE_EPOCHS;
 
-    let cycle_progression = period_offset
```
raphjaph, line 110: `cycle_progression` should be a number between 0 and 1_260_00 (6*210_00) that represents the amount of blocks passed in this cycle.

src/ordinal.rs
```diff
@@ -107,9 +107,17 @@ impl Ordinal {
...
-      .checked_sub(epoch_offset % PERIOD_BLOCKS)
```
raphjaph, line 111: `epoch_offset` is a number betwenn 0 and 210_000 that represents number of blocks mined since last Halving.

`epoch_offset % PERIOD_BLOCKS` gives us the `period_offset`, but only if we are at the beginning of a cycle (in the first of six halving epochs).

The correct period_offset:
`period_offset = cycle_progression % PERIOD_BLOCKS`

I'm a little bit confused now and will look at this tomorrow.

src/ordinal.rs
```diff
@@ -107,9 +107,17 @@ impl Ordinal {
...
-      .ok_or_else(|| anyhow!("Invalid relationship between epoch offset and period offset"))?;
```
raphjaph, line 112: I think you wanted to test that the absolute difference between the `epoch_offset` and the `period_offset` was always a multiple of `336` (`Epoch::BLOCKS % PERIOD_BLOCKS`).

Valid: `0°0'336''1234'''`
Valid: `0°1680'0''1234'''`
Invalid: `0°1680'1''1234'''`
Invalid:  `0°1'672''1234'''`

src/ordinal.rs
```diff
@@ -107,9 +107,17 @@ impl Ordinal {
...
+    //
+    // epoch_offset % PERIOD_BLOCKS can be 2015
+    //
+    // - What is the earliest ordinal that has this problem?
+    // - Why did I assume that this checked_sub would always succeed?
+    let cycle_progression =
+      (((period_offset as i64) - ((epoch_offset % PERIOD_BLOCKS) as i64)).abs()) as u64;
 
     if cycle_progression % (Epoch::BLOCKS % PERIOD_BLOCKS) != 0 {
       bail!("Invalid relationship between epoch offset and period offset");
```
raphjaph, line 123: Not sure what `cycle_progression` signifies exactly. (number of block in current cycle?)

casey:
Fixed in #485!

PR #492, opened 2022-09-08
Clock on homepage

raphjaph:
#407


raphjaph:
alread linking to clock in navbar

PR #497, opened 2022-09-08
Improve benchmark

casey:


casey:
cc: @raphjaph @VirtuallyThere

Merging in some improvements to the benchmark. I left `--height-limit` at 5000 to avoid hitting the timeout on macos.

Issue #500, opened 2022-09-08
Collect more timing information


casey:
Indexing time is divided into:
- Time spent writing to the database
- Time committing writes to the database

We should print timing information that shows how long we spent between commits (not including commits) and how long we spent actually committing.

casey:
This is better covered by a benchmarking issue.

PR #523, opened 2022-09-12
Collect and log more timing information (issue #500)

VirtuallyThere:
Collect more timing info as per #500 

Collect and log:
- time elapsed for database commit
- time elapsed for updating (=indexing to) database/btree in between commits

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -137,16 +137,7 @@ impl Index {
       let done = self.index_block(&mut wtx)?;
 
       if block % 1000 == 0 {
-        let commit_start_time = Instant::now();
-        wtx.commit()?;
-        let now = Instant::now();
-        log::info!(
-          "{}ms elapsed since previous commit, committed up to block {} in {}ms",
-          (now - last_commit_time).as_millis(),
-          self.height()?.n(),
-          (now - commit_start_time).as_millis()
-        );
-        last_commit_time = Instant::now();
+        last_commit_time = Self::commit_and_log(wtx, last_commit_time, self.height()?.n()).unwrap();
         wtx = self.database.begin_write()?;
       }
 
@@ -157,17 +148,26 @@ impl Index {
       block += 1;
     }
 
+    Self::commit_and_log(wtx, last_commit_time, self.height()?.n()).unwrap();
+
+    Ok(())
+  }
+
+  fn commit_and_log(
+    wtx: WriteTransaction,
+    last_commit_time: Instant,
+    height: u64,
+  ) -> Result<Instant> {
     let commit_start_time = Instant::now();
     wtx.commit()?;
     let now = Instant::now();
     log::info!(
-      "{}ms elapsed since previous commit, final commit up to block {} in {}ms",
-      (now - last_commit_time).as_millis(),
-      self.height()?.n(),
+      "{}ms elapsed since previous commit, committed up to block {} in {}ms",
+      (commit_start_time - last_commit_time).as_millis(),
+      height,
       (now - commit_start_time).as_millis()
     );
-
-    Ok(())
+    Ok(Instant::now())
   }
 
   pub(crate) fn index_block(&self, wtx: &mut WriteTransaction) -> Result<bool> {
--- a/src/index.rs
+++ b/src/index.rs
@@ -124,6 +124,7 @@ impl Index {
 
   pub(crate) fn index_ranges(&self) -> Result {
     let mut block = 0;
+    let mut last_commit_time = Instant::now();
     let mut wtx = self.database.begin_write()?;
 
     loop {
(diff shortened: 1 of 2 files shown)


src/index.rs
```diff
@@ -136,7 +137,16 @@ impl Index {
       let done = self.index_block(&mut wtx)?;
 
       if block % 1000 == 0 {
+        let commit_start_time = Instant::now();
         wtx.commit()?;
+        let now = Instant::now();
+        log::info!(
+          "{}ms elapsed since previous commit, committed up to block {} in {}ms",
+          (now - last_commit_time).as_millis(),
```
casey, line 145: To make the two times not overlap, should this be:

```suggestion
          (commit_start_time - last_commit_time).as_millis(),
```

casey:
Since the logging and time calculation is somewhat complex, how about refactoring it into a method that we call twice, instead of duplicating:

```
fn commit(wtx: WriteTransaction, last_commit: Instant, height: u64) -> Instant {
  todo!()
}
```

src/index.rs
VirtuallyThere: Yeah, if we don't want them to overlap that's right - I was thinking of the elapsed time reported to include the block commit time. Probably cleaner to have non-overlapping ranges.

VirtuallyThere:
> fn commit(wtx: WriteTransaction, last_commit: Instant, height: u64) -> Instant {
>   todo!()
> }

I had tried that bug struggled with passing the WriteTransaction around. Will try once more.

VirtuallyThere:
I extracted a method as suggested. Not sure the way I handle the potential errors from commit() is OK like this. Realize that extracting methods gets tricky with the error handling approach.

> Since the logging and time calculation is somewhat complex, how about refactoring it into a method that we call twice, instead of duplicating:
> 
> ```
> fn commit(wtx: WriteTransaction, last_commit: Instant, height: u64) -> Instant {
>   todo!()
> }
> ```

src/index.rs
```diff
@@ -136,7 +137,7 @@ impl Index {
       let done = self.index_block(&mut wtx)?;
 
       if block % 1000 == 0 {
-        wtx.commit()?;
+        last_commit_time = Self::commit_and_log(wtx, last_commit_time, self.height()?.n()).unwrap();
```
casey, line 140: Can you use `?` for error handling? Also thinking that `last_commit_time` can be taking by `&mut` so it can just be updated in-place:

```suggestion
    Self::commit_and_log(wtx, &mut last_commit_time, self.height()?.n())?;
```

casey, line 140: Can you use `?` for error handling? Also thinking that `last_commit_time` can be taking by `&mut` so it can just be updated in-place:

```suggestion
    Self::commit_and_log(wtx, &mut last_commit_time, self.height()?.n())?;
```

src/index.rs
```diff
@@ -147,11 +148,28 @@ impl Index {
       block += 1;
     }
 
-    wtx.commit()?;
+    Self::commit_and_log(wtx, last_commit_time, self.height()?.n()).unwrap();
```
casey, line 151:
```suggestion
    Self::commit_and_log(wtx, &mut last_commit_time, self.height()?.n())?;
```

src/index.rs
```diff
@@ -147,11 +148,28 @@ impl Index {
...
 
     Ok(())
   }
 
+  fn commit_and_log(
+    wtx: WriteTransaction,
+    last_commit_time: Instant,
```
casey, line 158:
```suggestion
    last_commit_time: &mut Instant,
```

src/index.rs
```diff
@@ -147,11 +148,28 @@ impl Index {
...
+    let commit_start_time = Instant::now();
+    wtx.commit()?;
+    let now = Instant::now();
+    log::info!(
+      "{}ms elapsed since previous commit, committed up to block {} in {}ms",
+      (commit_start_time - last_commit_time).as_millis(),
+      height,
+      (now - commit_start_time).as_millis()
+    );
+    Ok(Instant::now())
```
casey, line 170:
```suggestion
    *last_commit_time = Instant::now();
    Ok(())
```

src/index.rs
```diff
@@ -147,11 +148,28 @@ impl Index {
...
+    Self::commit_and_log(wtx, last_commit_time, self.height()?.n()).unwrap();
 
     Ok(())
   }
 
+  fn commit_and_log(
+    wtx: WriteTransaction,
+    last_commit_time: Instant,
+    height: u64,
+  ) -> Result<Instant> {
```
casey, line 160:
```suggestion
  ) -> Result {
```

src/subcommand/list.rs
```diff
@@ -2,22 +2,36 @@ use super::*;
 
 #[derive(Debug, Parser)]
 pub(crate) struct List {
-  outpoint: OutPoint,
+  #[clap(long, short, help = "Use extended output format")]
```
casey, line 5: I think the extended output format code snuck in here.

casey (changes requested):
Added some comments! I think you can use `?` for error handling.

src/subcommand/list.rs
VirtuallyThere: Ups, yes. I must have messed something up after I had merged those changes into my master.

Issue #487, opened 2022-09-07
Keep statistics on how many outputs we've seen


casey:
We have an `ord info` command that prints out out statistics about the database. One metric that would be good to track is how many outputs were processed in total. We should keep track of this while we're building the index, so we can print it out with `ord info`.

casey:
Done!

PR #525, opened 2022-09-12
Store and count the total number of outputs traversed when building the index

raphjaph:
I'm unsure about how to properly store a single value persistently in the database. In its own table? What should the key be? Or I make a table HEIGHT_TO_NUM_OUTPUTS that tracks how many outputs there are per block and then I just add up those values in the `info` command?

#487

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -13,6 +13,8 @@ const OUTPOINT_TO_ORDINAL_RANGES: TableDefinition<[u8; 36], [u8]> =
   TableDefinition::new("OUTPOINT_TO_ORDINAL_RANGES");
 const OUTPOINT_TO_TXID: TableDefinition<[u8; 36], [u8; 32]> =
   TableDefinition::new("OUTPOINT_TO_TXID");
+// Don't know if we need a whole table, cause I just want to store one value persistently.
+const OUTPUTS_TRAVERSED: TableDefinition<u8, u64> = TableDefinition::new("TOTAL_OUTPUTS_TRAVERSED");
 
 pub(crate) struct Index {
   client: Client,
@@ -56,6 +58,7 @@ impl Index {
     tx.open_table(HEIGHT_TO_HASH)?;
     tx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?;
     tx.open_table(OUTPOINT_TO_TXID)?;
+    tx.open_table(OUTPUTS_TRAVERSED)?; // set default value if empty? This looks really wrong
 
     tx.commit()?;
 
@@ -88,11 +91,13 @@ impl Index {
       .unwrap_or(0);
 
     let outputs_indexed = wtx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?.len()?;
-
+    // Double unwrap, let's go !!
+    let outputs_traversed = wtx.open_table(OUTPUTS_TRAVERSED)?.get(&0).unwrap().unwrap();
     let stats = wtx.stats()?;
 
     println!("blocks indexed: {}", blocks_indexed);
     println!("outputs indexed: {}", outputs_indexed);
+    println!("outputs traversed: {}", outputs_traversed);
     println!("tree height: {}", stats.tree_height());
     println!("free pages: {}", stats.free_pages());
     println!("stored: {}", Bytes(stats.stored_bytes()));
@@ -156,9 +161,11 @@ impl Index {
     let mut height_to_hash = wtx.open_table(HEIGHT_TO_HASH)?;
     let mut outpoint_to_ordinal_ranges = wtx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?;
     let mut outpoint_to_txid = wtx.open_table(OUTPOINT_TO_TXID)?;
+    let mut outputs_traversed = wtx.open_table(OUTPUTS_TRAVERSED)?;
 
     let start = Instant::now();
     let mut ordinal_ranges_written = 0;
+    let mut outputs_in_block = 0;
 
     let height = height_to_hash
       .range(0..)?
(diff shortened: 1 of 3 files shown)


raphjaph:
If you search for an outpoint that has been spent you get the following error message on the website
`https://ordinals.com/output/969d68907f0e2aeacb8b17979fa48ea3b817cc5bfc92b1675326b3c170991c88:0`
`Output unknown.`
Which makes sense. So the final product is an index of UTXOs not necessarily outpoints.

src/index.rs
```diff
@@ -13,6 +13,7 @@ const OUTPOINT_TO_ORDINAL_RANGES: TableDefinition<[u8; 36], [u8]> =
   TableDefinition::new("OUTPOINT_TO_ORDINAL_RANGES");
 const OUTPOINT_TO_TXID: TableDefinition<[u8; 36], [u8; 32]> =
   TableDefinition::new("OUTPOINT_TO_TXID");
+const OUTPUTS_TRAVERSED: TableDefinition<str, u64> = TableDefinition::new("OUTPUTS_TRAVERSED");
```
casey, line 16: Let's make it a general table for statistics, so we can easily use it for other things:

```suggestion
const STATISTICS: TableDefinition<u64, u64> = TableDefinition::new("STATISTICS");
```

We can create an enum for statistics:

```rust
#[repr(u64)]
enum Statistics {
  TotalOutputs = 0,
}
```

Enums with `repr(T)` can be converted to `T` with `as`: `Statistics::TotalOutputs as u64`, so we can use a `Statistics` value as a key.

src/index.rs
```diff
@@ -56,6 +57,7 @@ impl Index {
     tx.open_table(HEIGHT_TO_HASH)?;
     tx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?;
     tx.open_table(OUTPOINT_TO_TXID)?;
+    tx.open_table(OUTPUTS_TRAVERSED)?;
```
casey, line 60:
```suggestion
    tx.open_table(STATISTICS)?;
```

src/index.rs
```diff
@@ -87,12 +89,18 @@ impl Index {
       .map(|(height, _hash)| height + 1)
       .unwrap_or(0);
 
-    let outputs_indexed = wtx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?.len()?;
+    let utxos_indexed = wtx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?.len()?;
+
+    let outputs_traversed = wtx
+      .open_table(OUTPUTS_TRAVERSED)?
+      .get("outputs_traversed")?
+      .unwrap_or(0);
```
casey, line 97: I think this `.into()` will work, not sure:

```suggestion
    let outputs_traversed = wtx
      .open_table(STATISTICS)?
      .get(Statistics::OutputsTraversed.into())?
      .unwrap_or(0);
```

src/index.rs
```diff
@@ -156,9 +164,11 @@ impl Index {
     let mut height_to_hash = wtx.open_table(HEIGHT_TO_HASH)?;
     let mut outpoint_to_ordinal_ranges = wtx.open_table(OUTPOINT_TO_ORDINAL_RANGES)?;
     let mut outpoint_to_txid = wtx.open_table(OUTPOINT_TO_TXID)?;
+    let mut outputs_traversed = wtx.open_table(OUTPUTS_TRAVERSED)?;
```
casey, line 167:
```suggestion
    let mut statistics = wtx.open_table(STATISTICS)?;
```

src/index.rs
```diff
@@ -261,11 +272,20 @@ impl Index {
...
+        &mut outputs_in_block,
       )?;
     }
 
     height_to_hash.insert(&height, &block.block_hash())?;
 
+    // TODO: how to set default differently?
+    if outputs_traversed.is_empty()? {
+      outputs_traversed.insert("outputs_traversed", &0)?;
+    }
```
casey, line 284: I think we never need to set a default, as long as when we get it, we use zero if a value isn't present.

```suggestion
```

src/index.rs
```diff
@@ -261,11 +272,20 @@ impl Index {
...
+    // TODO: error handling
+    let sum = outputs_in_block + outputs_traversed.get("outputs_traversed")?.unwrap();
+    outputs_traversed.insert("outputs_traversed", &sum)?;
```
casey, line 287: Maybe something like this:

```suggestion
    statistics.insert(Statistics::OutputsTraversed.into(), statistics.get(Statistics::OutputsTraversed.into())?.unwrap_or(0) + outputs_in_block)?;    
```

There might be borrow errors, if so, then you can break it into two expressions like you had.

casey (changes requested):
Mostly looks good! We should create a general statistics table, with a dedicated enum as the key.

src/index.rs
raphjaph: nice, that's very handy

src/index.rs
raphjaph: Did it like this because `From<fn(u64) -> Statistics {Statistics::OutputsTraversed}>` (and therefore `into()`) is not implemented for `u64`
Can/should I implement it above somewhere?
```rust
    statistics.insert(
      &(Statistics::OutputsTraversed as u64),
      &(statistics
        .get(&(Statistics::OutputsTraversed as u64))?
        .unwrap_or(0)
        + outputs_in_block),
    )?;
```

casey: Gotcha. I think the right thing to do is to implement `From` like so:

```rust
impl From<Statistic> for u64 {
    fn from(statistic: Statistic) -> Self {
        statistic as u64
    }
}
```

(And I guess the name should be `Statistic`.)

The reason to avoid `as` is that it's error prone. Even though `Statistics` is declared as `repr(u64)`, `as u8` will work, and will silently truncate. So it's good to avoid it and use `.into()` instead.

raphjaph: Ok, good to know

casey (approved):

PR #526, opened 2022-09-13
Fix off-by-some --height-limit bug

### LINKED ISSUE #517: Off by some bug in `-height-limit`
When running
```
$ ord --height-limit 451000 index
$ ord info
```
on an index that contained 450006 blocks before the run, the output is
```
blocks indexed: 451008
outputs indexed: 46702642
tree height: 5
free pages: 66202117
stored: 6.82 GiB
overhead: 396.57 MiB
fragmented: 6.61 GiB
index size: 32 GiB
```

There seems to be some bug that results in an increasing deviation from the specified height limit with indices that are built over consecutive runs with increasing height limits.

casey:
Fixes #517. We were reading the height using `self.height()`, which opens a new read transaction, which will return the old height, until the current write transaction is committed.

Instead, read the height from the open write transaction, and increment it every time we index a block.

This should be ready to merge once I add a test.

todo:
- [x] remove transactions from dummy blocks?
- [x] #[rpc] -> #[jsonrpc_derive::rpc]
- [x] consider parsing options directly
- [x] do we need Default impls?
- [x] fix localhost address
- [x] convert height_limit to u64
- [x] convert --height-limit integration test into unit test
- [x] can i get rid of fallible conversion?
- [x] make height stuff coherent
- [x] reorder `BicoinRpcServerHandle`


src/test.rs
```diff
@@ -14,3 +22,118 @@ macro_rules! assert_regex_match {
...
+    match self.blocks.get(&block_hash) {
+      Some(block) => Ok(hex::encode(bitcoin::consensus::encode::serialize(block))),
+      None => Err(jsonrpc_core::Error::new(
+        jsonrpc_core::types::error::ErrorCode::ServerError(-8),
+      )),
+    }
+  }
+}
+
+pub(crate) struct BitcoinRpcServerHandle {
```
raphjaph, line 130: struct should appear before impl

tests/state.rs
```diff
@@ -264,7 +264,7 @@ impl State {
       .parse::<u64>()
       .unwrap();
 
-      if ord_height == bitcoind_height {
+      if ord_height == bitcoind_height + 1 {
```
raphjaph, line 267: be gone +1

raphjaph, line 267: be gone +1

raphjaph, line 267:
```suggestion
      if ord_height == bitcoind_height {
```

PR #545, opened 2022-09-21
Update index in `Index::open`

casey:
It's never useful to have an unsynced index, so do it on open, and rename `Index::index_ranges` to just `Index::index`.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -79,19 +79,14 @@ impl Index {
 
     tx.commit()?;
 
-    Ok(Self {
+    let index = Self {
       client,
       database,
       database_path,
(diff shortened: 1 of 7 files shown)


casey:
@raphjaph This is little refactor of the index. It removes the `Index::index` static method which opens and syncs an index in favor of always syncing the index in `Index::open`, and renames `Index::index_ranges` to `Index::index`, since `Index::index_ranges` sounds more like it's something specialized. Take a look!

raphjaph (approved):
yeah, this makes sense.  We should probably merge this and rebase #544 on top.

PR #560, opened 2022-09-26
Convert even more unit tests

raphjaph:
added some methods and RPC endpoints to the test `BitcoinRpcServer` so that we have a mempool and a transaction index

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -1088,7 +1088,7 @@ mod tests {
 
   #[test]
   fn clock_is_served_with_svg_extension() {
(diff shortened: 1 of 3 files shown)


src/subcommand/server.rs
```diff
@@ -1090,4 +1090,44 @@ mod tests {
   fn clock_is_served_with_svg_extension() {
     TestServer::new().assert_response_regex("clock.svg", StatusCode::OK, "<svg.*");
```
casey, line 1091:
```suggestion
    TestServer::new().assert_response_regex("/clock.svg", StatusCode::OK, "<svg.*");
```

PR #571, opened 2022-09-27
Convert some find tests

casey:


casey:
I'm not sure if the tests pass, but this is ready for review otherwise. I added a `Context` struct (we could also call it `TestContext` if that's clearer) for quickly setting up an `Index` for unit tests. You might want to rebase your changes on this branch, so you can use it, since it makes the tests much easier to write.

raphjaph (approved):

PR #586, opened 2022-09-30
add identify command to wallet

raphjaph:


src/subcommand/wallet.rs
```diff
@@ -0,0 +1,24 @@
...
+  super::*,
+  bitcoincore_rpc::{Auth, Client, RpcApi},
+};
+
+mod identify;
+
+#[derive(Debug, Parser)]
+pub(crate) enum Wallet {
+  Identify,
+  List,
```
casey: This can be removed since it's not implemented yet.

src/subcommand/wallet.rs
```diff
@@ -0,0 +1,24 @@
+use {
+  super::*,
+  bitcoincore_rpc::{Auth, Client, RpcApi},
```
casey, line 3: I'd move these into the `Identify` module.

src/subcommand/wallet/identify.rs
```diff
@@ -0,0 +1,30 @@
+use super::*;
+
+pub(crate) fn run(options: Options) -> Result {
+  let index = Index::open(&options)?;
+  index.index()?;
+
+  let client = Client::new(&options.rpc_url(), Auth::CookieFile(options.cookie_file()?))
```
casey, line 7: I would duplicate log message from Index::open here, since it's very useful.

src/subcommand/wallet/identify.rs
```diff
@@ -0,0 +1,30 @@
...
+    .context("Failed to connect to RPC URL")?;
+
+  let utxos = client.list_unspent(None, None, None, None, None).unwrap();
+  for utxo in utxos {
```
casey, line 11:
```suggestion
  for utxo in client.list_unspent(None, None, None, None, None)? {
```

src/subcommand/wallet/identify.rs
```diff
@@ -0,0 +1,30 @@
...
+    let output = OutPoint::new(utxo.txid, utxo.vout);
+    match index.list(output).unwrap() {
```
casey, line 13:
```suggestion
    match index.list(output)? {
```

src/subcommand/wallet/identify.rs
```diff
@@ -0,0 +1,30 @@
...
+      Some(List::Unspent(ordinal_ranges)) => {
+        for (offset, ordinal_range) in ordinal_ranges.iter().enumerate() {
+          let ordinal = Ordinal(ordinal_range.0);
+          let rarity = ordinal.rarity();
+          match rarity {
+            Rarity::Common => (),
+            _ => println!("{ordinal}\t{output}\t{offset}\t{rarity}"),
+          }
```
casey, line 21:
```suggestion
        let mut offset = 0;
        for range in &ordinal_ranges {
          let ordinal = Ordinal(range.0);
          let rarity = ordinal.rarity();
          if rarity > Rarity::Common {
            println!("{ordinal}\t{output}\t{offset}\t{rarity}");
          }
         offset += range.1 - range.0
```

src/subcommand/wallet/identify.rs
```diff
@@ -0,0 +1,30 @@
...
+        }
+      }
+      Some(_) => (),
+      None => (),
```
casey, line 25: These are error cases and must be handled. (They indicate that wallet sees UTXO but index does not.)

test-bitcoincore-rpc/src/lib.rs
```diff
@@ -204,6 +206,16 @@ pub trait Api {
...
     blockhash: Option<BlockHash>,
   ) -> Result<String, jsonrpc_core::Error>;
+
+  #[rpc(name = "listunspent")]
+  fn list_unspent(
+    &self,
+    minconf: Option<usize>,
+    maxconf: Option<usize>,
+    address: Option<bitcoin::Address>,
+    include_unsage: Option<bool>,
```
casey, line 216:
```suggestion
    include_unsafe: Option<bool>,
```

test-bitcoincore-rpc/src/lib.rs
```diff
@@ -255,6 +267,52 @@ impl Api for Server {
...
+    maxconf: Option<usize>,
+    address: Option<bitcoin::Address>,
+    include_unsafe: Option<bool>,
+    query_options: Option<String>,
+  ) -> Result<Vec<ListUnspentResultEntry>, jsonrpc_core::Error> {
+    assert_eq!(minconf, None, "");
+    assert_eq!(maxconf, None, "");
+    assert_eq!(address, None, "");
+    assert_eq!(include_unsafe, None, "");
+    assert_eq!(query_options, None, "");
```
casey, line 283: Add error messages.

tests/wallet.rs
```diff
@@ -0,0 +1,19 @@
+use {
+  super::*,
+  bitcoin::{blockdata::constants::COIN_VALUE, OutPoint},
+};
+
+#[test]
+fn show_second_uncommon_ordinal() {
```
casey, line 7:
```suggestion
fn identify() {
```

casey (changes requested):
Need test coverage for a few cases:

- `identify` shows zero ordinals
- `identify` shows one ordinal
- `identify` shows two ordinal
- `identify` shows non-zero ordinal offset (to test offset handling)
- `identify` doesn't show anything from output with only common ordinals

One integration tests to test command, the rest unit tests with a wrapper like this:

````
fn identify(utxo: Vec<Utxo>) -> Vec<(Ordinal, OutPoint, u64, Rarity> {
}
```

test-bitcoincore-rpc/src/lib.rs
```diff
@@ -282,6 +292,52 @@ impl Api for Server {
...
+          script_pub_key: Script::new(),
+          amount: Amount::default(),
+          confirmations: 0,
+          spendable: true,
+          solvable: true,
+          descriptor: None,
+          safe: true,
+        })
+        .collect(),
+    )
```
casey, line 339: Not sure if this is right:

```suggestion
    Ok(
      self
        .state
        .lock()
        .unwrap()
        .transactions
        .iter()
        .flat_map(|(txid, tx)| {
          (0..tx.output.len()).map(|vout| ListUnspentResultEntry {
          txid: outpoint.txid,
          vout: outpoint.vout,
          address: None,
          label: None,
          redeem_script: None,
          witness_script: None,
          script_pub_key: Script::new(),
          amount: Amount::default(),
          confirmations: 0,
          spendable: true,
          solvable: true,
          descriptor: None,
          safe: true,
        })
      })
      .collect()
    )
```

casey (changes requested):
Nice, looks good! Only change would be to add an additional test which passes a two different UTXOs with two different outpoints, each with an uncommon ordinal, and makes sure that the right ordinal is associated with the right outpoint.

casey (approved):
Nice, LGTM!

PR #588, opened 2022-09-30
Add new bounty

casey:
To do:
- [ ] Show hints in each bounty instead of in index page
- [ ] Hide index page
- [ ] Titles should be permalinks to page
- [ ] Make times line up


casey:
@raphjaph Merging now, so that it's ready for tomorrow, but please take a look anyways. We can fix anything you find in a follow-up PR.

PR #592, opened 2022-10-02
Add bitcoin.conf

casey:


casey:
I originally did this because I was thinking about running an lnd instance on ordinals.com, and lnd can use bitcoind's config file to connect to it. I decided against it, but this is still probably good to do, because it makes `bicoin-cli` easier to use, because you can pass it the path to the config file.

casey:
@raphjaph Merging now, because I want to deploy for tomorrow. (I deployed this already to mainnet, and I think it's fine to just redploy over it, but I don't want to think to hard. Take a look after the fact anyways!

PR #595, opened 2022-10-03
Fix bounty links

casey:
Since the bounties are now on a different domain, they need full URLs to any URLs on ordinals.com


casey:
@raphjaph Merging this now! Take a look!

PR #601, opened 2022-10-03
add list command

raphjaph:


src/options.rs
```diff
@@ -112,6 +116,19 @@ impl Options {
 
     Ok(path)
   }
+
+  pub(crate) fn rpc_client(&self) -> Result<Client> {
```
casey, line 120:
```suggestion
  pub(crate) fn bitcoin_rpc_client(&self) -> Result<Client> {
```

src/options.rs
```diff
@@ -112,6 +116,19 @@ impl Options {
...
+    let cookie_file = self.cookie_file()?;
+    let rpc_url = self.rpc_url();
+    log::info!(
+      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
+      cookie_file.display()
+    );
+    let client = Client::new(&rpc_url, Auth::CookieFile(cookie_file))
+      .context("Failed to connect to Bitcoin Core RPC at {rpc_url}")?;
+
+    Ok(client)
```
casey, line 130:
```suggestion
    Ok(Client::new(&rpc_url, Auth::CookieFile(cookie_file))
      .context("Failed to connect to Bitcoin Core RPC at {rpc_url}")?)
```

src/subcommand/wallet.rs
```diff
@@ -1,15 +1,18 @@
 use super::*;
 
 mod identify;
+mod list;
 
 #[derive(Debug, Parser)]
 pub(crate) enum Wallet {
+  List,
   Identify,
```
casey, line 9: These should be sorted:

```suggestion
  Identify,
  List,
```

src/subcommand/wallet.rs
```diff
@@ -1,15 +1,18 @@
...
 }
 
 impl Wallet {
   pub(crate) fn run(self, options: Options) -> Result<()> {
     match self {
+      Self::List => list::run(options),
       Self::Identify => identify::run(options),
```
casey, line 16:
```suggestion
      Self::Identify => identify::run(options),
      Self::List => list::run(options),
```

src/subcommand/wallet/list.rs
```diff
@@ -0,0 +1,92 @@
...
+pub(crate) fn run(options: Options) -> Result {
+  let index = Index::open(&options)?;
+  index.index()?;
+
+  let client = options.rpc_client()?;
+
+  let unspent = client.list_unspent(None, None, None, None, None)?;
+
+  let mut utxos = Vec::new();
+  for utxo in unspent {
```
casey, line 12:
```suggestion
  let mut utxos = Vec::new();
  for utxo in client.list_unspent(None, None, None, None, None)? {
```

src/subcommand/wallet/list.rs
```diff
@@ -0,0 +1,92 @@
...
+    let output = OutPoint::new(utxo.txid, utxo.vout);
+    match index.list(output)? {
```
casey, line 14: I think we should implement a method, `Index::list_unspent` which is like `Index::List`, but which returns `Result<Vec<(u64, u64)>>`, i.e, not a `List` enum. It should call `Index::list` internally, and return an error if the output is spent or not seen. Then, we can use this in `ord wallet list` and `ord wallet identify`, and de-duplicate the error handling code.

src/subcommand/wallet/list.rs
```diff
@@ -0,0 +1,92 @@
...
+      }
+      Some(List::Spent) => {
+        bail!("Output {output} in wallet but is spent according to index")
+      }
+      None => bail!("Ordinals index has not seen {output}"),
+    }
+  }
+
+  for (start, size, output, rarity, name) in list(utxos) {
+    println!("{start}\t{size}\t{output}\t{rarity}\t{name}");
```
casey, line 26: I was thinking that maybe we should have more consistent ordering of `ord wallet identify` and `ord wallet list`.

For `identify`: `output ordinal offset rarity name`
And `list`: `output ordinal size rarity name`

Since `list` should be in order, this is also pretty natural, as all the ranges in one output will be grouped together.

casey (changes requested):

src/index.rs
```diff
@@ -512,6 +512,16 @@ impl Index {
     }
   }
 
+  pub(crate) fn list_unspent(&self, outpoint: OutPoint) -> Result<Vec<(u64, u64)>> {
```
casey, line 515: I think this would be better placed in the `wallet` module, and be combined with the `list_unspent` function there.

casey (changes requested):
Only one comment about the location of `Index::list_unspent`.

casey (approved):

PR #619, opened 2022-10-04
Add rare.txt

casey:


casey:
I decided to go with `ordinal\tsatpoint`, for a few reasons:

- The page is already huge, so adding additional data makes it even huger.
- Everything else can be derived from ordinal and satpoint (we could create a google sheet that you can copy/paste the data into, and it would generate the extra tables)
- Encoding two items in a redb value is a little annoying
- Looking up the time for every rare ordinal transferred requires hitting the client, which might make indexing slower

casey:
@raphjaph Merging so I can start rebuilding the database now, but please take a look anyway!

PR #637, opened 2022-10-10
Add `ord rune publish`

casey:
- [x] `ord rune publish` with new returns created
- [x] `/rune/HASH` shows published rune
- [x] `/rune/UNKNOWN_HASH` returns error message and NOT FOUND
- [x] make sure hash is calculated on server
- [x] should /rune/HASH show the network, or is that implied?
- [x] `ord rune publish` with old returns OK
- [x] only allow runes with the correct network to be posted
- [x] mainnet is forbidden


casey:
@raphjaph This should be ready for review!

raphjaph (approved):

PR #639, opened 2022-10-11
Add /input page

casey:
- [x] Test empty scriptsig rendering
- [x] Test Server::input with coinbase transaction
- [x] Test block, transaction, input not found
- [x] don't render null previous output
- [x] don't render max sequence
- [x] don't render empty witness
- [x] Make titles monospace (anything else?)


casey:
@raphjaph This lets you view inputs by going to `/input/PATH`, i.e. `/input/BLOCK/TRANSACTION/VOUT`.

Once runes are in a format that they can be embedded in inputs, we can parse out and display runes on this page.

raphjaph (approved):

PR #647, opened 2022-10-12
Track number of commits

casey:


casey:
@raphjaph This originally logged to the database, but I decided against doing that. However in the process, I realized that we were sometimes committing when we didn't need to. This changes things to only commit when we need to, and furthermore to track the number of commits we've made.

raphjaph (approved):

PR #648, opened 2022-10-12
`ord wallet send` refined

raphjaph:
@casey we should probably walk through this together but you can have a look already


src/subcommand/wallet/send.rs
```diff
@@ -13,36 +13,16 @@ impl Send {
...
 
-    let amount = client
-      .get_transaction(&output.txid, Some(true))?
-      .amount
-      .to_sat()
-      .try_into()
-      .unwrap();
+    let unsigned_transaction = TransactionBuilder::new(utxos, self.ordinal, self.address)
+      .select_ordinal()?
+      .build_transaction();
```
casey, line 20:
```suggestion
    let unsigned_transaction = TransactionBuilder::build_transaction(utxos, self.ordinal, self.address)?;
```

```rust
  enum Error {
  }

  type Result<T> = Result<T, Error>;

    fn build_transaction(...) -> Result<Transaction, Error> {
      Self::new(…)
        .select_ordinal()?
        .build()
    }
```

PR #653, opened 2022-10-14
List prime ordinals on transaction page

casey:
This PR introduces the concept of a "prime ordinal" for a given output. The prime ordinal of an output is just the first ordinal in that output. This might be useful for a few reasons:

- We can consider the prime ordinal of an output that, when spent, reveals and thus inscribes a rune, as the issuer of that rune. This can be used for authentication, so that the creator of a well-known rune can publish more runes, and by seeing that the same prime rune was used for both, can be used to establish the authenticity of later runes.

- I originally tried and gave up on saving ordinal ranges for every historical output that has ever existed. We purge entries from the `OUTPOINT_TO_ORDINAL_RANGES` table when outpoints are spent, in order to keep the database size down. However, keeping track of the historical location of all primes might actually be tractable. We might be able to keep a mapping of all transactions ever made to the primes in the outputs, as well as a mapping of all primes to their current satpoint. This would then allow, for example, quickly looking up all runes issued by a given prime, the prime of a given inscribed rune, and a primes current address. (Also nice is that every uncommon or better ordinal is also a prime, since it was the prime in the coinbase transaction outputs.)

This adds a new large table to the database, so we'll need to make sure that this we can still sync within 48 hours on the ordinals.com box.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -23,6 +23,8 @@ const OUTPOINT_TO_ORDINAL_RANGES: TableDefinition<[u8; 36], [u8]> =
   TableDefinition::new("OUTPOINT_TO_ORDINAL_RANGES");
 const RUNE_HASH_TO_RUNE: TableDefinition<[u8; 32], str> = TableDefinition::new("RUNE_HASH_TO_RUNE");
 const STATISTIC_TO_COUNT: TableDefinition<u64, u64> = TableDefinition::new("STATISTIC_TO_COUNT");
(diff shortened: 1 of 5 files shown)


casey:
@raphjaph This PR also includes a rename of some of the tables, but I split it into its own commit, so each commit should be reviewed individually.

Also, we might not want to merge this before we actually test that we can still sync on the dev machine.

raphjaph (approved):

raphjaph (approved):

casey:
Closing! We might eventually want to do this, but it isn't high priority.

PR #669, opened 2022-10-18
Check that transaction fees are reasonable

casey:


casey:
@raphjaph Take a look! I added a checks that target fee rates are equal to the target fee rate. This requires avoiding calling `.build()` before fees are added, since that will now panic, so I changed fee estimation to be done on a dummy transaction with null inputs, instead of calling `self.build()`.

raphjaph (approved):
lgtm

Issue #686, opened 2022-10-21
Make error messages lowercase with no punctuation


raphjaph:

casey:
Closed by #693.

PR #693, opened 2022-10-21
Lowercase error messages

raphjaph:
#686


casey (changes requested):
There are still some uppercase error messages left, check for `bail!`, `anyhow!`, and strings inside of `.context` and `.with_context`.

casey (approved):

PR #694, opened 2022-10-21
Improve progress bar

casey:
- Set initial position correctly
- Only increment if we're not done
- If the position is greater than the length, update the length
- Clear the progress bar once we're done


casey:
@raphjaph I think this is ready to go! It fixes a bunch of issues with the progress bar.

PR #698, opened 2022-10-23
Ensure that send addressess are valid for current network

casey:
Addresses encode the network that they are valid for. Check this, so that it's not possible to send to, for example, a signet address on mainnet.


raphjaph (approved):
yes, this is very important!

Issue #371, opened 2022-08-24
Index is slow


casey:
What we know
------------

The ordinals.com index is syncing very slowly. At the current rate, around 3 seconds per block, it will take about 9 days to sync to chain tip. Previous syncs were more like 36 hours, so this is a severe performance regression.

A possible culprit is the new `OUTPOINT_TO_TXID` table. This table maps outpoints to the txid, if any, that spends them. This is required to improve the error messages when we don't have an outpoint in OUTPOINT_TO_ORDINAL_RANGES table.

The database used to top out at around 50GB, but is currently 89 gigs and growing.

Things to try
-------------

- [x] revert change and redeploy, but save the old database for inspection. gotta have working block explorer! (@casey)
- [x] using the logs, graph the time it took each block to sync. what does this graph look like? (@terror)
- [x] make a scatter plot with ordinal ranges written and time.  is there a trend there? (@terror)
- [x] make a scatter plot with txs in block and time.  is there a trend there? (@terror)
- [ ] if we estimate how much each table should take up, given how much data it contains, and add them together, is that about the size of the database? If not, maybe we have a space leak, either in our use of REDB, or in REDB itself.
- [ ] figure out how much ram the additional table will take up. is this under 128 gb? if so, deploy to a new server with 128 gigs of ram.
- [ ] Can we trigger this locally? What about on memory constrained VM or server?
- [ ] ping chris once we know a little more, he might have ideas. I think lmdb didn't have this performance cliff, so if there are any differences, that might help.

casey:
Blocked on https://github.com/cberner/redb/issues/344.

casey:
@VirtuallyThere, we should benchmark this more, but it seems like there's already a juicy target for optimization, looking at the flamegraph you posted: https://github.com/cberner/redb/issues/344#issuecomment-1239756024

We're spending around 40% of the time requesting blocks from bitcoin core, and in that thread and context switch stuff that's parallelizing calculating transaction IDs. That can be merged into one background thread that fetches new blocks and calculates TXIDs for transactions in advance. (Alternatively, if it's faster, we could just not calculate TXIDs ahead of time.)

VirtuallyThere:
@casey Hah, I came to the same conclusion this morning. Here are my thoughts.

**What we know (from the flamegraphs)**

1. We spend about 15% of time on what I categorized as "Rust parallelization" or "thread overhead" (the right part on the flamegraph)
[...]

PR #702, opened 2022-10-23
Make index faster by faster checking for uncommon or more rare ordinals

VirtuallyThere:
Working on #371 starting with a trivial improvement:

Replace check for uncommon or better rarity by specialized, more efficient method. This line in index_transaction accounted 70% of the time in index_transaction, so it's worth the more verbose, less general method.

There's more potential here, as self.epoch() can likely be sped up significantly - the binary search takes an inordinate amount of time still.


VirtuallyThere:
I tried to do this separate from #703 but am also Github challenged.

PR #712, opened 2022-10-25
Release 0.1.0

casey:
- Bump version: 0.0.6 → 0.1.0
- Update dependencies
- Start keeping a changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+Changelog
+=========
+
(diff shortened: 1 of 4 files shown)


casey:
@raphjaph I want to get a release out to post about on Twitter, so I'm going to go ahead and merge this, but take a look in arrears! I figure we should start keeping a changelog, and at least being a little careful about breaking changes, so I added one and bumped the version to 0.1.0, we can now update the patch version to indicate a non-breaking change, and update the minor version to indicate a breaking change, particularly changes that require rebuilding the index.

PR #723, opened 2022-10-27
Replace binary search with linear search for epoch construction

VirtuallyThere:


VirtuallyThere:
Unrolling the first four epochs gives a very significant speedup for my benchmark of indexing 1000 blocks starting from block 700000. Index time drops from 44 minutes (linear search) to 31 minutes (unrolled loop). Seems worth it.

Will try to use criterion benchmark for this function.

casey:
Maybe we should unroll the loop the whole way? Each epoch is half as big as the previous, so I think a linear search will always be more performant. Obviously small gains, but it's a little cleaner.

PR #725, opened 2022-10-27
Bounty 3

casey:
This is ready to go. I just made up a syntax for the identify command:

```
ord wallet identify --names frequency.tsv
```

We can change it if we implement something else.


raphjaph (approved):
I like splitting it up into two parts

PR #729, opened 2022-10-28
Search from TSV in ord wallet identify

raphjaph:


casey:
@raphjaph I think this is ready to go! I refactored the code so that every iteration of the loop, it advances through the needles (ordinals) list or haystack list (ranges), and stops when it runs out of either.

raphjaph:
Love the needles in haystacks framing. Also smart not to include the offset of the ordinal in the output, which makes it very fast now.

LGTM

Issue #747, opened 2022-10-29
Indexing on Mac aborts after transient errors from bitcoinrpc


VirtuallyThere:
With background block fetch #495 on my Mac, index runs frequently abort after encountering transient errors from bitcoinrpc. I suspect this is because in `get_block_with_retries` only retries errors in `get_block`, but not errors in `get_block_hash`.

VirtuallyThere:
Last version of #748 still has this issue:

[2022-10-29T09:46:32Z INFO  ord::index::updater] Wrote 1 ordinal ranges from 1 outputs in 0 ms
[2022-10-29T09:46:47Z ERROR ord::index::updater] Failed to fetch block 8173: JSON-RPC error: transport error: Couldn't connect to host: connection timed out
[2022-10-29T09:46:47Z INFO  ord::index::updater] Committing at block height 8173, 8283 outputs traversed, 3001 in map, 498 cached
[2022-10-29T09:46:47Z INFO  ord::index::updater] Flushing 3001 entries (94.0% resulting from 3193 insertions) from memory to database

Will incorporate feedback from casey and switch to more general retry mechanism.

casey:
This should be fixed, now that we're reusing sockets.

Issue #309, opened 2022-08-12
Socket or port exhaustion when making fast RPC calls to bitcoin core


casey:
`JsonRpc(Transport(SocketError(Error { kind: TimedOut, message: "connection timed out" })))`

This is due to using too many sockets or ports. Probably ports, specifically local ephemeral ports.

This is the rust-jsonrpc issue: https://github.com/apoelstra/rust-jsonrpc/issues/67

Need to reuse the socket, or add hyper transport to rust-jsonrpc.

Update: I'm not actually sure what's going on here. We tried the socket-reusing version and it didn't help.

We should:
- Test again with the latest PR version
- Do a comparison with `bitcoin-cli`. If `bitcoin-cli` doesn't have this problem, then it's a good indication that the issue is in jsonrpc.

VirtuallyThere:
Potentially explore moving to https://github.com/paritytech/jsonrpsee

raphjaph:
maybe try again with this fix: https://github.com/apoelstra/rust-jsonrpc/pull/72

PR #748, opened 2022-10-29
Fix get_block_with_retries by retrying both get_block_hash and get_block

VirtuallyThere:
Fix https://github.com/casey/ord/issues/747 by retrying both get_block_hash and get_block

### DIFF
--- a/src/index/updater.rs
+++ b/src/index/updater.rs
@@ -114,7 +114,7 @@ impl Updater {
         }
       }
 
-      match Self::get_block_with_retries(&client, height) {
+      match Self::get_block_with_retries(&client, height, 0) {
         Ok(Some(block)) => {
           if let Err(err) = tx.send(block) {
             log::info!("Block receiver disconnected: {err}");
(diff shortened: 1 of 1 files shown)


casey:
Just about to go to sleep, but here's a sketch of how to do this with a reusable wrapper that can be used to make any RPC retry:


```rust
  pub(crate) fn rpc_with_retries<F, T>(f: F) -> Result<T, bitcoincore_rpc::Error>
  where
    F: Fn() -> Result<T, bitcoincore_rpc::Error>,
  {
    let mut errors = 0;
    loop {
      match f() {
        Err(err) => {
          if cfg!(test) {
            return Err(err.into());
          }

          errors += 1;
          let seconds = 1 << errors;
          log::error!("RPC failed, retrying in {seconds}s: {err}");

          if seconds > 120 {
            log::error!("would sleep for more than 120s, giving up");
            return Err(err.into());
          }

          thread::sleep(Duration::from_secs(seconds));
        }
        Ok(result) => return Ok(result),
      }
    }
  }

  pub(crate) fn get_block_with_retries(client: &Client, height: u64) -> Result<Option<Block>> {
    match Self::rpc_with_retries(|| client.get_block_hash(height)).into_option()? {
      Some(hash) => Self::rpc_with_retries(|| client.get_block(&hash)).into_option(),
      None => Ok(None),
    }
  }
```

casey:
Just tested this, and it doesn't work, because not found is interpreted as an error. Here's a slightly better version:

```rust
  pub(crate) fn rpc_with_retries<F, T, E>(f: F) -> Result<T, E>
  where
    F: Fn() -> Result<T, E>,
    E: Display,
  {
    let mut errors = 0;
    loop {
      match f() {
        Err(err) => {
          if cfg!(test) {
            return Err(err.into());
          }

          errors += 1;
          let seconds = 1 << errors;
          log::error!("RPC failed, retrying in {seconds}s: {err}");

          if seconds > 120 {
            log::error!("would sleep for more than 120s, giving up");
            return Err(err.into());
          }

          thread::sleep(Duration::from_secs(seconds));
        }
        Ok(result) => return Ok(result),
      }
    }
  }

  pub(crate) fn get_block_with_retries(client: &Client, height: u64) -> Result<Option<Block>> {
    match Self::rpc_with_retries(|| client.get_block_hash(height).into_option())? {
      Some(hash) => Ok(Self::rpc_with_retries(|| {
        client.get_block(&hash).into_option()
      })?),
      None => Ok(None),
    }
  }
```

VirtuallyThere:
Mostly obsoleted by bitcoin RPC fix for socket issue on Mac (#309)

PR #750, opened 2022-10-31
Show difficulty target on block page

casey:
Not sure about this since it's additional clutter, but really like it from a pedagogical point of view, since I think the idea of a difficulty target is most easily understood by seeing it along with block hashes.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -419,7 +419,16 @@ impl Server {
       }
     };
 
-    Ok(BlockHtml::new(block, Height(height), index.height().unwrap()).page())
+    Ok(
+      BlockHtml::new(
+        block,
+        Height(height),
+        index
+          .height()
+          .map_err(|err| ServerError::Internal(anyhow!("failed to get index height: {err}")))?,
+      )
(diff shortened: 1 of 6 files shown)


src/subcommand/server.rs
```diff
@@ -133,7 +150,7 @@ impl Server {
 
       let router = Router::new()
         .route("/", get(Self::home))
-        .route("/block/:hash", get(Self::block))
+        .route("/block/:query", get(Self::block))
```
casey, line 153: Test getting block by both hash and height.

src/subcommand/server.rs
```diff
@@ -363,28 +380,46 @@ impl Server {
...
+            ))
+          })?
+          .ok_or_else(|| ServerError::NotFound(format!("block {hash} unknown")))?;
+
+        (block, info.height as u64)
+      }
+    };
 
-    Ok(BlockHtml::new(block, Height(info.height as u64)).page())
+    Ok(BlockHtml::new(block, Height(height), index.height().unwrap()).page())
```
casey, line 422: Fix unwrap

PR #763, opened 2022-11-02
don't send marked ordinals

raphjaph:


src/subcommand/wallet/send.rs
```diff
@@ -42,8 +42,36 @@ impl Send {
...
     ];
 
-    let unsigned_transaction =
-      TransactionBuilder::build_transaction(utxos, self.ordinal, self.address, change)?;
+    let mut marked_ordinals = Vec::new();
+    if let Ok(marked_ordinals_file) = fs::read_to_string(format!(
+      "{}/marked_ordinals.tsv",
+      options.data_dir().unwrap().display()
+    )) {
+      for (i, line) in marked_ordinals_file.lines().enumerate() {
```
casey, line 50: TSV parsing should be deduplicated into a function with this signature:

```
fn parse_tsv(tsv: &str) -> Result<Vec<(Ordinal, &str)>> {
...
```

src/subcommand/wallet/transaction_builder.rs
```diff
@@ -38,6 +38,7 @@ pub(crate) enum Error {
   NotEnoughCardinalUtxos,
   RareOrdinalLostToRecipient(Ordinal),
   RareOrdinalLostToFee(Ordinal),
+  MarkedOrdinalSent(Ordinal),
```
casey, line 41:
```suggestion
  MarkedOrdinalLostToRecipient(String),
  MarkedOrdinalLostToFee(String),
```

src/subcommand/wallet/transaction_builder.rs
```diff
@@ -55,6 +56,9 @@ impl fmt::Display for Error {
       Error::RareOrdinalLostToFee(ordinal) => {
         write!(f, "transaction would lose rare ordinal {ordinal} to fee")
       }
+      Error::MarkedOrdinalSent(ordinal) => {
+        write!(f, "transaction would also send marked ordinal {ordinal}")
```
casey, line 60:
```suggestion
        write!(f, "transaction would lose marked ordinal {ordinal} to fee")
```

src/subcommand/wallet/transaction_builder.rs
```diff
@@ -468,6 +476,19 @@ impl TransactionBuilder {
       }
     }
 
+    for input in &transaction.input {
```
casey, line 479: Should check if marked ordinal winds up in output owned by the user, if so then it's okay.

casey (changes requested):
Some comments, and for a follow-up PR, try to avoid selecting utxos that contain marked ordinals.

PR #821, opened 2022-11-21
Don't create tables when reopening database

casey:


casey:
We shouldn't need to do this. The delay in opening the database was caused by redb zeroing out freed pages, which it no longer does as of https://github.com/cberner/redb/pull/442

PR #826, opened 2022-11-22
Record commit start block count and timestamp

casey:
Store this information in the database for each commit, so we don't need to look at log files to get this information, we can just look at the index.

Also convert `info` command output to JSON, so we can represent data with nested fields.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -79,26 +79,26 @@ impl From<Statistic> for u64 {
 
 #[derive(Serialize)]
 pub(crate) struct Info {
-  blocks_indexed: u64,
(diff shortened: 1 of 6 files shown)


raphjaph (approved):

casey:
This commit adds a table and additional writes for every commit, so benchmarking before merging to make sure this doesn't introduce a performance regression.

casey:
I benchmarked a full sync on ordinals.com overnight and it took 15h23m vs a baseline of 15h50m, so ~27 minutes less, which is probably just noise. It doesn't seem to have an impact on performance, so merging.

PR #832, opened 2022-11-23
Update redb to version that doesn't zero freed pages

casey:
Lots of stuff went away, becuase redb no longer requires that we pass a max index size up-front, and will dynamically resize on demand.


casey:
Running a benchmark on ordinals.com with this commit now.

raphjaph:
Awesome, let's hope it speeds up!

PR #846, opened 2022-11-25
Enable tests on windows

casey:


casey:
@raphjaph This is working, have a look! The `HOME` environment variable isn't used to determine the location of `dirs::data_dir` on Windows, so I changed the tests to just pass `--data-dir` instead.

raphjaph (approved):

PR #881, opened 2022-12-02
do not accidentally send inscription

### LINKED ISSUE #871: Don't send utxos if additional inscriptions aside from the inscription being sent are present

raphjaph:
Fixes #871


src/subcommand/wallet/send.rs
```diff
@@ -53,6 +53,17 @@ impl Send {
       },
     };
 
+    for inscribed_satpoint in &inscription_satpoints {
```
casey, line 56: I think it would be better to move this check into the transaction builder, and also add a unit test for it, so that whenever we use the transaction builder it does the same check. We can also add another error variant to the transaction builder error type, something like `UtxoContainsAdditionalInscription`.

tests/wallet.rs
```diff
@@ -414,3 +414,34 @@ fn refuse_to_reinscribe_sats() {
...
+    vout: 0,
+  };
+
+  CommandBuilder::new(format!(
+    "--chain regtest wallet send {inscription_utxo}:55 bcrt1q6rhpng9evdsfnn833a4f4vej0asu6dk5srld6x"
+  ))
+  .rpc_server(&rpc_server)
+  .expected_exit_code(1)
+  .expected_stderr(format!(
+    "error: trying to send utxo containing another inscription at {inscription_utxo}:0\n",
```
casey, line 444: Maybe something like this?

```suggestion
    "error: cannot send {inscription_utxo} without also sending inscription {inscription_txid}\n",
```

casey (changes requested):

src/subcommand/wallet/send.rs
raphjaph: I added it to the `select_ordinal()` function. If you want I can also make a seperate builder function for it but then we will have to change all the unit tests.

src/subcommand/wallet/transaction_builder.rs
```diff
@@ -36,6 +36,7 @@ use {
 pub(crate) enum Error {
   NotInWallet(SatPoint),
   NotEnoughCardinalUtxos,
+  UtxoContainsAdditionalInscription((SatPoint, SatPoint)),
```
casey, line 39: This should be a struct variant to avoid ambiguity.

casey, line 39: Should mention the ID of the inscription that would be sent:

```suggestion
  UtxoContainsAdditionalInscription((SatPoint, InscriptionId, SatPoint)),
```

casey (changes requested):

casey (approved):
LGTM

PR #896, opened 2022-12-06
Set size limit to 1024 on signet and testnet

casey:
Signet and testnet are both low-volume networks without a functioning fee market. Because of the potential for spam, we don't want to allow large inscriptions on those networks. Set the size limit to 1024 bytes for both signet and testnet, which is large enough that we'll exercise the content data chunking, but small enough that spam shouldn't be an issue. Regtest is a private test network, so spam is not an issue, and mainnet has a robust fee market, so large inscriptions will have to pay correspondingly high fees.

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -30,6 +30,13 @@ impl Chain {
     }
   }
 
+  pub(crate) fn inscription_content_size_limit(self) -> Option<usize> {
+    match self {
+      Self::Mainnet | Self::Regtest => None,
+      Self::Testnet | Self::Signet => Some(1024),
+    }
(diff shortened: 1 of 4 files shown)


src/chain.rs
```diff
@@ -30,6 +30,13 @@ impl Chain {
     }
   }
 
+  pub(crate) fn inscription_content_size_limit(self) -> Option<usize> {
+    match self {
+      Self::Mainnet | Self::Regtest => None,
+      Self::Testnet | Self::Signet => Some(1024),
```
hashbender, line 36: annoying

PR #996, opened 2022-12-16
see latest inscriptions on home page

raphjaph:


src/subcommand/server/templates/home.rs
```diff
@@ -36,22 +38,32 @@ mod tests {
...
+              .unwrap()
+          ),
+          (
+            1260000,
+            "0000000000000000000000000000000000000000000000000000000000000000"
+              .parse()
+              .unwrap()
+          )
+        ],
+        vec![Inscription::new(
```
casey, line 56: You can use `inscription()` here

static/index.css
```diff
@@ -157,3 +157,7 @@ a.legendary {
 a.mythic {
   color: var(--mythic);
 }
+
+.inscriptions {
+  display: grid;
```
casey, line 162: How do inscriptions display?

PR #1001, opened 2022-12-17
add a flag to see all sats in your wallet, even the common ones

rot13maxi:
by default, `ord wallet satoshis` will only show sats that are more-rare than common. I want to see all the sats in my wallet, so this change adds an optional `--all` flag that shows them.

### DIFF
--- a/src/subcommand/wallet/satoshis.rs
+++ b/src/subcommand/wallet/satoshis.rs
@@ -7,6 +7,11 @@ pub(crate) struct Satoshis {
     help = "Find satoshis listed in first column of tab-separated value file <TSV>."
   )]
   tsv: Option<PathBuf>,
+  #[clap(
+  long,
+  help = "List all satoshis, regardless of rarity (by default only list rarer than common)."
+  )]
+  all: bool,
 }
 
 impl Satoshis {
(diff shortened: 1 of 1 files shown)


casey:
I'm wondering if we should introduce a new command for this. Currently, `ord wallet satoshis` either shows individual rare satoshis, or it looks for any satoshis in the wallet that are present in a supplied TSV file.

We could extend it to show the first satoshi of every range, regardless of rarity, but then the output format isn't ideal, since it doesn't show the size of the range.

We could introduce a new command, something like `ord wallet ranges`, which would display all ranges in all utxos in the wallet, and leave `ord wallet satoshis` for returning information about individual sats.

Actually, maybe I should ask first, what's the use-case for this command?

rot13maxi:
That sounds right. I think the thing I'm looking for is actually a `ranges` command. I'm going to close this for now and think a bit more about what will actually be useful for me.

casey:
We actually used to have a ranges command, but we removed it for some reason. The fact that satoshis are tracked by range definitely feels like an implementation detail, so maybe making `ord wallet satoshis` display ranges with a flag is really what we should do.

PR #1035, opened 2022-12-21
Support SVG and HTML inscriptions

casey:
This was easier than I thought. This serves all content with a CSP header value of `default-src 'none'`. When content is loaded in an iframe with this CSP, all outgoing requests from the iframe are disallowed. We can then serve SVG and HTML content in frames, and they can't make outgoing requests, lessening security concerns and avoiding breakage and references to off-chain content.

### DIFF
--- a/src/content.rs
+++ b/src/content.rs
@@ -1,5 +1,6 @@
 #[derive(Debug, PartialEq)]
 pub(crate) enum Content<'a> {
+  Html,
   Image,
   Svg,
   Text(&'a str),
--- a/src/content.rs
+++ b/src/content.rs
@@ -1,5 +1,6 @@
 #[derive(Debug, PartialEq)]
 pub(crate) enum Content<'a> {
-  Text(&'a str),
   Image,
+  Svg,
+  Text(&'a str),
 }
--- a/src/content.rs
+++ b/src/content.rs
@@ -1,7 +1,6 @@
 #[derive(Debug, PartialEq)]
 pub(crate) enum Content<'a> {
-  Html,
+  IFrame,
   Image,
-  Svg,
   Text(&'a str),
 }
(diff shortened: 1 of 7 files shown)


rot13maxi:
Have you tried putting XSS into an html inscription?

raphjaph (approved):

casey:
> Have you tried putting XSS into an html inscription?

Good question!

The answer is yes, but I think it's okay. (Famous last words.)

This PR serves all content with the following header:

```
Content-Security-Policy: default-src 'none' 'unsafe-eval' 'unsafe-inline'
```

This allows HTML inscriptions to include and execute arbitrary JS. However, `default-src 'none'` prevents all outgoing requests, so it can't load content from other sites, or make requests to other websites.

It does load from the same origin as other content from ordinals.com, but having the same origin as ordinals.com is not exploitable, because it isn't a sensitive browsing context. It doesn't have any cookies, or access to any special permissions.

If ordinals.com ever becomes a sensitive browsing context, e.g. we have cookies, we'll need to start serving inscription content from a different origin, so that it doesn't have access to those cookies.

We eventually want to relax the sandboxing, so that that inscriptions can request other inscriptions, so eventually we'll want to:

- Serve inscription content from a unique origin, e.g. content.ordinals.com
- Allow requests only to content.ordinals.com
- Serve inscription content and nothing else from content.ordinals.com
- Change the content security policy that content is served with to allow requests to content.ordinals.com

The goal is to allow HTML inscriptions that can build and remix other inscriptions. We might also want to serve other stuff from `content.ordinals.com`, like the current block hash, so that HTML inscriptions can use the current block hash as a source of randomness.

casey:
@rot13maxi Definitely let me know if the above makes sense or not, and if I'm missing anything.

PR #1047, opened 2022-12-22
add balance command

### LINKED ISSUE #1036: Add wallet balance command

rot13maxi:
Add a simple balance command. Here's what it looks like:

```
$ ./target/debug/ord --signet wallet balance
0.0099852 BTC
```

closes #1036

### DIFF
--- a/src/subcommand/wallet.rs
+++ b/src/subcommand/wallet.rs
@@ -1,5 +1,6 @@
 use {super::*, transaction_builder::TransactionBuilder};
 
+mod balance;
 mod create;
 mod inscribe;
 mod inscriptions;
@@ -12,6 +13,8 @@ mod utxos;
 
 #[derive(Debug, Parser)]
 pub(crate) enum Wallet {
+  #[clap(about = "Get wallet balance")]
+  Balance(balance::Balance),
   #[clap(about = "Create a new wallet")]
   Create(create::Create),
   #[clap(about = "Create an inscription")]
@@ -33,6 +36,7 @@ pub(crate) enum Wallet {
 impl Wallet {
   pub(crate) fn run(self, options: Options) -> Result {
     match self {
+      Self::Balance(balance) => balance.run(options),
       Self::Create(create) => create.run(options),
       Self::Inscribe(inscribe) => inscribe.run(options),
       Self::Inscriptions(inscriptions) => inscriptions.run(options),
--- a/src/subcommand/wallet.rs
+++ b/src/subcommand/wallet.rs
@@ -14,7 +14,7 @@ mod utxos;
 #[derive(Debug, Parser)]
 pub(crate) enum Wallet {
   #[clap(about = "Get wallet balance")]
-  Balance(balance::Balance),
+  Balance,
   #[clap(about = "Create a new wallet")]
   Create(create::Create),
   #[clap(about = "Create an inscription")]
@@ -36,7 +36,7 @@ pub(crate) enum Wallet {
 impl Wallet {
   pub(crate) fn run(self, options: Options) -> Result {
     match self {
-      Self::Balance(balance) => balance.run(options),
+      Self::Balance => balance::run(options),
       Self::Create(create) => create.run(options),
       Self::Inscribe(inscribe) => inscribe.run(options),
       Self::Inscriptions(inscriptions) => inscriptions.run(options),
--- a/src/subcommand/wallet/balance.rs
+++ b/src/subcommand/wallet/balance.rs
@@ -1,0 +1,20 @@
+use super::*;
+
+#[derive(Debug, Parser)]
+pub(crate) struct Balance {}
+
+impl Balance {
+  pub(crate) fn run(self, options: Options) -> Result {
+    println!(
+      "{}",
+      options
+        .bitcoin_rpc_client_for_wallet_command("ord wallet balance")?
+        .get_balances()?
+        .mine
+        .trusted
+        .to_sat()
+    );
+
+    Ok(())
+  }
+}
--- a/src/subcommand/wallet/balance.rs
+++ b/src/subcommand/wallet/balance.rs
@@ -1,20 +1,15 @@
 use super::*;
 
-#[derive(Debug, Parser)]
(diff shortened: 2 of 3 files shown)


casey:
Nice! Don't worry about rebasing. I'll just squash at the end.

I'm leaning towards displaying the balance in sats:

- BTC are enormous, so probably too large to be useful, and people aren't good at reading long decimals.
- The rest of the commands are sat oriented.
- If we accept BTC amounts as input, I'd rather avoid floating point numbers.
- Everything is pretty sat-oriented.

What do you think?

This could use an integration test in `tests/wallets.rs`. It should run `ord wallet balance`, get a zero balance, mine some coins to the wallet, and then run the command again and see a nonzero balance. I'm not sure if we'll need to implement a new test RPC server command, but we can cross that bridge when we get to it.

rot13maxi:
ended up doing a rebase anyway. 
Now displays balance in sats, and there's an integration test. Let me know if there's anything else you want to see in here

src/subcommand/wallet.rs
```diff
@@ -33,6 +36,7 @@ pub(crate) enum Wallet {
 impl Wallet {
   pub(crate) fn run(self, options: Options) -> Result {
     match self {
+      Self::Balance(balance) => balance.run(options),
```
casey, line 39: We don't actually need a struct here, because balance doesn't take any options. This can be:
```suggestion
      Self::Balance => balance::run(options),
```

src/subcommand/wallet/balance.rs
```diff
@@ -0,0 +1,20 @@
...
+        .bitcoin_rpc_client_for_wallet_command("ord wallet balance")?
+        .get_balances()?
+        .mine
+        .trusted
+        .to_sat()
+    );
+
+    Ok(())
+  }
+}
```
casey, line 20: Indentation is wrong, but this should be something like:

```suggestion
  pub(crate) fn run(options: Options) -> Result {
    println!(
      "{}",
      options
        .bitcoin_rpc_client_for_wallet_command("ord wallet balance")?
        .get_balances()?
        .mine
        .trusted
        .to_sat()
    );

    Ok(())
  }
```

Since we don't need a struct. I also think this can use `Options::bitcoin_rpc_client`, since this command is pretty safe, and we don't need to enforce not running it on mainnet or on wallets with large balances.

tests/wallet.rs
```diff
@@ -688,3 +688,21 @@ fn inscribe_gif() {
...
+    .rpc_server(&rpc_server)
+    .expected_stdout("0\n")
+    .run();
+
+  let coinbase_tx = &rpc_server.mine_blocks_with_subsidy(1, 1_000_000)[0].txdata[0];
+  let amount = coinbase_tx.output[0].value;
+
+  CommandBuilder::new("--regtest wallet balance")
+    .rpc_server(&rpc_server)
+    .expected_stdout(format!("{amount}\n"))
```
casey, line 706: I think I'd avoid the constant, since it requires an extra line and doesn't do a ton for readability:

```suggestion
  CommandBuilder::new("--regtest wallet balance")
    .rpc_server(&rpc_server)
    .expected_stdout("1000000\n")
```

You could also use mine_blocks and make the value 5000000000.

casey (changes requested):
Looking good, a few minor comments.

src/subcommand/wallet.rs
rot13maxi: cool! Still figuring out the internal structure here. thats helpful

src/subcommand/wallet/balance.rs
rot13maxi: :+1:

tests/wallet.rs
```diff
@@ -688,3 +688,20 @@ fn inscribe_gif() {
...
+#[test]
+fn wallet_balance() {
+  let rpc_server = test_bitcoincore_rpc::spawn_with(Network::Regtest, "ord");
+
+  CommandBuilder::new("--regtest wallet balance")
+    .rpc_server(&rpc_server)
+    .expected_stdout("0\n")
+    .run();
+
+  let _ = &rpc_server.mine_blocks(1);
```
casey, line 701: You can just do:

```suggestion
  rpc_server.mine_blocks(1);
```

`let _ =` is only needed when ignoring a type that has `#[must_use]` on it, like `Result`.

tests/wallet.rs
```diff
@@ -688,3 +688,20 @@ fn inscribe_gif() {
...
+
+  CommandBuilder::new("--regtest wallet balance")
+    .rpc_server(&rpc_server)
+    .expected_stdout(format!("5000000000\n"))
```
casey, line 705: Since this doesn't have any interpolations, format is unnecessary:

```suggestion
    .expected_stdout("5000000000\n")
```

There's a just recipe for running some of the checks that run on ci, try `just ci`, or look for the ci recipe in the justfile to find the commands to run. CI runs clippy (which would complain about this), checks formatting, and runs bin/forbid, which forbids certain strings from appearing in the codebase.

bin/forbid is actually quite useful. It forbids the string `todo`, so if you want to make sure you don't forget something before comitting, write `todo: I need to do blah blah blah` and CI will complain, so you don't need to worry about forgetting it.

casey (changes requested):
vv minor changes

tests/wallet.rs
rot13maxi: ah! I haven't used `just` before. I'll check that out so you don't have to keep commenting on things that the linter would check. thanks!

tests/wallet.rs
rot13maxi: right you are

tests/wallet.rs
casey: No worries! [just](https://github.com/casey/just/) is kind of like make, but only for saving and running commands, and less weird.

rot13maxi: yeah, this is great. should be a lot easier to make sure im conforming to style, etc. for future PRs :)

casey (approved):
LGTM!

PR #1048, opened 2022-12-22
Check core version before attempting inscription.

### LINKED ISSUE #1045: Check bitcoin core version when inscribing
We require 24.0 for raw taproot descriptors, so check the version on startup and return an error if it's < 24.0.

rot13maxi:
Example of the error that it throws:
```
error: Bitcoin Core version 240000 or greater required. Current Bitcoin Core reports version 230000
```
the current release `24.0.1` reports as 240001.

closes #1045


casey:
Nice! Don't worry about rebasing, I always just squash-merge in the end, and can ping you if there are conflicts that I can't resolve.

This could use an integration test. There is a mock Bitcoin Core API server, in `test-bitcoincore-rpc`, that we run tests against. The tests are currently failing, because we return `230000` in `getnetworkinfo`, which must be what `bitcoincore-rpc` uses to figure out the version.

Check out `tests/wallet.rs` for some examples.

To make things less annoying, I pushed a PR adding a builder to the test-bitcoincore-rpc crate to this PR. So to add the test you can:

- Create a new test in tests/wallet.rs
- Build a test-bitcoincore-rpc server using the builder that sets the version to less than 240000
- Make sure you get an error

Additionally, the version numbers that core returns are pretty unintelligible, so we should fix them before reporting them to the user. Something like:

```
fn format_bitcoin_core_version(version: usize) -> String {
  format!(
    "{}.{}.{}",
    version / 10000,
    version % 10000 / 100,
    version % 100
  )
}
```

rot13maxi:
great feedback, thanks!

casey:
Whoops, I think I jumped the gun on reviewing this. I pushed a diff formatting everything. I think your editor must have auto-formatted without consulting rustfmt.toml.

casey:
I think you accidentally nuked the via force-push the commit I added with the builder, so I resurrected it.

casey:
Wanted to get this in, since I think it's a common error!

rot13maxi:
been out of town for a bit and hadn't touched this. thanks for the commits!

PR #1049, opened 2022-12-22
add option to transaction-list command to specify count

### LINKED ISSUE #1043: Add flag to list all txs (or larger set of txs) in terminal
eg. ord wallet transactions -a

currently lists 10

rot13maxi:
closes #1043


casey:
What about calling this `--limit` instead? This could use a test, check out my comments on the other two PRs.

casey:
Also, should the limit be unlimited instead?

casey:
*default limit

casey:
Just added some tests, and changes the flag to `--limit`. Take a look and let me know what you think! I want to get this in, because I think having an implicit limit of 10 is a bit odd.

PR #1058, opened 2022-12-22
Show text inscriptions on homepage

casey:
I think this is worth doing, even if they're a little awkward looking. We want people to know that text inscriptions are possible.


raphjaph (approved):
There's a lot of overlap with #1058 but looks good.

PR #1067, opened 2022-12-23
Add content-encoding

casey:
Working on adding compression, but it's a bit tricky:

- A response with `Content-Encoding` should only be sent in response to a request with an appropriate `Accept-Encoding`. This means that the ord server can't just blindly respond with compressed content in response to `/content` requests. It must decompress the content if the request doesn't pass a `Accept-Encoding` header with the content's `Content-Encoding`

- Text content is inserted into a template on the server side, so compressed text content must be uncompressed server-side regardless

Uncompressing server-side is tricky.

Content is untrusted, so we have to avoid zip-bombs. However, it seems like always decompressing server-side is probably the way to go. Later, we can optimize by passing content through untouched if the client sends us an `Accept-Encoding` which matches the `Content-Encoding`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,6 +81,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


casey:
I'm going to close this for now. Rust-brotli doesn't seems to protect against zip bombs, so it's not usable to decompress untrusted content.

Issue #1068, opened 2022-12-23
Avoid scary firewall popup while developing


casey:
Could we switch the server address from `0.0.0.0:80` to `127.0.0.1:80` to avoid the scary firewall popup on macos?

raphjaph:
Tried to do this but my mac doesn't allow me to listen on 127.0.0.1:80 with sudo. Strangely enough it allows me to listen on 0.0.0.0:80 without sudo. The firewall pop up appears but I can connect from localhost without clicking on it. I tested with with netcat.

Doesn't work:
```
nc -l 127.0.0.1 80
nc: Permission denied
```

Works:
```
 nc -l 0.0.0.0 80

```
works

casey:
I noticed that it requires sudo for 127.0.0.1 and not 0.0.0.0, super weird. Maybe we'll have to switch to 127.0.0.1:8000. Ports below 1024 are "well known ports", and unixes require admin privs to bind to them.

A good first step for this issue would be to figure out which test, in particular, is causing this issue. It might just be one test, and then we can pass different --address / --http-port flags in that one test to fix the issue.

PR #1072, opened 2022-12-23
#[cfg(test)] only works for unit tests

raphjaph:
Just opened this to see if you know a better way of doing this. The `#[cfg(test)]` unfortunately does not work for integration tests: https://users.rust-lang.org/t/cfg-test-does-not-work-with-integration-tests/36630/5.

#1068

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -210,19 +210,16 @@ impl Server {
       #[cfg(not(test))]
       {
         (self.address.as_str(), port)
-          .to_socket_addrs()?
-          .next()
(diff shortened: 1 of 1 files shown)


casey:
Yah,`cfg(test)` is never true in binaries. You can `integration_test()` which is in main.rs to see if you're running in an integration test.

casey:
See my comments in #1068, maybe we can fix this by modifying just the test that's causing the issue, and not the default behavior of all tests.

raphjaph:
ok, then i'll close this

Issue #1074, opened 2022-12-23
Test importdescriptors: https://github.com/rust-bitcoin/rust-bitcoincore-rpc/issues/264


casey:

raphjaph:
I tried to do this today but couldn't figure out which code you uploaded to crates.io for the `ord-bitoincore-rpc` crate. Compiling `ord` with the master and ord branch of `casey/rust-bitcoincore-rpc` threw a bunch of errors. I'm probably doing something wrong with the patch command though.

casey:
Can you open a draft PR with the changes you made so I can take a look?

casey:
Whoops, didn't mean to close.

raphjaph:
#1075

PR #1075, opened 2022-12-23
use local copy

raphjaph:
I had to rename the crate  name in the `Cargo.toml` in `/casey/rust-bitcoincore-rpc` as well. And I also don't understand how the sub crates work (client vs. json). It's probably not complicated I just confused myself.

#1074


raphjaph:
Can I close this?

PR #1089, opened 2022-12-25
Add preview subcommand

casey:
The test is slow, so I put it behind `#[ignore]`.


raphjaph (approved):
If I understand this correctly sandboxing restricts what is executed on `ordinals.com` inside the `iframe` but if I click into the `/content` endpoint the execution of the script is allowed because of the `unsafe-inline` and `unsafe-eval`.

casey:
Which example are you referring to?

PR #1096, opened 2022-12-26
Add quickstart script for macos

casey:
I thought that a script with all the commands needed to get started on macos would be useful. We should probably link to it in the docs.


raphjaph:
jup, looks good

PR #1099, opened 2022-12-26
Upgrade to redb 0.11.0

cberner:
I haven't finished benchmarking the mmap backend against the new safe one, so I left ord using the mmap backend.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2380,9 +2380,9 @@ dependencies = [
 
 [[package]]
 name = "redb"
-version = "0.10.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


casey:
Nice! Thanks for the PR! Can `AccessGuard<T>` implement `Deref<T>`? Not sure if this is possible, but if it were, it might make them slightly more convenient.

cberner:
I don't think that's possible today, because `.value()` returns the lifetime of `&self`. If they have a `LendingDeref` in the future that would work though.

casey:
Can't `Deref::deref` borrow `&self`? 

The trait is `fn deref(&self) -> &Self::Target;` which has an elided lifetime, so it's actually `fn deref<'a>(&'a self) -> &'a Self::Target;`

cberner:
oh right, duh. I mixed this up with a different issue I had =P The problem with `Deref` is that it returns `&Self::Target`, but I need to return `Self::Target` for types like `u64`

PR #1107, opened 2022-12-29
Unrecognized even fields render inscriptions invalid

### LINKED ISSUE #1046: Add version number to inscriptions
This would be a relatively simple way to do upgrades. Not sure if a good idea though.

casey:
Fixes #1046. This doesn't add a version, but allows us to add one in the future. We can add a version field with an even tag, which will render it invalid for old clients, allowing for future upgrades.

### DIFF
--- a/src/inscription.rs
+++ b/src/inscription.rs
@@ -219,7 +219,7 @@ impl<'a> InscriptionParser<'a> {
       let content_type = fields.remove(CONTENT_TYPE_TAG);
 
       for tag in fields.keys() {
-        if let Some(lsb) = tag.get(0) {
+        if let Some(lsb) = tag.first() {
           if lsb % 2 == 0 {
             return Err(InscriptionError::UnrecognizedEvenField);
           }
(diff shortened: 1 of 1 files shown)


src/inscription.rs
```diff
@@ -214,9 +215,20 @@ impl<'a> InscriptionParser<'a> {
         }
       }
 
+      let content = fields.remove(CONTENT_TAG);
+      let content_type = fields.remove(CONTENT_TYPE_TAG);
+
+      for tag in fields.keys() {
+        if let Some(lsb) = tag.first() {
+          if lsb % 2 == 0 {
+            return Err(InscriptionError::UnrecognizedEvenField);
```
raphjaph, line 224: So non-breaking changes cause an InscriptionError?

raphjaph (changes requested):
Odd version number means breaking change and even means non-breaking?

raphjaph (approved):

PR #1108, opened 2022-12-29
Track inscription locations correctly

### LINKED ISSUE #861: Support tracking inscriptions at offset/vout other than 0
Our initial inscription tracking code assumes that inscriptions are at offset and vout 0. However, inscriptions can be at any offset in any transaction input and move to any offset in transaction outputs, so we should support this.

This supports future use-cases like:
- inscriptions being lost to fees
- multiple inscriptions are merged into a single UTXO

casey:
Fixes #861, eventually. This first diff just indexes output values of all the transactions we see, since we need output values to figure out where inscriptions are actually winding up.

To Do:
- Inscriptions can be lost to fee now or later?
- Inscriptions can be lost to fee on creation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -365,7 +365,7 @@ dependencies = [
  "matchit",
  "memchr",
  "mime",
- "percent-encoding 2.2.0",
+ "percent-encoding",
  "pin-project-lite",
  "rustversion",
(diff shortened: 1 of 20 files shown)


raphjaph (approved):
gnar and chonk

PR #1125, opened 2022-12-30
Fee-spent but unlost inscriptions

casey:
@raphjaph This is a truly horrific PR, but I wrote a bunch of tests and tested it from a few angles and I think it works.

This does a few things:
- Tracks inscriptions that are spent to fees
- Tracks inscriptions that are spent to fees that are then underpaid (thus destroying the sat). These have their locations updated to a satpoint consisting of a null outpoint, with an offset which increments as sats are lost
- Tracks rare sats that are spent to fees

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -798,6 +798,10 @@ mod tests {
       self.index.update().unwrap();
       blocks
     }
+
+    fn configurations() -> Vec<Context> {
+      vec![Context::with_args(""), Context::with_args("--index-sats")]
+    }
   }
 
   #[test]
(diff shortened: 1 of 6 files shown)


chendatony31:
Hi @casey,  do you have a block number example that meets the condition `Tracks inscriptions that are spent to fees that are then underpaid (thus destroying the sat). These have their locations updated to a satpoint consisting of a null outpoint, with an offset which increments as sats are lost`

chendatony31:
Is `destroying the sat` or `lost sat` mean the coinbase transaction didn't pay all of the fees and reward to the miner?

PR #1127, opened 2022-12-30
Check schema version when opening database

casey:


casey:
@raphjaph This should be ready for review

PR #1158, opened 2023-01-03
create wallet with a taproot descriptor

raphjaph:


src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,56 @@
...
+      xkey.fingerprint(&secp256k1),
+      DerivationPath::master()
+        .child(ChildNumber::Hardened { index: 86 })
+        .child(ChildNumber::Hardened { index: 0 })
+        .child(ChildNumber::Hardened { index: 0 }),
+    )),
+    xkey,
+    derivation_path: DerivationPath::master().child(ChildNumber::Normal { index: 0 }),
+    wildcard: Wildcard::Unhardened,
+  });
```
casey, line 34: The descriptor now looks correct, but it doesn't actually reflect how it was derived.

The origin shows a derivation path, but the xkey is still the original, master, underived xkey. We need to use a derived key, like so:

```suggestion
  let master_xkey = ExtendedPrivKey::new_master(options.chain().network(), &seed)?;

  let fingerprint = xkey.fingerprint(&secp256k1)
  
  let derivation_path = DerivationPath::master()
        .child(ChildNumber::Hardened { index: 86 })
        .child(ChildNumber::Hardened { index: 0 })
        .child(ChildNumber::Hardened { index: 0 });
        
  let derived_xkey = master_xkey.derive_priv(secp256k1, derivation_path)?;

  let secret_key = DescriptorSecretKey::XPrv(DescriptorXKey {
    origin: Some((fingerprint, derivation_path)),
    xkey: derived_xkey,
    derivation_path: DerivationPath::master().child(ChildNumber::Normal { index: 0 }),
    wildcard: Wildcard::Unhardened,
  });
```

src/options.rs
```diff
@@ -164,6 +190,14 @@ impl Options {
         );
       }
     }
+
+    for descriptor in client.list_descriptors(None)?.descriptors {
+      let desc = descriptor.desc;
+      if !Regex::new(r"(tr|rawtr).*").unwrap().is_match(&desc) {
```
casey, line 196: This should be a little more specific, we should require exactly one tr descriptor, 0 or more rawtr descritpros, and no other descriptors.

casey (changes requested):

src/options.rs
```diff
@@ -132,20 +141,37 @@ impl Options {
...
 
     if self.chain() == Chain::Mainnet {
       bail!("`{command}` is unstable and not yet supported on mainnet.");
     }
     Ok(client)
   }
 
-  pub(crate) fn bitcoin_rpc_client_for_wallet_command(&self, command: &str) -> Result<Client> {
-    let client = self.bitcoin_rpc_client()?;
+  pub(crate) fn bitcoin_rpc_client_for_wallet_command(
```
casey, line 169: I think you're right about not passing the min version everywhere, let's only check the version in wallet commands, and they can all require the same min version.

src/options.rs
```diff
@@ -164,6 +190,21 @@ impl Options {
         );
       }
     }
+
+    let mut num_tr = 0;
+    for descriptor in client.list_descriptors(None)?.descriptors {
+      let desc = descriptor.desc;
+      if Regex::new(r"tr.*").unwrap().is_match(&desc) {
```
casey, line 197: We can use `.starts_with("tr(")`

casey: Can also do:

```rust
let tr = descriptors.iter().filter(|descriptor| descriptor.starts_with("tr(")).count();
let rawtr = descriptors.iter().filter(|descriptor| descriptor.starts_with("rawtr(")).count();

if tr != 2 {
  bail!(…);
}

if descriptors.len() != 2 + rawtr {
  bail!(…);
}
```

casey: The error messages for both bails should be more descriptive, something like "this does not appear to be an ord wallet, please create an ord wallet using `ord wallet create`".

tests/wallet/create.rs
```diff
@@ -0,0 +1,56 @@
...
+    .build();
+
+  CommandBuilder::new("--chain regtest wallet create")
+    .rpc_server(&rpc_server)
+    .run();
+
+  assert_eq!(rpc_server.descriptors().len(), 2);
+  assert_regex_match!(
+    &rpc_server.descriptors()[0],
+    r"tr\(\[.*/86'/0'/0'\]tprv.*/0/\*\)#.*"
```
casey, line 31:
```suggestion
    r"tr\(\[[[:xdigit:]]{8}/86'/0'/0'\]tprv[a-zA-Z1-9]*/0/\*\)#[[:xdigit]]{8}"
```

tests/wallet/create.rs
```diff
@@ -0,0 +1,56 @@
...
+  );
+  assert_regex_match!(
+    &rpc_server.descriptors()[1],
+    r"tr\(\[.*/86'/0'/0'\]tprv.*/1/\*\)#.*"
```
casey, line 35:
```suggestion
    r"tr\(\[[[:xdigit:]]{8}/86'/0'/0'\]tprv[a-zA-Z1-9]*/0/\*\)#[[:xdigit]]{8}"
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,78 @@
...
+};
 
 pub(crate) fn run(options: Options) -> Result {
   options
-    .bitcoin_rpc_client_mainnet_forbidden("ord wallet create")?
-    .create_wallet("ord", None, None, None, None)?;
+    .bitcoin_rpc_client_for_wallet_command("ord wallet create", 240000)?
+    .create_wallet("ord", None, Some(true), None, None)?;
+
+  let secp256k1 = bitcoin::secp256k1::Secp256k1::new();
```
casey, line 14:
```suggestion
  let secp = bitcoin::secp256k1::Secp256k1::new();
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,78 @@
...
+  let mut seed = [0; 32];
+  bitcoin::secp256k1::rand::thread_rng().fill_bytes(&mut seed);
+
+  let master_xkey = ExtendedPrivKey::new_master(options.chain().network(), &seed)?;
```
casey, line 18:
```suggestion
  let master_private_key = ExtendedPrivKey::new_master(options.chain().network(), &seed)?;
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,78 @@
...
+
+  let fingerprint = master_xkey.fingerprint(&secp256k1);
+
+  let derivation_path = DerivationPath::master()
+    .child(ChildNumber::Hardened { index: 86 })
+    .child(ChildNumber::Hardened { index: 0 })
```
casey, line 24: I believe this should reflect the network. `0` on mainnet, and `1` otherwise.

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,78 @@
...
+    .child(ChildNumber::Hardened { index: 0 });
+
+  let derived_xkey = master_xkey.derive_priv(&secp256k1, &derivation_path)?;
```
casey, line 27:
```suggestion
  let derived_private_key = master_xkey.derive_priv(&secp256k1, &derivation_path)?;
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,78 @@
...
+
```
casey, line 28: fn derive_and_import_descriptor(master_key, derivation_path) {
  …
}

casey (changes requested):

src/subcommand/wallet.rs
```diff
@@ -97,7 +97,7 @@ fn get_unspent_outputs(options: &Options) -> Result<BTreeMap<OutPoint, Amount>>
 }
 
 fn get_change_addresses(options: &Options, n: usize) -> Result<Vec<Address>> {
-  let client = options.bitcoin_rpc_client(0)?;
+  let client = options.bitcoin_rpc_client()?;
```
casey, line 100: Should this be a wallet command?

tests/wallet.rs
```diff
@@ -165,22 +181,22 @@ fn send_inscribed_sat() {
 
 #[test]
 fn send_on_mainnnet_refuses_to_work_with_wallet_name_foo() {
-  let rpc_server = test_bitcoincore_rpc::builder().wallet_name("foo").build();
-  let txid = rpc_server.mine_blocks(1)[0].txdata[0].txid();
+  let rpc_server = test_bitcoincore_rpc::builder().build();
 
   CommandBuilder::new(
-    format!("wallet send bc1qzjeg3h996kw24zrg69nge97fw8jc4v7v7yznftzk06j3429t52vse9tkp9 {txid}:0:0"),
```
casey: We need to somehow test that send fails.

casey: (You can create the wallet with the bitcorerpc Client directly, and then test that send refuses to operate)

src/options.rs
```diff
@@ -136,17 +145,29 @@ impl Options {
...
+    if bitcoin_version < MIN_VERSION {
+      bail!(
+        "Bitcoin Core {} or newer required, current version is {}",
+        Self::format_bitcoin_core_version(MIN_VERSION),
+        Self::format_bitcoin_core_version(bitcoin_version),
+      );
+    }
+
     if self.chain() == Chain::Mainnet {
       let wallet_info = client.get_wallet_info()?;
```
casey, line 172: Only perform this check if the command is not ord wallet create, so, we don't break mainnet.

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,104 @@
...
+    util::bip32::{ChildNumber, DerivationPath, ExtendedPrivKey, Fingerprint},
+    Network,
+  },
+  bitcoincore_rpc::bitcoincore_rpc_json::{ImportDescriptors, Timestamp},
+  miniscript::descriptor::{Descriptor, DescriptorSecretKey, DescriptorXKey, Wildcard},
+};
+
+#[derive(Debug, Parser)]
+pub(crate) struct Create {
+  #[clap(long, default_value = "ord", help = "Give <NAME> to wallet")]
```
casey, line 14:
```suggestion
  #[clap(long, default_value = "ord", help = "Create wallet with <NAME>")]
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,104 @@
...
+  pub(crate) name: String,
+}
+
+impl Create {
+  pub(crate) fn run(&self, options: Options) -> Result {
+    if !(self.name == "ord" || self.name.starts_with("ord-")) {
+      bail!("`ord wallet create` may only be used on mainnet with a wallet named `ord` or whose name starts with `ord-`");
```
casey, line 21:
```suggestion
      bail!("`ord wallet create` may only be used with a wallet named `ord` or whose name starts with `ord-`");
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,104 @@
...
+      DerivationPath::master()
+        .child(ChildNumber::Hardened { index: 86 })
+        .child(ChildNumber::Hardened { index: 0 })
+        .child(ChildNumber::Hardened { index: 0 })
+    } else {
+      DerivationPath::master()
+        .child(ChildNumber::Hardened { index: 86 })
+        .child(ChildNumber::Hardened { index: 1 })
+        .child(ChildNumber::Hardened { index: 0 })
+    };
```
casey, line 46:
```suggestion
    let derivation_path = DerivationPath::master()
        .child(ChildNumber::Hardened { index: 86 })
        .child(ChildNumber::Hardened { index: if options.chain().network() == Network::Bitcoin {
          0
        } else {
          1
        } })
        .child(ChildNumber::Hardened { index: 0 });
```

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,104 @@
...
+  pub(crate) name: String,
+}
+
+impl Create {
+  pub(crate) fn run(&self, options: Options) -> Result {
+    if !(self.name == "ord" || self.name.starts_with("ord-")) {
+      bail!("`ord wallet create` may only be used on mainnet with a wallet named `ord` or whose name starts with `ord-`");
+    }
+
+    options
```
casey, line 24: We can reuse this client and pass it into derive_and_import_descriptor

test-bitcoincore-rpc/src/lib.rs
```diff
@@ -223,6 +227,15 @@ impl Handle {
   pub fn lock(&self, output: OutPoint) {
     self.state().locked.insert(output);
   }
+
+  pub fn network_flag(&self) -> String {
+    match self.state().network {
+      Network::Bitcoin => "".into(),
```
casey, line 233: This can be `--chain mainnet`

tests/wallet.rs
```diff
@@ -164,22 +182,22 @@ fn send_inscribed_sat() {
 
 #[test]
 fn send_on_mainnnet_refuses_to_work_with_wallet_name_foo() {
-  let rpc_server = test_bitcoincore_rpc::builder().wallet_name("foo").build();
-  let txid = rpc_server.mine_blocks(1)[0].txdata[0].txid();
+  let rpc_server = test_bitcoincore_rpc::builder().build();
 
   CommandBuilder::new(
-    format!("wallet send ord1qcqgs2pps4u4yedfyl5pysdjjncs8et5u8gcumw {txid}:0:0"),
```
casey, line 171: We should keep this test, and test that send refuses by using rust-bitcoincore-rpc to directly create a wallet with the wrong name.

tests/wallet/create.rs
```diff
@@ -0,0 +1,83 @@
...
+    .stderr_regex(
+      "error: this does not appear to be an ord wallet, please create one using `ord wallet create`\n",
+    )
+    .expected_exit_code(1)
+    .run();
+}
+
+#[ignore]
+#[test]
+fn consecutive_create_throws_error() {}
```
casey, line 83: Good test, we should do this.

rot13maxi: +1

src/options.rs
```diff
@@ -164,6 +185,27 @@ impl Options {
...
+        .count();
+
+      let rawtr = descriptors
+        .iter()
+        .filter(|descriptor| descriptor.desc.starts_with("rawtr("))
+        .count();
+
+      if tr != 2 || descriptors.len() != 2 + rawtr {
+        bail!(
+          "this does not appear to be an ord wallet, please create one using `ord wallet create`"
```
rot13maxi, line 204: you can also label descriptors in a wallet. maybe labeling the ord descriptors and then looking for descriptors with that label would be a more reliable heuristic

raphjaph: We want to make sure that there are no other descriptors in the wallet other than the two (receive and change) taproot descriptors. This is to make sure we control the type of input in transaction construction (for better fee estimation). 

The `rawtr` descriptors store the recovery key for the commit tx in case it goes through but not the reveal tx. Those are set to inactive. I guess we should check that the `rawtr` are set to inactive.

Allowing other descriptors in the same wallet (even with different labels) would make everything a bit more messy imo.

tests/wallet.rs
```diff
@@ -696,59 +736,50 @@ fn inscribe_with_optional_satpoint_arg() {
...
-
-  CommandBuilder::new("wallet transactions")
+  CommandBuilder::new("--chain regtest wallet transactions")
     .rpc_server(&rpc_server)
     .run();
 
   rpc_server.mine_blocks(1);
 
-  CommandBuilder::new("wallet transactions")
+  CommandBuilder::new("--chain regtest wallet transactions")
```
casey, line 751: Can this be mainnet?

tests/wallet.rs
```diff
@@ -696,59 +736,50 @@ fn inscribe_with_optional_satpoint_arg() {
...
     .rpc_server(&rpc_server)
     .stdout_regex("[[:xdigit:]]{64}\t1\n")
     .run();
 }
 
 #[test]
 fn transactions_with_limit() {
-  let rpc_server = test_bitcoincore_rpc::spawn();
+  let rpc_server = test_bitcoincore_rpc::builder()
+    .network(Network::Regtest)
```
casey, line 760: Can this be mainnet?

tests/wallet/create.rs
```diff
@@ -0,0 +1,98 @@
+use super::*;
+
+#[test]
+fn create() {
+  let rpc_server = test_bitcoincore_rpc::builder()
+    .network(Network::Regtest)
```
casey, line 6: Can this be mainnet?

tests/wallet/create.rs
```diff
@@ -0,0 +1,98 @@
...
+    .rpc_server(&rpc_server)
+    .run();
+
+  assert!(rpc_server.wallets().contains("ord"));
+}
+
+#[test]
+fn wallet_creates_correct_mainnet_taproot_descriptor() {
+  let rpc_server = test_bitcoincore_rpc::builder()
+    .network(Network::Bitcoin)
```
casey, line 21: Mainnet is the default.

tests/wallet/create.rs
```diff
@@ -0,0 +1,98 @@
...
+    .rpc_server(&rpc_server)
+    .stderr_regex(
+      "error: this does not appear to be an ord wallet, please create one using `ord wallet create`\n",
+    )
+    .expected_exit_code(1)
+    .run();
+}
+
+#[test]
+fn consecutive_create_throws_error() {
```
casey, line 82: This test can be removed

src/subcommand/wallet/create.rs
```diff
@@ -1,8 +1,97 @@
...
+
+#[derive(Debug, Parser)]
+pub(crate) struct Create {
+  #[clap(long, default_value = "ord", help = "Create wallet with <NAME>")]
+  pub(crate) name: String,
+}
+
+impl Create {
+  pub(crate) fn run(&self, options: Options) -> Result {
+    if !(self.name == "ord" || self.name.starts_with("ord-")) {
```
casey, line 20: We should add a test for this, that if you pass a bad value to `create` it refuses to run.

src/options.rs
```diff
@@ -135,13 +144,45 @@ impl Options {
...
```
casey, line 183: Do we have a test for this?

casey (changes requested):

casey (approved):

PR #1167, opened 2023-01-05
Downgrade redb: 0.11.0 → 0.10.0

casey:


casey:
This didn't make a difference, so the performance regression is definitely something we did.

PR #1210, opened 2023-01-12
Load wallet automatically

raphjaph:


casey:
I don't think I understand the current state of this PR. What's the metadata extractor for? Can't we just save the currently loaded wallet on the state, and then write a test that after a command the wallet has been loaded? Or are you trying to check that it's calling API calls with the right URL?

PR #1214, opened 2023-01-13
Ignore inscriptions on a given sat after the first

### LINKED ISSUE #1128: Test behavior when a sat is reinscribed
The ord wallet will refuse to reinscribe a sat, but we should have a test that covers what happens when the index encounters an transaction that reinscribes a sat.

casey:
Fixes #1128.


casey:
Must also modify get_inscription_by_id to not return skipped inscriptions

raphjaph (approved):

PR #1234, opened 2023-01-15
Remove HTTP to HTTPS redirect implementation in favor of HAProxy

casey:
Eventually we can remove Let's Encrypt support entirely, but that's a more complicated configuration, so this is a good start.


casey:
Now thinking that we probably shouldn't merge this, since HAProxy's Let's Encrypt support isn't the best, which would prevent us from adopting it for more things.

PR #1281, opened 2023-01-18
Use a two-element array instead of a vec for change addresses

casey:
This also broke the fuzz target, so I integrated it with the rest of the workspace, so it should get checked for syntax errors on CI.


src/subcommand/wallet/transaction_builder.rs
```diff
@@ -43,17 +43,18 @@ use {
...
     outgoing_satpoint: SatPoint,
     inscribed_satpoint: SatPoint,
     inscription_id: InscriptionId,
   },
-  Dust {
-    output_value: Amount,
-    dust_value: Amount,
-  },
   ValueOverflow,
 }
```
raphjaph, line 60: Maybe Target::Exact -> Target::TryExact ?

.github/workflows/ci.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v2
```
raphjaph, line 22: What's the reason for the downgrade?

raphjaph (commented):
looks good to me. other than a couple questions

src/subcommand/wallet/transaction_builder.rs
raphjaph: Something to reflect that a non exact value can happen in some edge case like in this PR

PR #1299, opened 2023-01-19
Fuzz test building transactions with inscribed UTXOs in wallet

casey:
todo:

- [ ] Add a test that the invariant checking that inscriptions are not used as outpoints can be hit
- is inscription check corret


raphjaph:
I don't think we need this anymore

PR #1321, opened 2023-01-21
Un-reverse thumbnail row order

### LINKED ISSUE #1320: Descending chronological order for inscriptions on ordinals.com
Right now, rows of inscriptions are in ascending chronological order. That makes the latest inscription fall on the top right corner. That's a bit confusing, it would probably be better to have the latest inscription on the top left corner (ie the first thing a viewer sees, as opposed to the fourth).

casey:
Fixes #1320. I did this so that left and right arrows would be consistent with thumbnail order, but @haribosan thought it was confusing, and I'm inclined to agree, so let's revert it.

### DIFF
--- a/static/index.css
+++ b/static/index.css
@@ -170,6 +170,7 @@ a.mythic {
 
 .thumbnails {
   display: flex;
+  flex-direction: row-reverse;
   flex-wrap: wrap;
 }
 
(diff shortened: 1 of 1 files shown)


haribosan:
I think that the arrows should now be inverted

casey:
Made a separate issue for nav arrow order: https://github.com/casey/ord/issues/1327

PR #1366, opened 2023-01-25
Compress responses

### LINKED ISSUE #891: Add compression via content-encoding
We should add a field to inscriptions which contains a text string which is served with resources as the `content-encoding` header value. Users can then apply compression which should lead to a nice size reduction for lots of kinds of content. The command line tools should automatically apply compression if it results in a size decrease.

Brotli is well supported and very good, so we can use that: https://caniuse.com/?search=brotli

casey:
Fixes #891.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,6 +81,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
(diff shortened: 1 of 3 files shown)


staccDOTsol:
ah darn you already did it

https://github.com/casey/ord/pull/1926

PR #1390, opened 2023-01-26
Use JSON for all command output

casey:


src/arguments.rs
```diff
@@ -10,7 +10,7 @@ pub(crate) struct Arguments {
 }
 
 impl Arguments {
-  pub(crate) fn run(self) -> Result {
+  pub(crate) fn run(self) -> Result<Box<dyn Output>> {
```
rot13maxi, line 13: Oooooooh. we're at that stage of the project :)

rot13maxi (approved):
this is gonna be great

PR #1391, opened 2023-01-27
Use unproxied domains in justfile recipes

casey:
I configured {signet.,testnet.,}ordinals.net  to serve the explorer instances, for a few reasons:

- Cloudflare blocks SSH traffic. Having the .net domains allows us to SSH directly into our origin servers.
- Our proxied servers can't get certificates for the .com domains. Having the .net domains allows them to get .net certifcates, which allows us to configure cloudflare to use encryption when connecting to the origin servers.
- The .net domains give us a backup in case something goes wrong with cloudflare proxying.
- We can compare what's served at .com and .net domains to figure out what cloudflare is doing.

### DIFF
--- a/justfile
+++ b/justfile
@@ -30,13 +30,13 @@ deploy branch chain domain:
 
 deploy-all: deploy-testnet deploy-signet deploy-mainnet
 
-deploy-mainnet branch="master": (deploy branch "main" "ordinals.com")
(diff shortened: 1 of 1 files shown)


casey:
Signet and testnet are already proxied and seem to be working. I'll proxy .com soon.

PR #1400, opened 2023-01-27
Default preview is /favicon.png

raphjaph:


raphjaph:
Unfortunately the domain is needed in front of the path to the image or twitter won't load it. I'm unsure if it's smart to pass the domain into every page or if we should just set `https://ordinals.com` everywhere. Similar to how the docs are linked.

casey:
I think we can pass the domain into every page without too much trouble. The page takes a `Chain`, but it could take a `PageConfig` object instead:

```
struct PageConfig {
  chain: Chain,
  host: String,
}
```

So all the route functions that take a `Chain` would now take a `PageConfig`, and we wouldn't have to pass additional objects.

`favicon.png` is in static, so if you want to specifically reference it, you can do so at `/static/favicon.png`

PR #1411, opened 2023-01-29
Typo

goodwinmark:
pluralization of "sat" under Fungibility considerations

" === Fungibility considerations ===

Since any sat can be sent to any address at any time, SATS that are transferred, even those with some public history, should be considered to be fungible with other sats with no such history."


casey:
Nice, ty!

PR #1414, opened 2023-01-29
Remove redirect

casey:
So I can stick ordinals.com behind cloudflare.

### DIFF
--- a/deploy/ord.service
+++ b/deploy/ord.service
@@ -16,8 +16,7 @@ ExecStart=/usr/local/bin/ord \
   server \
   --acme-contact mailto:casey@rodarmor.com \
   --http \
-  --https \
-  --redirect-http-to-https
+  --https
 Group=ord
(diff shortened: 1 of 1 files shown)


src/media.rs
```diff
@@ -51,6 +51,7 @@ impl FromStr for Media {
 
 const TABLE: &[(&str, Media, &[&str])] = &[
   ("application/pdf", Media::Pdf, &["pdf"]),
+  ("application/pgp-signature", Media::Text, &[]),
```
cryptoquick, line 54:
```suggestion
  ("application/pgp-signature", Media::Text, &["pgp"]),
```

cryptoquick: This will fix the CI test that failed. While I've also seen .gpg used, it's usually for encrypted files.

cryptoquick (changes requested):

PR #1419, opened 2023-01-30
Add mp4 support

### LINKED ISSUE #1350: Consider supporting additional video format: H.264 codec in MP4 container
Browsers:
macOS Safari Version 16.2 (18614.3.7.1.5)
iOS Safari (whatever version is on an iPhone running iOS 16.1.2)

Inscription:
https://ordinals.com/inscription/fdcf0b5725301493f3bbb0564d9a1ecac9ab5baa0d3c6523ca77186184f3c3fai0

Video player appears but shows black background and zero second video. 

After a bit of searching, I think that this is a limitation of Safari, in which case this issue can be closed... Might be worth noting somewhere though.

cryptoquick:
[ILLICIT NFT ALERT](https://ordinals.com/inscription/2b6421c68f123e6332700133d842e14346cf2bdcda5c660465c4b3ee0f78284ei0)
Closes #1350


cryptoquick:
After running the server, I just realized why you don't like "illegal NFT" formats. I think you basically have to reindex, lol

casey:
@cryptoquick I want to do a check that the client only allows H.264 in MP4 when it uploads. If you want to look into doing that feel free, I think there's an MP4 parsing crate somewhere, otherwise feel free to leave this open and I can add that later.

You shouldn't have to reindex, illicit content types are indexed just fine, it's just the frontent that doesn't show them. However, we did change the database schema on master, so that might be what you're running into.

cryptoquick:
Can you tell why this didn't load?
https://ordinals.com/inscription/2b6421c68f123e6332700133d842e14346cf2bdcda5c660465c4b3ee0f78284ei0

casey:
Yup, `video/mp4` is an unsupported content type (not present in table in `media.rs`), so the server uses `PreviewUnknownHtml` when serving /preview, which is just an empty page.

We could add it to the `media.rs` table, but not give it a file extension. This would allow inscriptions with type `video/mp4` to be previewed, but `ord` would still not allow publishing them, because files would have an unrecognized extension.

Then later when we check the codec in `.mp4`s when inscribing, we can add the `.mp4` extension to the media table so that they can be inscribed without having to hack the client.

casey:
See the route in server.rs for `/preview`. (I think it's just `fn preview`.)

src/media.rs
```diff
@@ -63,6 +63,7 @@ const TABLE: &[(&str, Media, &[&str])] = &[
   ("text/html;charset=utf-8", Media::Iframe, &["html"]),
   ("text/plain;charset=utf-8", Media::Text, &["txt"]),
   ("video/webm", Media::Video, &["webm"]),
+  ("video/mp4", Media::Video, &["mp4"]),
```
casey, line 66:
```suggestion
  ("video/mp4", Media::Video, &[]),
```

How about this? Server can preview inscriptions with `video/mp4` content types, but can't publish them until we can do a better job of making sure they'll be viewable.

cryptoquick: That could work for now!

cryptoquick:
What I saw in preview was really weird to me. Maybe good for testing? I couldn't figure out what it was for or how to use it.

Also, just so I don't assume, why the h264 requirement? Compatibility?

casey:
> Also, just so I don't assume, why the h264 requirement? Compatibility?

Yah, that's correct. Cross-browser compatibility of containers and codecs is really bad. The reasonable choices for good compatibility are H.264 in MP4 and VP8/VP9/AV1 in WebM. MP4 is containers used with many codecs, and many of those codecs are unsupported. In particular, MP4 is often used with HEVC, which both doesn't have great support (doesn't work in firefox for example) and is patent encumbered. WebM though, even though it's a container, not a codec, is almost always used with broadly supported codecs.

A user might preview a HEVC in MP4 inscription with their browser, but then find that many other browsers won't display it after publishing it.

cryptoquick:
It's looking like adding the codec check is actually gonna be pretty easy, I should have a commit for that soon.

src/inscription.rs
```diff
@@ -44,13 +44,17 @@ impl Inscription {
...
+      .extension()
+      .ok_or_else(|| anyhow!("file must have extension"))?
+      .to_str()
+      .ok_or_else(|| anyhow!("unrecognized extension"))?;
+
+    if extension == "mp4" {
+      Media::check_mp4_codec(path)?;
+    }
+
+    let content_type = Media::content_type_for_extension(extension)?;
```
casey, line 57: Let's move these three checks into `Media::content_type_for_path`, e.g. extract the extension, check the codecs, and then return the `Media` value.

src/media.rs
```diff
@@ -90,4 +114,12 @@ mod tests {
       r"unsupported file extension `\.foo`, supported extensions: apng .*"
     );
   }
+
+  #[test]
+  fn check_mp4_codec() {
+    assert!(
+      Media::check_mp4_codec(Path::new("bitcoin-pup.mp4")).is_ok(),
```
casey, line 121: We could use a negative test too, to check that an MP4 with an unsupported video codec fails the check.

src/media.rs
```diff
@@ -33,6 +37,26 @@ impl Media {
...
+
+    let mp4 = Mp4Reader::read_header(reader, size)?;
+
+    for track in mp4.tracks().values() {
+      if let TrackType::Video = track.track_type()? {
+        if let MediaType::H264 = track.media_type()? {
+          return Ok(());
+        } else {
+          return Err(anyhow!("h264 format supported, only"));
+        }
```
casey, line 54: I have no idea what can go into an MP4, but let's check all tracks, and also put the unsupported media type in the error message:

```
      if let TrackType::Video = track.track_type()? {
        let media_type = track.media_type()?;
        if let MediaType::H264 != media_type {
          return Err(anyhow!("Unsupported video codec, only H.264 is supported in MP4s: {media_type}"));
        }
```

casey (commented):
Some comments!

src/inscription.rs
cryptoquick: Good thinking! That'll be cleaner. I'll rename the method, and do as you suggested.

src/media.rs
cryptoquick: I'll look into it.

src/media.rs
cryptoquick: I just realized, the code I had before had slightly different logic than when I explained what I wanted... I'm gonna redo some of that logic also.

cryptoquick:
@casey icymi, this pr is rfr

casey:
@cryptoquick Nice, thanks for the ping! I tweaked a little, just the tests. I moved the example files into `examples/` with descriptive filenames, and split up the tests into two with descriptive names. Also sorted the media table, and fixed a test which broke because it was expecting only valid inscriptions in `examples/`. Hope you don't mind!

Thanks for this, vv nice addition ❤️

PR #1428, opened 2023-01-31
Add avif

hashbender:
Tested with this inscription:

https://ordinals.com/inscription/b21ee1e6b643444c0c88943e175a007d689590e604991cf72c7f29cb472a304di0


cryptoquick:
AVIF is a good format, it's much more compact, so it's better on fees and make sense to add to ordinals. I'd like to also see AV1 support once that's more widely supported. Maybe it already is in browsers, actually.

cryptoquick:
Hm, not currently supported on Safari or Edge:
https://caniuse.com/av1

Similar story for AVIF, but there's a preview version of Safari:
https://caniuse.com/avif

src/media.rs
```diff
@@ -56,6 +56,7 @@ const TABLE: &[(&str, Media, &[&str])] = &[
   ("audio/mpeg", Media::Audio, &["mp3"]),
   ("audio/wav", Media::Audio, &["wav"]),
   ("image/apng", Media::Image, &["apng"]),
+  ("image/avif", Media::Image, &["avif"]),
```
casey, line 59: Let's remove the extension. This will allow inscriptions with the `image/avif` content type to display on explorers, but won't let users create them with `ord wallet inscribe FILE.avif`, since they won't have a recognized extension.

```suggestion
  ("image/avif", Media::Image, &[]),
```

This will cause some tests to break, but they can be fixed by doing this in content_type_for_extension:

```rust
    let mut extensions = TABLE
      .iter()
      .flat_map(|(_, _, extensions)| extensions.first().cloned())
      .collect::<Vec<&str>>();
```

casey (changes requested):
Nice!

My bar for browser support for new content types has been around ~95% of users, according to caniuse.com. It looks like AVIF is [currently around 80%](https://caniuse.com/avif). My worry is that people publish inscriptions that then can't be viewed by many users.

What I suggest we do is add it to the media table, but not give it an extension, see my comment inline.

cryptoquick:
@casey Only problem is that approach always breaks the MacOS test

cryptoquick:
Come to think of it, since AVIF images can be much smaller, sometimes 2-5x smaller in size, this could be a good way for fee-conscious users to save money at the expense of some compatibility. There will be gaps in the explorer and those could be awkward, but maybe those could be hidden through browser feature detection... It appears something like this code could do the detection to hide unviewable inscriptions:
https://github.com/Kagami/avif.js/blob/69b52903c1f0d713e9a5082daac162b377b6ad83/avif.js#L66-L69

casey (approved):
Looks good, but I removed the `.avif` extension, I'm still not sure about allowing formats that might preview in the user's local browser, but not for others.

casey:
See #1521 for enabling this all the way.

Crazy to think that the more efficient we make inscriptions, the more efficiently everyone can use the chain, including normal bitcoin transactions 🤯

PR #1449, opened 2023-02-01
Render JSON and YAML inscriptions as text

casey:
People are publishing YAML and JSON documents as text/plain. Not sure if they should get special treatment (syntax highlighting?) but at least we should recognize the content types so people can use the right content types.


src/media.rs
```diff
@@ -50,6 +50,8 @@ impl FromStr for Media {
 }
 
 const TABLE: &[(&str, Media, &[&str])] = &[
+  ("application/json", Media::Text, &["json"]),
+  ("application/yaml", Media::Text, &["yaml"]),
```
windsok, line 54: `.yml` is a sometimes used file extension too

casey: Seems reasonable to me! Added.

casey:
@windsok How comfortable are you reviewing the code? I've been operating with the following policy for merges:

- If someone contributes code, I review it and merge it
- If I write code, @raphjaph reviews it before I merge it

This code I've written, so I'd like to get another review before merging. This is very simple, but it's nice to sanity check, and make sure my reasoning for a feature makes sense.

- Does this solve the problem that's trying to be solved?
- Are there tests?
- Is this going to be disruptive somehow, etc.

For this PR, since you've taken a look, do you feel comfortable giving an ack? Ideally you should understand the code and what it's doing, and any implications, and if you don't you can ask me questions, even very simple or basic ones, and I'll explain. No pressure if you're busy or not interested!

windsok (commented):
ACK e286752956b98411b4e7030afb48834e5ea7fda9

I understand this change fairly well as I might have hacked in a illegitimate inscription or two 😉

Rendering these as text makes sense. I guess could add some syntax highlighting in the future but this should let people start inscribing their collection JSON etc with the correct mime type

windsok (approved):

casey:
Thank you!

Also, everything is tested, so when a PR, like this, has a functional change but not a corresponding new test or change to an existing test, you should be suspicious!

In this case, it's okay. So far, we haven't been testing "data", things where really we would just be testing a constant. In this case, this table feels like data, so although we've tested the functions that access this table, we haven't tested individual table entries, which is why this change doesn't have a test.

PR #1450, opened 2023-02-01
add a delay flag to wait for the commit before broadcasting the reveal

### LINKED ISSUE #1446: Optionally separate broadcast commit tx and following tx
Even just by showing tx hex in the `--dry-run` would be helpful to achieve this.

This would avoid an observer of the mempool copying the inscription in another tx and with a greater fee possibly having an older genesis tx than the original creator.

rot13maxi:
adds a `--delay` flag to the `wallet inscribe` command. If it's passed, it will:

- write the serialized reveal transaction to a file (incase the next step crashes or is cancelled)
- after the commit transaction is broadcast, will wait until it has at least one confirmation before broadcasting the reveal transaction.

close #1446


casey:
Sorry for the delay!

I think this should be split into a few PRs. Don't feel like you need to do any/all of them:

1. The first should provide a general way for backing up commit and reveal transactions: We could put all commit and reveal transactions in the data dir, in a `transactions` or `txs` directory. We could do this automatically, or only when the user passes a flag. This would let users inspect/rebroadcast all transactions, and might give them options after the fact when something goes wrong. As far as format I like binary, since saving hex in a file seems like an extra encoding, and we could also provide tools for dumping things in the txs directory. (`--dry-run` shouldn't save transactions because the user doesn't intend to broadcast them.)

2. Some kind of `--nobroadcast` flag, which in combination with saving txs, would let users have very flexible strategies about commit/reveal timing. They could in a one-liner broadcast a saved commit tx, wait for it to confirm manually, and then manually broadcast the reveal transaction.

3. This PR, which I think is the common case so great to make easy. This needs a test. It should check that when the flag is passed the commit and reveal show up in subsequent blocks instead of the same block.

casey (changes requested):
See comment!

casey:
I actually think my previous comments were silly. Adding a --delay flag and waiting is fine and simple, even if it's not perfectly optimal. It's old, but if you feel like rebasing and adding a test, feel free to do so and ping this PR.

PR #1456, opened 2023-02-01
Lazily load iframes

casey:
This doesn't seem to fix the problem, so it's a draft.


casey:
I can't test how well this works without deploying it somewhere, and reducing load on the server is always good, and I'm pretty sure that we're using this correctly, so I think this is safe to merge. In general, requests for inscription content, which is always inside iframes, are the highest volume, slowest requests, so anything that might reduce their number, like not requesting off-screen iframes is great.

PR #1463, opened 2023-02-02
nix: make it such that cargo test will run on nix

niftynei:
Add missing 'pkg-config openssl' so nix can run `cargo test`


casey:
NIFTY! Nice, thank you for the PR, merging ^_^. Just added a commit sorting the deps alphabetically.

PR #1466, opened 2023-02-02
Log recoverable updater errors as warnings

casey:


src/index/updater.rs
```diff
@@ -229,7 +229,7 @@ impl Updater {
 
           errors += 1;
           let seconds = 1 << errors;
-          log::error!("failed to fetch block {height}, retrying in {seconds}s: {err}");
+          log::warning!("failed to fetch block {height}, retrying in {seconds}s: {err}");
```
windsok, line 232: I think this should be `log::warn`

casey: Perfect, fixed. Assume it's good to go otherwise? Are there any recoverable errors that should be warnings?

raphjaph (approved):
Looks good, wanted to do it like this as well

casey:
Created #1528 to track any stragglers. @raphjaph this is good to merge!

PR #1473, opened 2023-02-02
Add moderation guide

casey:


casey:
@VirtuallyThere @batcavekid Added a moderation guide, please take a look and let me know if it looks good! I try to do one additional review for all PRs, so feel free to ack so I can merge. Obvs might relax this policy if something is super urgent, but seems good to do so I don't do something stupid.

windsok (approved):

windsok:
Looks good to me as an initial guide. Are there any other config options that should be documented?

docs/src/guides/moderation.md
```diff
@@ -0,0 +1,50 @@
...
+
+In order to prevent particular inscriptions from being displayed on an `ord`
+instance, they can be included in a YAML config file, which is loaded with the
+`--config` option.
+
+To hide inscriptions, first create a config file:
+
+```yaml
+hidden:
+- 0000000000000000000000000000000000000000000000000000000000000000i0
```
windsok, line 21: it's probably obvious to most people, but we could be explicit that this is the inscription id

casey:
Awesome, thanks for reviewing! The only config option right now is hidden. It was an emergency patch once the first butthole appeared. Didn't know the ordinals logo would be so prophetic ◉

Issue #1487, opened 2023-02-03
Allow setting higher fee rate for commit transaction separately


andrewtoth:
Option `--fee-rate` sets the fee for both commit and reveal txs. However, we might want our commit tx to confirm quickly and not want to pay the higher fee for the entire reveal. Perhaps have a separate `--commit-fee-rate` option that will set the commit tx fee rate to that? The `--fee-rate` option (or its default) will be used for both if `--commit-fee-rate` is not set, and only for the reveal tx if `--commit-fee-rate` is set.

This can then be leveraged with `--delay` (https://github.com/casey/ord/issues/1446) and we can add fee bumping by replacing the inscription with the same file but only bumping the commit tx. Or replacing with a different file if we change our minds before confirmation, or canceling altogether before confirmation by replacing the commit with a tx back to ourselves or another destination.

In fact it can work without `--delay` but make feebumping/replacement much cheaper also. This way we can still make the commit tx pay for the entire commit plus reveal plus 1 sat/vbyte to RBF them both, but then still have the reveal tx pay the smaller amount.

PR #1490, opened 2023-02-03
Add commit-fee-rate option for specifying individual fee rates

andrewtoth:
Closes https://github.com/casey/ord/issues/1487.

### DIFF
--- a/src/subcommand/wallet/inscribe.rs
+++ b/src/subcommand/wallet/inscribe.rs
@@ -35,6 +35,11 @@ pub(crate) struct Inscribe {
     help = "Use fee rate of <FEE_RATE> sats/vB"
   )]
   pub(crate) fee_rate: FeeRate,
+  #[clap(
+    long,
+    help = "Use fee rate of <COMMIT_FEE_RATE> sats/vB for commit transaction only.\nIf not set uses fee-rate for commit transaction."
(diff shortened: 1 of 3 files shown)


src/subcommand/wallet/inscribe.rs
```diff
@@ -45,6 +50,14 @@ pub(crate) struct Inscribe {
 
 impl Inscribe {
   pub(crate) fn run(self, options: Options) -> Result {
+    if let Some(commit_fee_rate) = self.commit_fee_rate {
```
casey, line 53: I think this check should be removed. There are actually reasons my one might want to use a low fee rate for the commit transaction. You could use a low fee on commit and a high fee on reveal, to encourage miners to include both or neither, using CPFP.

casey (commented):
Wasn't able to do a full review yet, sorry for the delay! But wanted to note that we should probably allow commit-fee-rate to be less than fee-rate. See comment!

src/subcommand/wallet/inscribe.rs
andrewtoth: Removed

casey:
This looks great! I pushed f04eb06 to make the help string shorter, hopefully it looks okay!

PR #1492, opened 2023-02-03
Support STL inscriptions

casey:
Add `model/stl` with extension `.stl` to `Media::TABLE` with type `Media::Unknown`. This will allow using`.stl` files with `ord wallet inscribe`. However, they will not have any preview, since the media type is still unknown.

A follow-up PR can add some kind of rendering/preview support.

### DIFF
--- a/src/media.rs
+++ b/src/media.rs
@@ -98,6 +98,7 @@ const TABLE: &[(&str, Media, &[&str])] = &[
   ("image/png", Media::Image, &["png"]),
   ("image/svg+xml", Media::Iframe, &["svg"]),
   ("image/webp", Media::Image, &["webp"]),
+  ("model/stl", Media::Unknown, &["stl"]),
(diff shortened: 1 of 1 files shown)


rot13maxi:
woah. That's gonna open some doors

rot13maxi (approved):
looks sane

casey:
Nice, thanks for the review!

joe-rodgers:
I uploaded a .stl inscription and the content download outputs a .bin file, not a .stl. I am regarded and it took me all day to figure out to change the file name to .stl in windows explorer and it worked like a charm. (thanks Goodwin for the idea).

Not sure if it would be possible to append the .stl filetype to the file name or something like that so it is more intuitive.

https://ordinals.com/inscription/423b8403b8250bf226dec3d2ed342d971491ca8aaeaaab9d5d628d7dc744e2adi0

Issue #1471, opened 2023-02-02
Cannot run ord server after a crash due to cberner/redb#515


hantuzun:
A bug in the underlying db (https://github.com/cberner/redb/issues/515) is preventing ord to run server after a crash.

I interrupted ord as it was indexing and seemed stuck at block 494299:

```
th 0, needed 12.
[indexing blocks] ████████████████████░░░░░░░░░░░░░░░░░░░░░░ 384299/774744[2023-02-02T17:10:08Z ERROR ord::index::updater] failed to fetch block 384332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] █████████████████████░░░░░░░░░░░░░░░░░░░░░ 404299/774744[2023-02-02T17:25:51Z ERROR ord::index::updater] failed to fetch block 404332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] ██████████████████████░░░░░░░░░░░░░░░░░░░░ 409299/774744[2023-02-02T17:30:41Z ERROR ord::index::updater] failed to fetch block 409332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] ██████████████████████░░░░░░░░░░░░░░░░░░░░ 414299/774744[2023-02-02T17:35:18Z ERROR ord::index::updater] failed to fetch block 414332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] ██████████████████████░░░░░░░░░░░░░░░░░░░░ 419299/774744[2023-02-02T17:40:51Z ERROR ord::index::updater] failed to fetch block 419332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] ███████████████████████░░░░░░░░░░░░░░░░░░░ 424299/774744[2023-02-02T17:45:45Z ERROR ord::index::updater] failed to fetch block 424332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] ███████████████████████░░░░░░░░░░░░░░░░░░░ 429299/774744[2023-02-02T17:50:27Z ERROR ord::index::updater] failed to fetch block 429332, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[...]

PR #1513, opened 2023-02-05
Fix corrupted databases by using redb 0.13.0

hantuzun:
This PR fixes https://github.com/casey/ord/issues/1471 by bringing the fix for https://github.com/cberner/redb/issues/515.


casey:
Excellent, thank you for identifying the issue, working with @cberner, and opening this PR with the fix! Merging now.

Why did the lifetimes change? Everything compiles and the tests passes, so it's fine, I just don't understand the what caused the change.

cberner:
I changed redb to only accept `'static` lifetimes for the keys and values. Those lifetimes have no meaning because all the `get()`, `insert()`...etc functions generate their own lifetimes based on `&self` via the `SelfType` GAT

Issue #1502, opened 2023-02-04
Use p2p for fetching blocks from bitcoind instead of rpc


andrewtoth:
Fetching raw blocks via RPC is much slower than via p2p if the node is listening.
The RPC interface requires parsing the json request, fetching the block form disk and deserializing it into `CBlock`, serializing it into hex, then putting it into json and sending over the wire. The client also has to deserialize the json response and then the block from hex.
The RPC interface also requires authentication and is rate limited. `rpcworkqueue` and `rpcthreads` should be configured for more reliable indexing. 
P2P requires `listen=1` to be set (it is default but some configurations set it to `listen=0` by defaults, like when setting `connect`), and requires a handshake before serving. After the handshake, blocks are fetched directly from disk without deserializing and sent across the wire in binary form. It requires no authentication and will not fall over from exceeding work queue.
See https://github.com/romanz/electrs/blob/master/src/p2p.rs.

andrewtoth:
Using REST is also a more suitable application for this than RPC. It requires no authentication and can transmit blocks in binary form. It also does not have to parse a json request. It does requires `rest=1` in config though. However, it still deserializes a block to `CBlock` internally before sending. Ideally https://github.com/bitcoin/bitcoin/pull/26415 would be merged and this would be by far the fastest way to fetch blocks.
See https://github.com/andrewtoth/rust-bitcoincore-rest.

Issue #1377, opened 2023-01-26
RPC error: transport error: HTTP response too short: length 0, needed 12


casey:
> While indexing, I'm getting errors like " failed to fetch block 705096, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12." exactly every 5000 blocks.  Seems to continue though.  Is this a known issue?

dude513:
Exactly every 5000 blocks, me as well.
Started somewhere in a 200000 range, I can't recall.

`[indexing blocks] ......... 333974/774477`

`ERROR ord::index::updater] failed to fetch block 334007, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.`

nitronick600:
It's due to the commit that happens in updater.rs 

Not sure exactly how rust works, but my guess would be the sync channel is blocked waiting for the `loop` to read off of it and the client ends up getting hung when attempting to write new block data

dude513:
It's indexing slower now and with more frequent errors. @casey

![image](https://user-images.githubusercontent.com/123891835/215961662-cc5310bc-eccc-49bd-bd16-77731e85f68f.png)

60000 blocks in 12 hours.

nitronick600:
It's very write-heavy - using an NVME will speed things up a lot

OrpingtonClose:
> It's very write-heavy - using an NVME will speed things up a lot

used a ramdisk to do a ord index but it is similarily slow

dude513:
@hashbender it has nothing to do with disks, I am on SSD.

I am confused why it is not the highest priority now because ord is not usable.

I tried to roll back to different versions and with `0.4.0` I got almost all blocks within a few minutes, however below this version it throws some weirdness when trying to `ord index`.

We know that there is a bug somewhere between `0.4.0` and `0.4.2`.

Even though I got almost all blocks (10000 left) now it is stuck with any version.
It throws an error every single block.

```
[indexing blocks] █████░░░ 764050/774732[2023-02-02T15:45:07Z ERROR ord::index::updater] failed to fetch block 764083, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[indexing blocks] █████░░░ 764051/774732[2023-02-02T15:47:01Z ERROR ord::index::updater] failed to fetch block 764084, retrying in 2s: JSON-RPC error: transport error: HTTP response too short: length 0, needed 12.
[...]

Issue #1364, opened 2023-01-25
Re-enable skipping transactions below first inscription height


casey:
I'm not sure how to do this. We'd need to get UTXO set values when we hit inscriptions. Maybe we only need UTXO values for transactions that contain inscriptions, in which case maybe the impact of fetching input transactions from bitcoin core is low.

We can also use the new verbosity level of getrawtransaction to return the value of all inputs in one go: https://github.com/bitcoin/bitcoin/issues/23319

raphjaph:
- Check wether `getrawtransaciton --verbose` is in newest Bitcoin Core release
- Start with PR agains rust-bitcoincore-rpc

raphjaph:
Waiting for new release of Bitcoin Core

andrewtoth:
Could we not just use `getblock` with `verbosity=3` to get the blocks with all input values?

casey:
> Could we not just use `getblock` with `verbosity=3` to get the blocks with all input values?

I suspect that this would be really slow. `verbosity=3` includes a lot of extra data. However, it might be good to benchmark this and make sure it's the case.

andrewtoth:
Also, if we are going the route of calling `getrawtransaction` individually vs one `getblock verbosity=3`, then we should probably be using `async`. We can then hit the REST endpoint concurrently with a `join_all` instead of sequentially. We could potentially batch the `getrawtransaction` calls but `rust-bitcoincore-rpc` doesn't support the batching feature :/.

andrewtoth:
Ok I figured it out. Managed to sync by combining p2p and REST and reverting #1357 in 1:06. I think we can turn this back on :).

Batching calls would overwhelm bitcoind, and `getblock verbosity=3` is slower than just synchronously fetching missing inputs via REST.

andrewtoth:
With latest REST PR using hyper myself on Linux and @so7ow on Mac synced in under an hour 🚀 https://github.com/casey/ord/issues/1648#issuecomment-1427946904.

andrewtoth:
Check out https://github.com/casey/ord/pull/1759. It still hasn't been explained why skipping transactions were removed, but that PR manages to sync efficiently without having all inputs.

PR #1516, opened 2023-02-06
Use p2p to fetch blocks and headers

andrewtoth:
Uses the p2p interface to fetch block hashes and headers much more efficiently. Fetches headers 2k at a time, and fetches blocks in chunks of 10.

Syncs the first 200k or so blocks much faster, but haven't benchmarked the whole chain. I don't suspect it will be much more gain for the whole chain since most of the time will be spent indexing to the db, and blocks are already being fetched in the background. However, this is a much more efficient way of fetching this data.

Still have to fix tests because they don't support connecting via p2p yet.

Closes https://github.com/casey/ord/issues/1502.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -945,6 +945,8 @@ mod tests {
         "ord".into(),
         "--rpc-url".into(),
         rpc_server.url().into(),
+        "--p2p-port".into(),
+        rpc_server.p2p_port().into(),
         "--data-dir".into(),
         tempdir.path().into(),
         "--cookie-file".into(),
--- a/src/index.rs
+++ b/src/index.rs
@@ -16,9 +16,11 @@ use {
   redb::{Database, ReadableTable, Table, TableDefinition, WriteStrategy, WriteTransaction},
   std::collections::HashMap,
   std::sync::atomic::{self, AtomicBool},
+  url::Url,
 };
 
 mod entry;
+mod p2p;
 mod rtx;
 mod updater;
 
@@ -43,7 +45,6 @@ define_table! { STATISTIC_TO_COUNT, u64, u64 }
 define_table! { WRITE_TRANSACTION_STARTING_BLOCK_COUNT_TO_TIMESTAMP, u64, u128 }
 
 pub(crate) struct Index {
-  auth: Auth,
   client: Client,
   database: Database,
   path: PathBuf,
@@ -52,7 +53,8 @@ pub(crate) struct Index {
   genesis_block_coinbase_txid: Txid,
   height_limit: Option<u64>,
   reorged: AtomicBool,
-  rpc_url: String,
+  p2p_url: String,
+  network: Network,
 }
 
 #[derive(Debug, PartialEq)]
@@ -142,7 +144,7 @@ impl Index {
 
     let auth = Auth::CookieFile(cookie_file);
 
-    let client = Client::new(&rpc_url, auth.clone()).context("failed to connect to RPC URL")?;
+    let client = Client::new(&rpc_url, auth).context("failed to connect to RPC URL")?;
 
     let data_dir = options.data_dir()?;
 
@@ -229,9 +231,20 @@ impl Index {
     let genesis_block_coinbase_transaction =
       options.chain().genesis_block().coinbase().unwrap().clone();
 
+    let p2p_url = match Url::parse(&rpc_url) {
+      Ok(url) => url.host().unwrap().to_string(),
+      Err(_) => rpc_url
+        .split(':')
+        .collect::<Vec<_>>()
+        .first()
+        .unwrap()
+        .to_string(),
+    };
+
+    let p2p_url = format!("{}:{}", p2p_url, options.p2p_port());
+
     Ok(Self {
       genesis_block_coinbase_txid: genesis_block_coinbase_transaction.txid(),
-      auth,
       client,
       database,
       path,
@@ -239,7 +252,8 @@ impl Index {
       genesis_block_coinbase_transaction,
       height_limit: options.height_limit,
       reorged: AtomicBool::new(false),
-      rpc_url,
+      p2p_url,
+      network: options.chain().network(),
     })
   }
 
--- a/src/index/p2p.rs
+++ b/src/index/p2p.rs
@@ -1,0 +1,183 @@
+use {
+  anyhow::{bail, ensure, Context, Result},
+  bitcoin::{
+    consensus::{encode, Decodable},
+    hashes::Hash,
+    network::{
+      constants,
+      message::{NetworkMessage, RawNetworkMessage},
+      message_blockdata::{GetHeadersMessage, Inventory},
+      message_network::VersionMessage,
+      Address,
+    },
+    secp256k1::{self, rand::Rng},
+    Block, BlockHash, BlockHeader, Network,
+  },
+  log::{debug, trace, warn},
+  std::{
+    io::{ErrorKind, Write},
+    net::{IpAddr, Ipv4Addr, SocketAddr, TcpStream},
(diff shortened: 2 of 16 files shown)


andrewtoth:
Might fix issues related to https://github.com/casey/ord/issues/1377, since those look like the RPC threads are falling over from too many requests.

I haven't fully groked the indexing logic for fetching raw transactions, but if we are doing a lot of `getrawtransaction` RPC calls it might be worth using REST interface for that. Benchmarking a `getrawtransaction` RPC vs a REST binary request for a large tx shows REST request is about 2x faster.

casey:
This is awesome! Keep in mind that https://github.com/casey/ord/issues/1364 will wind up making sync faster.

Some thoughts!

1. I don't want to make things much more complicated, so keeping an initial PR as simple and focused as possible would be great.
2. We need to make sure that they must use a trusted BTC node, since ord doesn't enforce consensus.
3. All our tests run against our own mocked bitcoind, which is in `test-bitcoincore-rpc`. This is required to keep tests fast. Spinning up a bitcoind instances is slow, and tests are in our edit-test-debug cycle, so they must stay fast. As long as they stay fast and comprehensive, the method we use for testing doesn't matter, but one way to keep them fast would be to add P2P protocol support to `test-bitcoincore-rpc`. It only needs the minimum for the tests to work, so doesn't have to have comprehensive support for the full protocol.
4. I didn't even know that bitcoind had a rest interface. Is it documented anywhere? How is it different from the JSON RPC interface? I love rest and prefer it over JSON RPC now that I've had experience with both.

andrewtoth:
Thanks @casey.
1. Sorry for the REST tangent, that can be a follow-up.
2. Right. I suppose we can just use the same address as rpc and just specify the p2p port as an option.
3. Ok, will give that a shot.
4. REST is documented [here](https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md). I have an implementation that we can pull in as a cargo crate [here](https://github.com/andrewtoth/rust-bitcoincore-rest). But, that uses `async`/`await` so not sure if it will be useful in this project. REST is unauthenticated and doesn't have to go through the json parsing, so is much faster. It can also reply with raw binary data which is faster to serialize and smaller to send over the wire. But it also only shares public data, so we can't use it for things like the wallet. If https://github.com/bitcoin/bitcoin/pull/26415 was merged then I think that would be an even better candidate to fetch blocks than p2p, but the requirement to deserialize the block before serializing back into binary makes it slower than using the p2p endpoint right now. We can query it for utxo set data as well, but not sure what your ideas are for #1364.

src/index.rs
```diff
@@ -228,17 +230,29 @@ impl Index {
...
 
+    let p2p_url = match Url::parse(&rpc_url) {
+      Ok(url) => url.host().unwrap().to_string(),
+      Err(_) => rpc_url
+        .split(':')
+        .collect::<Vec<_>>()
+        .first()
+        .unwrap()
+        .to_string(),
+    };
```
andrewtoth, line 241: Not sure what the best way to parse the url is. By default it is `127.0.0.1:8332/wallet/ord`, which doesn't parse as a url, and in tests it is `http://127.0.0.1:...` which won't parse with a `split(':')`. Maybe regex?

andrewtoth:
@casey I think this is ready for a closer look.

src/index/p2p.rs
```diff
@@ -0,0 +1,182 @@
...
+  let services = constants::ServiceFlags::NONE;
+
+  NetworkMessage::Version(VersionMessage {
+    version: constants::PROTOCOL_VERSION,
+    services,
+    timestamp,
+    receiver: Address::new(&addr, services),
+    sender: Address::new(&addr, services),
+    nonce: secp256k1::rand::thread_rng().gen(),
+    user_agent: format!("/ord:{}/", "0.4.2"),
```
andrewtoth, line 178: This would ideally pull the version string from Cargo.toml.

casey:
@andrewtoth Awesome, thank you! Putting it on the [project backlog](https://github.com/users/casey/projects/3) so it doesn't fall off our radar.

jpistone:
WOW thank you! I was ticking <1 block per second. Found this and pulled your changes, now at ~50-100 per second. This should definitely be a top priority as most questions on the discord are about this right now (3 days - 2 weeks to index is rough).

Quick note: commit seems to happen every 5k blocks & takes about 3 minutes on my machine (I believe that's what was causing the json-rpc errors before this patch). This may confuse users when they see the block count hang for some time - I killed & restarted indexing myself a few times thinking that it had frozen.

image

dude513:
After 16 hours the index is at block `420001` and moving very slowly, no errors/crashes yet.

I am on `p2p-indexing` branch git cloned from `andrewtoth/ord`. @andrewtoth

andrewtoth:
This has got a lot of successful testing from users in #1648.

dude513:
> This has got a lot of successful testing from users in #1648.

With `rest=1` and on a `speedup-improvements` branch I got 99% of all blocks in a few minutes.
I guess last 9000 blocks will be the most satisfying. :)

andrewtoth:
One user had set `maxuploadtarget` to something too low in their `bitcoin.conf`, and so this disallowed the p2p interface from serving everything it needs to. We should advise users to add `whitebind=127.0.0.1:8333` to their config files.

veryordinally:
@casey We should definitely consider this, did an initial review and test.

veryordinally (commented):
Reviewed conceptually and high-level flow and logic - looks good to me and ready for more detailed review.

casey:
I'm sorry for taking so long to collect my thoughts on this! Bitcoin's RPC and P2P interfaces are very different. The P2P interfaces is an untrusted interface used between full nodes on the bitcoin network. The RPC interfaces is an authenticated interface between a client and a trusted node. Using the P2P interface for applications where the node must be trusted is dicey. You can use it to directly connect to a trusted node, but you can also use it to connect to a totally untrusted node. I think that expressing this subtle difference to our users would be hard, and that lots of people, because running a full node is challenging, would try to connect to some random node, and have their security degraded as a result.

Also #1364 fixes the same issue as this would, namely, speeding up initial sync, and is a very clean and low-impact chain that doesn't introduce another connection to manage. So, I would like to see that implemented first, before deciding if also using P2P or REST is a good idea.

Also, keeping ord simple to configure is important (The degens are dying out there 😆) and every new outgoing connection introduces code, configuration, and  maintainence complexity.

Sorry for not figuring this out sooner, it's only been in the last few days that I've been able to sit down and review PRs in detail, and previously I didn't give it enough attention.

Let's make sure we nail down the following in an issue before anyone writes a big PR:

- What is the problem we're trying to solve or the benefit we're trying to eliminate?
- What is the best design?
- What is the testing strategy?

Sorry to close! We're still figuring out this whole massive open source project thing ourselves!

andrewtoth:
@casey disappointing conclusion. I disagree with your reasoning.

This requires no other configuration options. It works out of the box. It's been tested by many users already without problems. The only configuration issue was someone who had maxuploadtarget already set, which is obviously an advanced configuration option.
This patch would require using the same node as the RPC connection. It cannot be used to connect to a different node because only the port is configurable, not the host.
It is much more efficient for all the reasons stated in the initial issue #1502 . Blocks are fetched directly from disk without serialization steps, and can be requested in batches. It puts less stress on bitcoind and is faster.
Without this change, #1364 would be making RPC requests on multiple threads. This can overload the RPC workqueue and cause disconnections.
You can see the issues with fetching blocks via RPC in the updater code. There is a lot of code that has retries, and users are complaining that it aborts after retrying forever.

kalibox:
Does anyone know how much space ord index files take and where they're located?

PR #1524, opened 2023-02-06
Make all wallet commands load wallet

raphjaph:


casey:
Let's add a `Wallet` struct that can only be constructed by loading the wallet:

```rust
struct Wallet {
  foo: (),
}

impl Wallet {
  fn new(options: &Options) -> Result<Self> {
    options.bitcoin_rpc_client_for_wallet_command(false)?;

    Ok(Self)
  }
}
```

casey (changes requested):
see comment

casey (changes requested):
PR reviewed in the salt mines

tests/lib.rs
```diff
@@ -63,10 +63,19 @@ struct Create {
   mnemonic: Mnemonic,
 }
 
-fn create_wallet(rpc_server: &test_bitcoincore_rpc::Handle) {
-  CommandBuilder::new(format!("--chain {} wallet create", rpc_server.network()))
+fn create_wallet(rpc_server: &test_bitcoincore_rpc::Handle, wallet_name: Option<String>) {
```
casey, line 66: Can we always use the name ord? Is the old test actually testing anything? If not, it can be removed.

casey (changes requested):
Minor nits, see comments.

casey (approved):
Excellent!

Issue #1399, opened 2023-01-27
Support `rpc_user` and `rpc_pass`


garyray-k:
According to an ask in the Discord, ord uses a cookie for RPC authentication. Please support `rpc_user` and `rpc_pass` to enable easier (non-tech friendly) remote connections to bitcoin core.

raphjaph:
We added this check out PR https://github.com/casey/ord/pull/1527

PR #1527, opened 2023-02-06
Allow rpc_user and rpc_pass in config file

raphjaph:
#1399 

Probably closes:
#1785
#1562


raphjaph:
Let's not support this as flags but as a option in the config file

bliotti:
Probably need to change Title "Instead of"?  
Was worried we were dropping  .cookie

bliotti:
Tested [37d36ea](https://github.com/casey/ord/pull/1527/commits/37d36ea00d0607ed1c52a790423e265bf4345baa)  , both methods work.

**However `rpcuser` and `rpcpassword` are deprecated.** 

```
2023-02-06 21:34:02 bitcoin.conf file exists, skipping.
2023-02-06 21:34:02 2023-02-07T02:34:02Z Bitcoin Core version v24.0.1 (release build)
2023-02-06 21:34:02 2023-02-07T02:34:02Z Signet derived magic (message start): 0a03cf40
2023-02-06 21:34:02 2023-02-07T02:34:02Z Using the 'sse4(1way),sse41(4way),avx2(8way)' SHA256 implementation
2023-02-06 21:34:02 2023-02-07T02:34:02Z Using RdSeed as an additional entropy source
2023-02-06 21:34:02 2023-02-07T02:34:02Z Using RdRand as an additional entropy source
2023-02-06 21:34:02 2023-02-07T02:34:02Z Default data directory /root/.bitcoin
2023-02-06 21:34:02 2023-02-07T02:34:02Z Using data directory /root/.bitcoin/signet
2023-02-06 21:34:02 2023-02-07T02:34:02Z Config file: /root/.bitcoin/bitcoin.conf
2023-02-06 21:34:02 2023-02-07T02:34:02Z Config file arg: rpcpassword=****
2023-02-06 21:34:02 2023-02-07T02:34:02Z Config file arg: rpcuser=****
2023-02-06 21:34:02 2023-02-07T02:34:02Z Config file arg: server="1"
2023-02-06 21:34:02 2023-02-07T02:34:02Z Config file arg: signet="1"
2023-02-06 21:34:02 2023-02-07T02:34:02Z Config file arg: txindex="1"

2023-02-06 21:34:02 2023-02-07T02:34:02Z Config options rpcuser and rpcpassword will soon be deprecated. Locally-run instances may remove rpcuser to use cookie-based auth, or may be replaced with rpcauth. Please see share/rpcauth for rpcauth auth generation.

2023-02-06 21:34:02 2023-02-07T02:34:02Z [http] starting 4 worker threads
2023-02-06 21:34:02 2023-02-07T02:34:02Z Using wallet directory /root/.bitcoin/signet/wallets
2023-02-06 21:34:02 2023-02-07T02:34:02Z init message: Verifying wallet(s)…
2023-02-06 21:34:02 2023-02-07T02:34:02Z Using /16 prefix for IP bucketing
```

raphjaph:
Yea, using rpc_user and rpc_pass should not be encouraged

andrewtoth:
> rpcuser and rpcpassword will soon be deprecated. 

They will not be deprecated from use in clients. The deprecation refers to using bare `rpcuser/rpcpassword` in Bitcoin Core's `bitcoin.conf`. When configuring bitcoind you should not use `rpcuser/rpcpassword` but `rpcauth`. For clients `rpcuser/rpcpassword` will never be deprecated, how else would they authenticate without access to the cookie file?

raphjaph:
> They will not be deprecated from use in _clients_. The deprecation refers to using bare `rpcuser/rpcpassword` in favor of generating an `rpcauth` with the user and password in Bitcoin Core's `bitcoin.conf`. When configuring bitcoind you should not use `rpcuser/rpcpassword` but `rpcauth`, which does not expose the password in plaintext in the config file. For clients `rpcuser/rpcpassword` will not be deprecated.

Yes, true. Passing credentials as command line flags is still not ideal because those will be stored in plaintext in your bash history. I've add it as flags anyway because that fear is probably overblown.

raphjaph:
`rpc_user` and `rpc_pass` can now be passed in to `ord` in a plethora of ways. In order of precedence:
- in the `ord.yaml` config file
- as environmental variables
- as command line options
- through the cookie file

Generally I think we have to overhaul or configuration setup. I think there should be global `Config` or `Options` struct maybe initialized with `lazy_static!` à la [this](https://www.philipdaniels.com/blog/2019/rust-configuration-example/) method. We were creating rpc clients in different places in `index.rs` that didn't go through or `Option` methods. These were verifying that the client is and Bitcoin Core are the correct version, etc. This PR also contains some `.clone()`s in places that might be inefficient.

Sjors:
In addition to these methods, you could also read `rpcuser` and `rpcpassword` from `bitcoin.conf`, like `bitcoin-cli` does. For example a user might want to run their `bitcoind` under a different user than `ord`, in which case it won't have access to the cookie file, but you can create a `bitcoin.conf` file with the RPC credentials. (this isn't super useful in practice because Bitcoin Core doesn't have a way yet to limit RPC permissions)

raphjaph:
> In addition to these methods, you could also read `rpcuser` and `rpcpassword` from `bitcoin.conf`, like `bitcoin-cli` does. For example a user might want to run their `bitcoind` under a different user than `ord`, in which case it won't have access to the cookie file, but you can create a `bitcoin.conf` file with the RPC credentials. (this isn't super useful in practice because Bitcoin Core doesn't have a way yet to limit RPC permissions)

I think `rpcuser/rpcpass` from `bitcoin.conf` is being deprecated in favor of `rpcauth`. This will only store the username, password hash and salt.

Sjors:
Assuming a system with two users: `bitcoin` and `ord`. It makes sense for the `bitcoin` user to use `rpcauth` so only the hash is stored. But the `ord` user would use `rpcpass`. It's a bit confusing that both `bitcoind` and `bitcoin-cli` use the same config file. I don't think this will stop working anytime soon.

src/options.rs
```diff
@@ -137,24 +141,45 @@ impl Options {
...
-
     let rpc_url = self.rpc_url();
 
-    log::info!(
-      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
-      cookie_file.display()
-    );
+    let config = self.load_config().unwrap();
+
+    let auth = if let (Some(rpc_user), Some(rpc_pass)) = (config.rpc_user, config.rpc_pass) {
```
casey, line 148: For context:

https://stackoverflow.com/questions/32272911/precedence-of-configuration-options-environment-registry-configuration-file-a

src/options.rs
```diff
@@ -137,24 +141,45 @@ impl Options {
...
+      Auth::UserPass(rpc_user, rpc_pass)
+    } else if let (Some(rpc_user), Some(rpc_pass)) = (self.rpc_user.clone(), self.rpc_pass.clone())
+    {
+      log::info!(
+        "Connecting to Bitcoin Core RPC server at {rpc_url} using rpc_pass for `{rpc_user}`",
+      );
+
+      Auth::UserPass(rpc_user, rpc_pass)
+    } else if let (Some(rpc_user), Some(rpc_pass)) =
+      (env::var_os("RPC_USER"), env::var_os("RPC_PASS"))
```
casey, line 162: disaggregate `rpc_user` and `rpc_pass`

src/options.rs
```diff
@@ -561,6 +586,45 @@ mod tests {
         .unwrap(),
       Config {
         hidden: iter::once(id).collect(),
+        rpc_user: None,
+        rpc_pass: None,
+      }
+    );
+  }
+
+  #[test]
```
casey, line 595: unit test and precedent test

src/options.rs
```diff
@@ -561,6 +586,45 @@ mod tests {
...
+  fn test_rpc_user_and_pass_flags() {
+    let options = Arguments::try_parse_from([
+      "ord",
+      "--rpc-user",
+      "satoshi",
```
casey, line 600: use `foo` and `bar`

casey (changes requested):
Great start! Let's do the following:

- Write a function which takes a cli flag value, a environment variable, a config file option, and a default value, and returns the correct value. Unit test this so we don't have to test all combinations of configuration layers in integration tests.
- Write integration tests which test all three configuration layers independently: Config with CLI flags, environment variables, config file, and default value for RPC user and RPC pass. (We don't need to comprehensively test existing options.)
- Use `foo`, `bar`, and `baz` consistently when arbitrary strings or variable names are needed in tests. (Don't need to refactor other stuff, just this PR.

src/index/updater.rs
```diff
@@ -34,6 +34,7 @@ pub(crate) struct Updater {
   height: u64,
   index_sats: bool,
   sat_ranges_since_flush: u64,
+  options: Options,
```
andrewtoth, line 37: Do we really need to make `options`  a member of the `Updater` struct? It's only used one time on line 101, `let rx = Self::fetch_blocks_from(index, self.height, self.index_sats, self.options.clone())?;` where we are already passing in `index` which has the `options` as a member already.

raphjaph: good point, I'll try to remove it.

src/index/updater.rs
```diff
@@ -179,13 +179,13 @@ impl Updater {
...
     index_sats: bool,
+    options: Options,
   ) -> Result<mpsc::Receiver<BlockData>> {
     let (tx, rx) = mpsc::sync_channel(32);
 
     let height_limit = index.height_limit;
 
-    let client =
-      Client::new(&index.rpc_url, index.auth.clone()).context("failed to connect to RPC URL")?;
+    let client = options.bitcoin_rpc_client()?;
```
andrewtoth, line 188: Why not just `let client = index.options.bitcoin_rpc_client()?;`, and then we don't have to pass the cloned `options` in here at all?

src/options.rs
```diff
@@ -136,25 +140,56 @@ impl Options {
...
+    log::info!("Connecting to Bitcoin Core at {}", self.rpc_url());
 
-    log::info!(
-      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
-      cookie_file.display()
+    let rpc_user = Options::derive_var(
+      self.rpc_user.clone().map(|string| string.into()),
+      env::var_os("RPC_USER"),
+      self
+        .load_config()
```
andrewtoth, line 159: Could load config and store in a variable above this statement and reuse below for `rpc_pass`?

src/options.rs
```diff
@@ -136,25 +140,56 @@ impl Options {
...
 
-    let rpc_url = self.rpc_url();
+  pub(crate) fn auth(&self) -> Auth {
+    log::info!("Connecting to Bitcoin Core at {}", self.rpc_url());
 
-    log::info!(
-      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
-      cookie_file.display()
+    let rpc_user = Options::derive_var(
+      self.rpc_user.clone().map(|string| string.into()),
```
andrewtoth, line 156:
```suggestion
      self.rpc_user.as_ref().map(|string| string.into()),
```

src/options.rs
```diff
@@ -136,25 +140,56 @@ impl Options {
...
 
-    let client =
-      Client::new(&rpc_url, Auth::CookieFile(cookie_file.clone())).with_context(|| {
-        format!(
-          "failed to connect to Bitcoin Core RPC at {rpc_url} using cookie file {}",
-          cookie_file.display()
-        )
-      })?;
+    let rpc_pass = Options::derive_var(
+      self.rpc_pass.clone().map(|string| string.into()),
```
andrewtoth, line 167:
```suggestion
      self.rpc_pass.as_ref().map(|string| string.into()),
```

src/index/updater.rs
raphjaph: jup

raphjaph: jup

raphjaph:
Thanks for the review @andrewtoth! 

I've updated the things.

src/subcommand/server.rs
```diff
@@ -2266,7 +2288,8 @@ mod tests {
 
   #[test]
   fn feed() {
-    let server = TestServer::new_with_sat_index();
+    // let server = TestServer::new_with_sat_index();
```
andrewtoth, line 2291: Commented line should be removed

src/subcommand/server.rs
```diff
@@ -937,6 +937,28 @@ mod tests {
       Self::new_server(test_bitcoincore_rpc::spawn(), None, ord_args, server_args)
     }
 
+    fn new_with_regtest() -> Self {
```
andrewtoth, line 940:
```suggestion
    fn new_regtest() -> Self {
```
seems simpler?

andrewtoth (changes requested):

andrewtoth:
Two small nits but otherwise LGTM.

src/config.rs
```diff
@@ -3,6 +3,8 @@ use super::*;
 #[derive(Deserialize, Default, PartialEq, Debug)]
 pub(crate) struct Config {
   pub(crate) hidden: HashSet<InscriptionId>,
+  pub(crate) rpc_user: Option<String>,
+  pub(crate) rpc_pass: Option<String>,
```
casey, line 7: Struct fields should be sorted.

src/options.rs
```diff
@@ -37,6 +37,10 @@ pub(crate) struct Options {
   pub(crate) index_sats: bool,
   #[clap(long, short, help = "Use regtest. Equivalent to `--chain regtest`.")]
   pub(crate) regtest: bool,
+  #[clap(long, help = "Authenticate to Bitcoin Core RPC as <RPC_USER>.")]
+  pub(crate) rpc_user: Option<String>,
+  #[clap(long, help = "Authenticate to Bitcoin Core RPC with <RPC_PASS>.")]
+  pub(crate) rpc_pass: Option<String>,
```
casey, line 43: I notice that both `rpc_user` and `rpc_pass` are optional. What is the behavior when we pass `rpc_user` without `rpc_pass`, and vice versa? We should have test coverage for all four cases. (up/Up/uP/UP)

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
-    let cookie_file = self
-      .cookie_file()
-      .map_err(|err| anyhow!("failed to get cookie file path: {err}"))?;
+  fn derive_var(
+    arg: Option<OsString>,
+    env: Option<OsString>,
+    config: Option<OsString>,
+    default: Option<OsString>,
+  ) -> Option<OsString> {
+    arg.or(env).or(config).or(default)
```
casey, line 149: Genau!

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
+  }
 
-    let rpc_url = self.rpc_url();
+  pub(crate) fn auth(&self) -> Auth {
+    log::info!("Connecting to Bitcoin Core at {}", self.rpc_url());
+
+    let config = self.load_config().unwrap();
```
casey, line 155: This function must return an error, and we should include here a good error message. There are many reasons why config loading might fail, so this will be a common error that users encounter.

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
 
-  pub(crate) fn bitcoin_rpc_client(&self) -> Result<Client> {
-    let cookie_file = self
-      .cookie_file()
-      .map_err(|err| anyhow!("failed to get cookie file path: {err}"))?;
+  fn derive_var(
+    arg: Option<OsString>,
+    env: Option<OsString>,
+    config: Option<OsString>,
+    default: Option<OsString>,
```
casey, line 147: The current function signature doesn't allow passing, for example, strings, osstr, cstr, etc. We should make the function signature more generic, so the caller can pass any combination of types that can be converted into an osstr, which will allow fewer conversions at the call site.

I was thinking of suggesting this:

```suggestion
    arg: Option<impl AsRef<OsStr>>,
    env: Option<impl AsRef<OsStr>>,
    config: Option<impl AsRef<OsStr>>,
    default: Option<impl AsRef<OsStr>>,
```

But use of generics forces the compiler to emit a monomerized function for every combination of argument types.

I'm leaning towards this, which should be faster to compile, and which is a bit simpler to read:

```suggestion
    arg: Option<&OsStr>,
    env: Option<&OsStr>,
    config: Option<&OsStr>,
    default: Option<&OsStr>,
```

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
-    let rpc_url = self.rpc_url();
+  pub(crate) fn auth(&self) -> Auth {
+    log::info!("Connecting to Bitcoin Core at {}", self.rpc_url());
+
+    let config = self.load_config().unwrap();
+
+    let rpc_user = Options::derive_var(
+      self.rpc_user.as_ref().map(|string| string.into()),
+      env::var_os("RPC_USER"),
+      config.rpc_user.map(|string| string.into()),
```
casey, line 160: This will need some fiddling to work with the changed function signature.

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
+      None,
+    );
 
-    log::info!(
-      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
-      cookie_file.display()
+    let rpc_pass = Options::derive_var(
+      self.rpc_pass.as_ref().map(|string| string.into()),
```
casey, line 165: This will need some fiddling to work with the changed function signature.

casey (changes requested):
Did a partial review, check out my comments!

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
     )
   }
 
-  pub(crate) fn bitcoin_rpc_client(&self) -> Result<Client> {
-    let cookie_file = self
-      .cookie_file()
-      .map_err(|err| anyhow!("failed to get cookie file path: {err}"))?;
+  fn derive_var(
```
casey, line 143: Also, drive-by post-partial-review comment: I think this function should take `Option<&str>` and return `Option<&str>`.

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
-    let client =
-      Client::new(&rpc_url, Auth::CookieFile(cookie_file.clone())).with_context(|| {
-        format!(
-          "failed to connect to Bitcoin Core RPC at {rpc_url} using cookie file {}",
-          cookie_file.display()
-        )
-      })?;
+    match (rpc_user, rpc_pass) {
+      (Some(rpc_user), Some(rpc_pass)) => Auth::UserPass(
+        rpc_user.into_string().expect("rpc_user is invalid UTF-8"),
```
casey, line 173: These should be converted to `&str` at the call site. Additionally, we need to return an error if the rpc_user and rpc_pass aren't UTF-8

src/options.rs
```diff
@@ -136,25 +140,50 @@ impl Options {
...
+        rpc_pass.into_string().expect("rpc_pass is invalid UTF-8"),
+      ),
+      _ => Auth::CookieFile(self.cookie_file().unwrap()),
```
casey, line 176: Unwraps should be avoided. Also, the cookie file not being present is a common error, so we need to return a good error message.

src/options.rs
```diff
@@ -561,6 +590,39 @@ mod tests {
         .unwrap(),
       Config {
         hidden: iter::once(id).collect(),
+        rpc_user: None,
+        rpc_pass: None,
```
casey, line 594: These should be sorted.

casey (changes requested):
Some more changes.

src/options.rs
raphjaph: It tries to get it from env or config file. Finally it looks at the cookie file.

src/options.rs
```diff
@@ -136,25 +140,69 @@ impl Options {
...
+
+    let rpc_pass = Options::derive_var(
+      self.bitcoin_rpc_pass.as_deref(),
+      Some("BITCOIN_RPC_PASS"),
+      config.rpc_pass.as_deref(),
+      None,
+    )?;
+
+    match (rpc_user, rpc_pass) {
+      (Some(rpc_user), Some(rpc_pass)) => Ok(Auth::UserPass(rpc_user, rpc_pass)),
```
casey, line 185: We should handle the `None/Some` and `Some/None` cases by throwing an error, and we should have tests for both of these cases.

tests/index.rs
```diff
@@ -1,4 +1,4 @@
-use super::*;
+use {super::*, crate::command_builder::ToArgs};
```
casey, line 1: What is the import used for? Can we remove it?

tests/index.rs
```diff
@@ -35,3 +35,32 @@ fn re_opening_database_does_not_trigger_schema_check() {
     .rpc_server(&rpc_server)
     .run();
 }
+
+#[test]
+fn index_runs_with_rpc_user_and_pass_as_env_vars() {
```
casey, line 40: Let's modify this test to make sure we're using the correct values. Let's record the user and pass which are passed each time a client connects, and then do something like `assert_eq!(rpc_server.client_credentials, &[("foo", "bar")])`.

tests/server.rs
```diff
@@ -298,3 +298,50 @@ fn expected_sat_time_is_rounded() {
     r".*<dt>timestamp</dt><dd><time>.* \d+:\d+:\d+ UTC</time> \(expected\)</dd>.*",
   );
 }
+
+#[test]
+fn server_runs_with_rpc_user_and_pass_as_env_vars() {
```
casey, line 303: Let's do the same style of test in index_runs_with_rpc_user_and_pass_as_env_vars

casey (changes requested):
SO CLOSE

tests/index.rs
raphjaph: Yes, we need it for the `to_args()` function call in one of the tests.

tests/index.rs
raphjaph: The way the our mock bitcoin core instance is written I do not have access to the Basic Authentication http header. So I can't know what credentials are passed in.

tests/server.rs
raphjaph: see above

casey (approved):

PR #1535, opened 2023-02-06
switch recoverable block fetching error message to warning

windsok:
I think we actually forgot to switch this recoverable error message to `warn` in https://github.com/casey/ord/pull/1466 (I missed it in my review)


casey:
Just seeing this now. Did we get this?

windsok:
On reviewing the code further, I believe I was mistaken and this message is actually the final message after `ord` gives up trying, so makes sense for it to be an error.

PR #1536, opened 2023-02-07
add option to set inscription destination

rot13maxi:
Split just this feature out of #1512


src/subcommand/wallet/inscribe.rs
```diff
@@ -46,6 +46,8 @@ pub(crate) struct Inscribe {
   pub(crate) no_backup: bool,
   #[clap(long, help = "Don't sign or broadcast transactions.")]
   pub(crate) dry_run: bool,
+  #[clap(long, help = "Send reveal output to <INSCRIPTION_DESTINATION>")]
+  pub(crate) inscription_destination: Option<Address>,
```
casey, line 50:
```suggestion
  #[clap(long, help = "Send inscription to <DESTINATION>.")]
  pub(crate) destination: Option<Address>,
```

src/subcommand/wallet/inscribe.rs
```diff
@@ -63,7 +65,9 @@ impl Inscribe {
 
     let commit_tx_change = [get_change_address(&client)?, get_change_address(&client)?];
 
-    let reveal_tx_destination = get_change_address(&client)?;
+    let reveal_tx_destination = self
+      .inscription_destination
+      .unwrap_or(get_change_address(&client)?);
```
casey, line 70: Since `get_change_address` will do a little work, and allocate a change address, you can use `unwrap_or_else`, which takes a closure that is only ever called if the `Option` is `None`:

```suggestion
      .unwrap_or_else(|| get_change_address(&client)?);
```

In general, this should be preferred when calling a function that might be a little slower, or which might have side effects.

This `_else` naming pattern is common, lots of combinators like `Result` have it too. Sometimes it's also called `foo` and `with_foo`, where `foo` takes a value, and `with_foo` takes a closure that returns a value.

src/subcommand/wallet/inscribe.rs
```diff
@@ -635,4 +639,30 @@ mod tests {
       error
     );
   }
+
+  #[test]
+  fn inscribe_with_specific_destination() {
```
casey, line 644: I think this test might actually be testing something that's already being tested, e.g. the specific destination address. For this, because it introduces a flag, the only test I think we need is an integration test which calls inscribe with `--destination`, and makes sure that the mined reveal transaction has the right destination.

casey (changes requested):
Nice, a few random comments. Love the test, but I think we need an integration test that does this e2e w/the flag.

src/subcommand/wallet/inscribe.rs
rot13maxi: I suck at naming things. thanks for the suggestion!

src/subcommand/wallet/inscribe.rs
rot13maxi: oh right, `unwrap_or` will eagerly evaluate the default. great catch!

src/subcommand/wallet/inscribe.rs
rot13maxi: ack

src/subcommand/wallet/inscribe.rs
rot13maxi: So it turns out you can't put the `?` in a closure for `unwrap_or_else`, which makes sense. So I put an `expect` on that instance of `get_change_address`. if you'd prefer expect-free code, I'm happy to change it back

rot13maxi:
@casey feedback addressed when you have time to take another look :)

src/subcommand/wallet/inscribe.rs
```diff
@@ -52,6 +52,8 @@ pub(crate) struct Inscribe {
   pub(crate) no_limit: bool,
   #[clap(long, help = "Don't sign or broadcast transactions.")]
   pub(crate) dry_run: bool,
+  #[clap(long, help = "Send inscription to <DESTINATION>")]
```
raphjaph, line 55:
```suggestion
  #[clap(long, help = "Send inscription to <DESTINATION>.")]
```

tests/wallet/inscribe.rs
```diff
@@ -355,6 +355,32 @@ fn inscribe_with_dry_run_flag_fees_inscrease() {
...
+    .output::<ord::subcommand::wallet::receive::Output>()
+    .address;
+
+  let txid = CommandBuilder::new(format!(
+    "wallet inscribe --destination {destination} degenerate.png"
+  ))
+  .write("degenerate.png", [1; 520])
+  .rpc_server(&rpc_server)
+  .output::<Inscribe>()
+  .reveal;
```
raphjaph, line 375:
```suggestion
  .reveal;
  
```

raphjaph (approved):

Issue #1430, opened 2023-01-31
GLIBC_2.32 not found


jacobburrell:
I've got the following issue twice on two separate instances:

`ord: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ord)`

Has anyone had the same issue? 

Ubuntu 20.04.1 LTS

dude513:
Consider upgrading to Ubuntu 22.04 to have a higher version of GLIBC.

```
sudo apt update -y && sudo apt upgrade -y
sudo do-release-upgrade -d
```

To check GLIBC version.

```
ldd --version | head -n1
```

jacobburrell:
Thanks!

That worked. Probably should mention in dependency or requirement somewhere, or improve the error message.

casey:
Good idea! Can someone open a PR to add this to the readme?

AlexLloyd0:
I have the same issue on Debian 11, and an OS upgrade to is not feasible for this machine. Building from source worked though

pavillag:
Is there a way to update `:~$ ldd --version | head -n1
ldd (Ubuntu GLIBC 2.27-3ubuntu1.6) 2.27` to any of the supported ord versions? 
![image](https://user-images.githubusercontent.com/30845498/217844800-c7abe682-fbe2-48fb-9567-124b816a85b7.png)

missingpurpose:
> Is there a way to update `:~$ ldd --version | head -n1 ldd (Ubuntu GLIBC 2.27-3ubuntu1.6) 2.27` to any of the supported ord versions? ![image](https://user-images.githubusercontent.com/30845498/217844800-c7abe682-fbe2-48fb-9567-124b816a85b7.png)

Having the same issue, are you using the AWS Full node as well?

0xBeans:
same issues as @pavillag and @missingpurpose , running on EC2

dude513:
`ord` requires GLIBC version 2.32 or higher.
To compile from source, first we need to install gawk and bison that are used in compiling Glibc:

```
sudo apt-get install gawk bison -y
```

Download and unpack the source, then configure a separate build directory and install glibc-2.34:

```
wget -c https://ftp.gnu.org/gnu/glibc/glibc-2.34.tar.gz
tar -zxvf glibc-2.34.tar.gz && cd glibc-2.34
mkdir glibc-build && cd glibc-build
../configure --prefix=$HOME/glibc-2.34
```
[...]

PR #1556, opened 2023-02-07
Update README.md

dude513:
Adding dependencies notes in README.md.
Reference: https://github.com/casey/ord/issues/1430.


casey (changes requested):
Thanks for the PR!

I'd rather not suggest upgrading ubuntu, since that's a big step that might break a user's system. For a first step, I'd like to just document the GLIBC version required, and if it's possible to install it, provide instructions for just installing the right version.

Also, we have a "building" section, so it's probably better to include this info there.

casey:
Feel free to reuse this PR and just modify it!

dude513:
> Feel free to reuse this PR and just modify it!

On my way.

dude513:
I'm not sure if we need an instruction on how to install Glibc from source, please tell me if we should remove it: https://github.com/casey/ord/pull/1556/commits/6d6d3afa3091cc8cb5eee3ea1d8df86c20d9c57a.

casey (changes requested):
I think we should probably remove the glibc-from-source instructions. If someone doesn't have the right version, the best thing is probably to figure out how to upgrade their OS, or use their OS's provided way of installing glibc.

dude513:
> I think we should probably remove the glibc-from-source instructions. If someone doesn't have the right version, the best thing is probably to figure out how to upgrade their OS, or use their OS's provided way of installing glibc.

You have mentioned that upgrading is not safe to recommend to users, likewise there is no "OS's provided way of installing glibc", but only compiling from source. For example on Ubuntu 20 `apt-cache policy libc6` the candidate version is `2.31-0ubuntu9.9`.

I left a comment here https://github.com/casey/ord/issues/1430#issuecomment-1425434981, this is something I would rather use, or the only option left is to say what glibc version is required and let users to figure out how to upgrade it.

raphjaph:
This is probably fixed by upgrading Ubuntu now.

PR #1562, opened 2023-02-07
feat: add ability to connect to bitcoind using inline credentials

lgalabru:
Hey @casey!
I'm trying to use `ord` locally, within a containerized setup. 
The cookie method is unfortunately not an option, was there a particular reason for not adding the user/pass strategy?
This PR was tested locally and is fixing my issue.
Thanks!

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -148,24 +148,24 @@ impl Options {
   }
 
   pub(crate) fn bitcoin_rpc_client(&self) -> Result<Client> {
-    let cookie_file = self
-      .cookie_file()
-      .map_err(|err| anyhow!("failed to get cookie file path: {err}"))?;
-
     let rpc_url = self.rpc_url();
 
-    log::info!(
-      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
-      cookie_file.display()
-    );
+    let auth = if let Some((user, pass)) = self.rpc_userpass() {
+      log::info!(
+        "Connecting to Bitcoin Core RPC server at {rpc_url} using rpc_username `{user}`",
+      );
+      Auth::UserPass(user, pass)
+    } else {
+      let cookie_file = self.cookie_file()
+        .map_err(|err| anyhow!("failed to get cookie file path: {err}"))?;
+      log::info!(
+        "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
+        cookie_file.display()
+      );
+      Auth::CookieFile(cookie_file)
+    };
 
-    let client =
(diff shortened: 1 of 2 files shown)


lgalabru:
When running `ord` as a server, I think a common good practice devops is to pass environment variables, so instead of running:
```
$ ord -r --rpc-username admin --rpc-password password info
```
You'd pass 
```
ord -r --rpc-username $BTC_RPC_USERNAME --rpc-password $BTC_RPC_PASSWORD info
```
Which works in this PR.
I think config files are great and complementary to command flags, but not designed either for solving secrets management issues.

casey:
If environment variables are preferred, than I'd be totally fine with a PR that caused `ord` check for those ORD_BITCOIND_RPC_USER and ORD_BITCOIND_RPC_PASS and use them directly, instead of needing to pass them. Would that be better?

lgalabru:
Environment variables seem adequate for power users features (say `ORD_DEBUG`, etc) - poor discovery won't block new users to run `ord`.
If the `cookie-file` auth strategy is available through command flags, then other auth strategies should also be available via this interface.
From an ecosystem point of view, users running `ord` will also be running `bitcoind` so designing an experience consistent with `bitcoind` where env var could be reused could make a lot of sense:
```
$ bitcoind -help
  ...
RPC server options:
  ...
  
  -rpccookiefile=<loc>
       Location of the auth cookie. Relative paths will be prefixed by a
       net-specific datadir location. (default: data dir)

  -rpcpassword=<pw>
       Password for JSON-RPC connections

  -rpcuser=<user>
       Username for JSON-RPC connections
```

casey:
Agree on all points. I just want to prioritize specific use cases that specific users have. We eventually want a consistent, multi-layer interface for all config options, but for now, since username and pass are best left out of command line args for the reasons above, I want to make sure we implement better alternatives first.

KeepYourCoins:
Instead of taking these in from command line args, can they be parsed from the bitcoin.conf file?

src/options.rs
```diff
@@ -80,6 +84,20 @@ impl Options {
...
 
+  pub(crate) fn rpc_userpass(&self) -> Option<(String, String)> {
+    let user = match &self.rpc_username {
+      Some(username) => username.clone(),
+      None => return None
+    };
+
+    let pass = match &self.rpc_password {
+      Some(password) => password.clone(),
+      None => "".to_string()
```
KeepYourCoins, line 95: Why is this returning an empty string over a `None`?

zone117x:
Are there any recommendations for moving this PR along?

raphjaph:
We added this check out PR https://github.com/casey/ord/pull/1527

PR #1578, opened 2023-02-08
Drain value_cache instead of reallocating

andrewtoth:


raphjaph (approved):
Nice!

casey:
This seems like an obvious win to me, and I think we should merge regardless. That being said, did you benchmark this? Is there a speedup? Optimization without data is often super tricky, so I always want to make sure we're optimizing the right stuff.

andrewtoth:
I did not benchmark, no. We can hold off merging if you want until I do. It just seemed obvious when reviewing the indexing code so I thought it would be good to PR on its own.

casey (changes requested):
A benchmark would be great! Since we're don't think it's critical, there's no rush, and good to take a principled stance with benchmarking and optimization. I've spent a lot of time optimizing things that don't matter, which would have been saved by benchmarking first. @veryordinally probably has ideas here

andrewtoth:
Since this only happens every 5000 blocks, it doesn't have a meaningful impact on benchmarks. Closing.

Issue #1558, opened 2023-02-07
Using REST is also a more suitable interface for this application than RPC. It requires no authentication and can transmit blocks in binary form. It also does not have to parse a json request. It does require `rest=1` in config though. However, it still deserializes a block to `CBlock` internally before sending. Ideally https://github.com/bitcoin/bitcoin/pull/26415 would be merged and this would be by far the fastest way to fetch blocks.


casey:
Using REST is also a more suitable interface for this application than RPC. It requires no authentication and can transmit blocks in binary form. It also does not have to parse a json request. It does require `rest=1` in config though. However, it still deserializes a block to `CBlock` internally before sending. Ideally https://github.com/bitcoin/bitcoin/pull/26415 would be merged and this would be by far the fastest way to fetch blocks.
See https://github.com/andrewtoth/rust-bitcoincore-rest.

_Originally posted by @andrewtoth in https://github.com/casey/ord/issues/1502#issuecomment-1416828485_

casey:
Just going through my mentions and saw this again. I love the idea of using REST where we can. I find JSON-RPC very awkward, and don't really like the way it uses HTTP, but doesn't use HTTP verbs, headers, query parameters, status codes, and instead recreates those things in JSON.

I would love to see an initial simple use of the REST API merged, default-off, so we can start exploring it. We use HTTP a lot and I expect to expose an HTTP REST API, so it makes sense to use it internally.

andrewtoth:
I don't think we need to make it an option. If the rest endpoint is responding we use it, otherwise continue with RPC.

andrewtoth:
Most simple http libraries are async, which doesn't really work with the current architecture. I found https://github.com/oxigraph/oxhttp which seems to be what we need. Not sure if you have a better library in mind?

If you'd like to see the benefits, using ApacheBench and requesting via REST vs RPC an 85kb transaction `f8af3c34c574e3039c47de0a32bce6b2753e4511337d0c9e13d31d7f2c83e374`:
Make sure you have `rest=1, rpcuser=user, rpcpassword=password` for these
`ab -n 10000 -c 1 "http://127.0.0.1:8332/rest/tx/f8af3c34c574e3039c47de0a32bce6b2753e4511337d0c9e13d31d7f2c83e374.bin"`
for REST
```
[...]

PR #1636, opened 2023-02-10
Use REST interface for calling get_raw_transaction

andrewtoth:
Closes https://github.com/casey/ord/issues/1558.

Use the REST endpoint for fetching raw transactions instead of RPC endpoint. Only really useful when implementing https://github.com/casey/ord/issues/1364 by reverting https://github.com/casey/ord/pull/1357.

When combined with https://github.com/casey/ord/pull/1516, I synced to 775812 (without `--index-sats`) in 1 hour 6 minutes. And no more pesky JSON-RPC errors. 

The progress bar is kind of silly now though. It should really show progress from first inscription block to current height if no `--index-sats`.

This needs to have bitcoind run with `-rest` though, otherwise painful slowness and rpc errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2179,6 +2179,7 @@ dependencies = [
  "hex",
  "html-escaper",
  "http",
+ "hyper",
  "indicatif",
  "lazy_static",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ futures = "0.3.21"
 hex = "0.4.3"
 html-escaper = "0.2.0"
 http = "0.2.6"
+hyper = { version = "0.14.24", features = ["http1", "client"] }
 indicatif = "0.17.1"
 lazy_static = "1.4.0"
 log = "0.4.14"
--- a/src/index.rs
+++ b/src/index.rs
@@ -19,6 +19,7 @@ use {
 };
 
 mod entry;
+mod rest;
 mod rtx;
 mod updater;
 
--- a/src/index/rest.rs
+++ b/src/index/rest.rs
@@ -1,0 +1,41 @@
+use {
+  anyhow::Result,
+  bitcoin::{consensus::deserialize, BlockHash, Transaction, Txid},
+  hyper::{client::HttpConnector, Client, Uri},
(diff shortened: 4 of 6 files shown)


andrewtoth:
The library `oxhttp` creates a new `TcpStream` connection for every request, which unfortunately breaks on Macs (https://github.com/apoelstra/rust-jsonrpc/issues/67). I tested a rewritten version copying `SimpleHttpTransport` from `rust-jsonrpc`, but it unfortunately has degraded performance. From https://github.com/apoelstra/rust-jsonrpc/pull/76#issuecomment-1320956515, it looks like it's much faster to create a new socket connection to bitcoind for every request. Benchmarking a sync with the same setup that I got 1:06 with using a single socket connection, I get ~5hrs :(.

Benchmarking RPC vs REST, it is about ~10-20% faster using REST. However, being able to use a new connection per request for either method is definitely what we want to do. I think the performance is so much better we might want to support reusing connections if not using Mac? I know a lot of users are running this in cloud instances running on linux, so might be worth it?

andrewtoth:
Ok update on this, I switched out `oxhttp` with `hyper`. This does two things - hyper pools connections and reuses them for requests if they are going to the same host. So in theory this should work on Mac. I still need someone on Mac to confirm for me.

Second, it is `async`, so while we don't batch requests like JSON-RPC can, we can execute requests concurrently. This patch now executes up to 15 requests concurrently, since the default `-rpcworkqueue` is set to 16 and if we exceed that it breaks. With this configuration I managed to sync to block 77260 in 80 minutes.

But there's more. If you configure bitcoind to have a higher `rpcworkqueue`, you can sync even faster! I set it to 120 and managed to sync in just 39 minutes.

andrewtoth:
@tzarebczan awesome! Did you also have #1357 reverted? Otherwise it doesn't use this feature.

What about #1516, did you try with that? If not, that would make sense why you have the JSON-RPC error, because it would still be fetching blocks via JSON-RPC.

tzarebczan:
I used speedup improvements branch from your repo

andrewtoth:
Closing in favor of https://github.com/casey/ord/pull/1759

PR #1637, opened 2023-02-10
Handle block count RPC error gracefully

andrewtoth:
If we fail to fetch the block count, it will just return an error from `update_index` with a JSON-RPC error. However, any uncommitted blocks will not be committed because it doesn't exit the loop. They will have to be indexed again. 

This PR will just log the error but continue indexing, so we can keep our cache. It will show a progress greater than 100%, but better than having a JSON-RPC error and uncommitted blocks.

The json-rpc client times out and fails if it's not used for a long time, like during a long index update. Since a new client is created to fetch blocks in the background, the `index.client` is not used until near the end, where it already timed out.


andrewtoth:
In my testing this call seems to always fail. Not sure why since I'm using only REST and p2p in the entire `updater.rs` except for this call and the one on line 81 which is the initial one. But probably something to investigate further. I think this should be merged regardless.

andrewtoth:
The reason it fails is because the socket connection times out because it's not used during sync with REST.

donnlee:
I confirm that https://github.com/casey/ord/commit/50f2ad7aebcca87bd31d2b28fa2d8ab67d7cf7cc on ord 0.5.0 has fixed the time-wasting issue of blocks re-indexing at 775000 every time.
This is awesome. I was able to finish the last 1% of indexing today 🎉 
Now i have an ord index compatible with 0.5.0

vicariousdrama:
I second @donnlee . My node succeeded on this as well (4GB raspberry pi)

dougEfresh:
This PR works!

raphjaph (approved):
ACK by peanut gallery!

jesuspb2:
Hi guys, 

I have the version 0.5.0 and I still have the issue of indexing: HTTP response too short: length 0, needed 12

What can I do in this case? 
Thank you

andrewtoth:
@jesuspb2 upgrade to 0.5.1.

PR #1650, opened 2023-02-10
add Chinese intro and regtest  inscription guide

dajuguan:
Add detailed Chinese explaination on the idea of Ordinal number and Inscriptions. Besides, the bitcoin regtest net inscription guide is introduced, which can also be adapted to an English version.


dajuguan:
file is wrong

PR #1651, opened 2023-02-10
Add Chinese explaination and regtest net guides

dajuguan:
Add detailed Chinese explaination on the idea of Ordinal number and Inscriptions. Besides, the bitcoin regtest net inscription guide is introduced, which can also be adapted to an English version.


dajuguan:
@casey

raphjaph:
We have a Chinese translation now at https://docs.ordinals.com/zh/introduction.html

PR #1654, opened 2023-02-10
Cache all output values when we fetch a new tx

andrewtoth:
Everytime we have a cache miss when we are skipping txs, we fetch the raw transaction but then throw away any outputs we're not interested in. This can result in looking up the same tx every time one of its outputs is spent. Instead, cache all outputs so that we only ever look up a tx once.


andrewtoth:
Seems like this makes the sync slower due to longer commit times.

PR #1662, opened 2023-02-11
Sketch of API

casey:


dajuguan:
Awesome, waiting for the full APIs!

ynohtna92:
Went ahead and used this template for our litecoin fork, feel free to review it so we both can use the same implementation. Should be able to just apply the commit as a patch/cherrypick with it still being compatible.

https://github.com/ynohtna92/ord-litecoin/commit/a9b7a3f2e1e449228a9fa348b699d47e0ab72472

We need to add an extra field to the database so that it is able to index by address as well, as this seems quite useful for wallets containing multiple utxos.

ie. `https://ordinals.com/address/:address`

ynohtna92:
Moved here https://github.com/casey/ord/pull/2036

raphjaph:
We're using this now, so will close.

PR #1680, opened 2023-02-11
Add teleburn command

casey:
People are interested in burning high-value Ethereum NFTs, so this feature needs to be extremely safe:

- [ ] add tests
- [ ] review cryptographic assumptions

---

- [ ] generate ethereum address with checksum (no ethereum dependencies for this, must be implemented manually)
- [ ] manual end-to-end test (we can just make sure that you can derive genesis inscription ID from the ethereum address)
- [ ] make sure that the index is synced and that the inscription that you're burning is in your wallet
- [ ] add ability to reverse teleburn: get the inscription id from the teleburn address
- [ ] show teleburn address on inscription page
- [ ] only accept ethereum addresses with checksum when deserializing


Rob1Ham:
I'll do a couple tests to verify this works.

casey:
Is length extension an issue?

rot13maxi:
No, it shouldn't be

rot13maxi:
knocked out a few of your checklist items in #1825

rot13maxi:
The reverse lookup is an interesting feature. On the one hand, it would be super cool to let anyone search for an eth address and find the inscription that it was teleburned to. On the other hand, I'm pretty sure the only way to do that efficiently is to keep an index that maps teleburn address to inscription ID for every inscription. I started hacking that up here:
https://github.com/rot13maxi/ord/compare/master...rot13maxi:ord:teleburn

I'm new to redb, so there might be a better way to do it, but the tl;dr is that we add an entry to a simple str->InscriptionId table when we update the index. I haven't profiled it yet to see what the perf hit is.

rot13maxi:
hey @raphjaph and @casey  what else are you looking for in this?

raphjaph (approved):
let it burn!

PR #1697, opened 2023-02-12
Remove passing config file in ord service file

### LINKED ISSUE #1665: Passing ord.yaml config file in ord.service causes error on deploy without manual intervention
Needed to create the configuration file manually since `ord` doesn't create it if it doesn't exist

terror:
Fixes #1665


casey:
We need this config on ordinals.com, so we can't remove it.

Should we can make ord take a --config-dir option?

If this flag is passed, ord will look in that dir for ord.yaml, and if it doesn't find it, just not load any config. Then, we can put ord.yaml in a dir that already exist, like /var/lib/ord, and for users without it ord will not complain.

terror:
> We need this config on ordinals.com, so we can't remove it.
> 
> Should we can make ord take a --config-dir option?
> 
> If this flag is passed, ord will look in that dir for ord.yaml, and if it doesn't find it, just not load any config. Then, we can put ord.yaml in a dir that already exist, like /var/lib/ord, and for users without it ord will not complain.

Yah that sounds good to me.

raphjaph (approved):

PR #1669, opened 2023-02-11
Enable Passphrase for Restore Command

Psifour:
## Description
Implements an option to pass a passphrase when restoring from mnemonic.

## Related Issue
Closes #1671 

## Motivation and Context
Without this change users are only able to (easily) import a wallet from Sparrow if they did NOT use a passphrase.

## How Has This Been Tested?
Default case has been verified to call to_seed() with no passphrase and all single word data passed with the option is verified to replace that. Utilizes the standardized behavior as seen in other sub-commands.

PR #1698, opened 2023-02-12
Expansion of CONTRIBUTING

Psifour:
Still requires documentation for Translation/Localization efforts and a PR template.


raphjaph:
I'll convert to Draft since it's not ready for review yet

Psifour:
Absolutely should be a draft, I have no knowledge of any translation/localization being worked on currently and would love an assist if you know who is doing so (assuming it is being done).

I have a PR template (based on GitHub template) that I typically use, can be seen on #1669. If this is acceptable as a starting point I will include it in documentation for now.

Psifour:
Localization and PR templates were put in backlog. Documentation ready, but a proof-read or grammatical review would be appreciated.

casey (changes requested):
Sorry for letting this sit! This is great stuff, but the `CONTRIBUTING` file really functions as a legal disclaimer which informs contributors of the copyright status of their contributions, should they make them. I'd like to keep it short. Also, check out the [contributing docs page](https://docs.ordinals.com/contributing.html), it might overlap with this. I think the right move here is to merge these changes into the contributing docs page.

PR #1713, opened 2023-02-13
Add brotli compression for text-based inscriptions

terror:
Tests:
- [x] content encoding tag is serialized corretly. (just create inscription with content encoding field and turn it into an envelope)
- [x] test that content encoding appears on /inscription page
- [x] add unit tests for acceptencodingbrotli (test that it's correct with multiple values, spaces, qvalues, etc)
- [x] allow content encodings other than brotli (may need to get header value, and check if inscription.content_encoding() is present, instead of using AcceptEncodingBrotli)
- [x] inscriptions are compressed when passing the `--compress` flag
- [x] inscriptions are not compressed if compression doesn't make content smaller
- [x] test that when getting /content, server returns 406 if no content encoding header is passed when inscription is compressed
- [x] test that when getting /preview, server returns 406 if no content encoding header is passed when inscription is compressed


casey:
You can add a bool to Media::TABLE, true for text formats and false for non text formats, which determines what should be compressed. Only text should be compressed for now.

Psifour:
I was working on this concurrently. Will merge my work here instead of a new PR.

Psifour:
>  Only text should be compressed for now.

I had tested the majority of our existing MIME types to verify which would benefit from brotli. The majority show gains with the 3d types being exceptionally well suited for compression.
image

src/inscription.rs
```diff
@@ -8,24 +8,35 @@ use {
...
-  std::{iter::Peekable, str},
+  brotli::enc::writer::CompressorWriter,
+  std::{io::Write, iter::Peekable, str},
 };
 
 const PROTOCOL_ID: &[u8] = b"ord";
 
 const BODY_TAG: &[u8] = &[];
 const CONTENT_TYPE_TAG: &[u8] = &[1];
+const CONTENT_ENCODING_TAG: &[u8] = &[2];
```
terror, line 19: Make this an odd number

src/inscription.rs
```diff
@@ -35,20 +46,35 @@ impl Inscription {
   pub(crate) fn from_file(chain: Chain, path: impl AsRef<Path>) -> Result<Self, Error> {
     let path = path.as_ref();
 
+    let (content_type, compress) = Media::content_type_and_compress_for_path(path)?;
```
terror, line 49: Just compress everything

src/envelope.rs
```diff
@@ -80,8 +81,9 @@ impl From<RawEnvelope> for ParsedEnvelope {
     let content_type = remove_field(&mut fields, &CONTENT_TYPE_TAG);
```
casey, line 81: Sort alphabetically.

src/inscription.rs
```diff
@@ -38,9 +40,14 @@ pub struct Inscription {
 
 impl Inscription {
   #[cfg(test)]
-  pub(crate) fn new(content_type: Option<Vec<u8>>, body: Option<Vec<u8>>) -> Self {
```
casey, line 41: Let's leave this functions is, and create inscriptions which need content encoding with the `Inscription {…}` constructor.

src/media.rs
```diff
@@ -19,44 +20,45 @@ pub(crate) enum Media {
...
+    ("text/markdown;charset=utf-8", CompressionMode::Text, Media::Markdown, &["md"]),
+    ("text/plain", CompressionMode::Text, Media::Text, &[]),
+    ("text/plain;charset=utf-8", CompressionMode::Text, Media::Text, &["txt"]),
+    ("text/x-python", CompressionMode::Text, Media::Code, &["py"]),
+    ("video/mp4", CompressionMode::Generic, Media::Video, &["mp4"]),
+    ("video/webm", CompressionMode::Generic, Media::Video, &["webm"]),
   ];
 
-  pub(crate) fn content_type_for_path(path: &Path) -> Result<&'static str, Error> {
+  pub(crate) fn content_type(path: &Path) -> Result<(&'static str, CompressionMode), Error> {
```
casey, line 61:
```suggestion
  pub(crate) fn content_type_for_path(path: &Path) -> Result<(&'static str, CompressionMode), Error> {
```

src/inscription.rs
```diff
@@ -189,6 +230,10 @@ impl Inscription {
     str::from_utf8(self.content_type.as_ref()?).ok()
   }
 
+  pub(crate) fn content_encoding(&self) -> Option<&str> {
```
casey, line 233: Return `Option<HeaderValue>`, and content encoding values which are invalid header values will be ignored.

src/subcommand/server/accept_encoding.rs
```diff
@@ -0,0 +1,123 @@
+use {super::*, axum::extract::FromRef};
+
+pub(crate) struct AcceptEncoding(pub(crate) Vec<String>);
+
+#[async_trait::async_trait]
+impl<S> axum::extract::FromRequestParts<S> for AcceptEncoding
```
casey, line 6: Let's change this to return the header value as a `String`, to avoid allocation a `Vec<String>`, and then we can do the check in the content response function. Content repsonse should take a `AcceptEncoding` so we don't have a random `String` arg

src/subcommand/server/accept_encoding.rs
```diff
@@ -0,0 +1,123 @@
...
+                })
+                .collect::<Vec<String>>()
+            })
+            .unwrap_or_default()
+        })
+        .unwrap_or_default(),
+    ))
+  }
+}
+
```
casey, line 42:
```
impl AcceptEncoding {
  fn acceptable(encoding: &str) -> bool {
          self.0.split
              .split(',')
                .any(|value| {
                  value
                    .split(';')
                    .next()
                    .unwrap_or_default()
                    .trim()
                    == encoding
                })
  }
}
``

File: src/templates/inscription.rs [hunk 9]
```diff
@@ -467,4 +467,57 @@ mod tests {
...
+          <dd><a class=monospace href=/tx/1{64}>1{64}</a></dd>
+          <dt>location</dt>
+          <dd class=monospace>1{64}:1:0</dd>
+          <dt>output</dt>
+          <dd><a class=monospace href=/output/1{64}:1>1{64}:1</a></dd>
+          <dt>offset</dt>
+          <dd>0</dd>
+          <dt>ethereum teleburn address</dt>
+          <dd>0xa1DfBd1C519B9323FD7Fd8e498Ac16c2E502F059</dd>
+        </dl>
```
casey, line 518:
```suggestion
.*
          <dt>content encoding</dt>
          <dd>br</dd>
.*
```

src/test.rs
```diff
@@ -125,7 +125,19 @@ impl From<InscriptionTemplate> for Inscription {
 }
 
 pub(crate) fn inscription(content_type: &str, body: impl AsRef<[u8]>) -> Inscription {
-  Inscription::new(Some(content_type.into()), Some(body.as_ref().into()))
+  Inscription::new(Some(content_type.into()), None, Some(body.as_ref().into()))
+}
+
+pub(crate) fn inscription_with_encoding(
```
casey, line 131: Let's just use the `Inscription{…}` constructor directly.

static/preview-text.js
```diff
@@ -1,3 +1,12 @@
+const inscription = document.documentElement.dataset.inscription;
+
+const response = await fetch(`/content/${inscription}`);
+const text = await response.text();
+
```
casey, line 5:
```suggestion
```

tests/wallet/inscribe.rs
```diff
@@ -1,3 +1,5 @@
+use std::ops::Deref;
+
 use super::*;
```
casey, line 4:
```suggestion
use {super::*, std::ops::Deref};

```

casey (changes requested):

src/subcommand/server.rs
```diff
@@ -995,26 +1001,44 @@ impl Server {
         .and_then(|content_type| content_type.parse().ok())
         .unwrap_or(HeaderValue::from_static("application/octet-stream")),
     );
+
+    if let Some(content_encoding) = inscription.content_encoding() {
+      if accept_encoding.accepts(content_encoding.to_str().unwrap_or_default()) {
```
casey: Let's call it `is_acceptable`

src/subcommand/server/accept_encoding.rs
```diff
@@ -0,0 +1,116 @@
+use {super::*, axum::extract::FromRef};
+#[derive(Default)]
+pub(crate) struct AcceptEncoding(pub(crate) String);
```
casey, line 3: Let's make this an `Option<String>`, also, I think the inner value can be private.

src/subcommand/server/accept_encoding.rs
```diff
@@ -0,0 +1,116 @@
...
+        .headers
+        .get("accept-encoding")
+        .map(|value| value.to_str().unwrap_or_default().to_owned())
+        .unwrap_or_default(),
+    ))
+  }
+}
+
+impl AcceptEncoding {
+  pub(crate) fn accepts(self, encoding: &str) -> bool {
```
casey:
```suggestion
  pub(crate) fn is_acceptable(self, encoding: &str) -> bool {
```

casey (changes requested):
Very minor nits, plus looks like tests and lint are failing.

terror:
TODO:
- Switch to dropbox rust brotli implementation
- Handle these tests:
    - subcommand::server::tests::responses_are_brotlied
    - subcommand::server::tests::text_preview
    - subcommand::server::tests::text_preview_returns_error_when_content_is_not_utf8
    - subcommand::server::tests::text_preview_text_is_escaped

PR #1737, opened 2023-02-14
Add troubleshooting guide for syncing bitcoind

andrewtoth:
After working tech support on discord and twitter the past few days, I've added a troubleshooting guide for syncing bitcoind. These are some of the most common issues users new to running Bitcoin Core are having getting started.


casey (changes requested):
Nice! This looks like it depends on the PR that uses the REST interface, which isn't merged yet. Can you revise this to work with the current version of `ord`? All documentation should reflect what's been merged, not what may be merged in the future.

PR #1742, opened 2023-02-14
Add collecting with sparrow wallet guide to handbook

windsok:


raphjaph (approved):
LGTM!

refazi:
there is an issue with ord and the way it picks addresses from the wallet that can be "fixed" on sparrow with increasing the gap limit to 1000-2000, i think it's worth mentioning in the guide

PR #1751, opened 2023-02-15
Add `Dockerfile` to build Docker image

nevets963:
This PR adds a `Dockerfile` so that a Docker image can be built and therefore makes it easier to run `ord` in different environments, OSes, etc.

You can build the image using:
```
docker build -t casey/ord .
```

And run it using something like (on Umbrel):
```
docker run --rm --net umbrel_main_network -p 4000:80 -v $(pwd)/ord-data:/var/lib/ord \
  -v $(pwd)/umbrel/app-data/bitcoin/data/bitcoin/:/var/lib/bitcoind casey/ord \
  --data-dir /var/lib/ord --bitcoin-data-dir /var/lib/bitcoind --rpc-url 10.21.21.8 \
  --index-sats server --http
```
Note: I would not suggest people run `ord` using the above command. This is simply what I used to test the Docker image.


andrewtoth:
Have you tried https://github.com/LukeMathWalker/cargo-chef? It will make subsequent builds much faster. I recommend using it for the dockerfile.

RarePepeMaximalist:
This looks like a great start!

One thing we are going to have to figure out is the best way to pass in parameters for RPC host, data directory, etc.

Looking at how this is done for `electrs` on umbrel, they pass everything as environment variables into the docker image, and then the `elecrs` app is able to pickup these env vars and use them - https://github.com/getumbrel/umbrel-apps/blob/master/electrs/docker-compose.yml#L30-L36

We might want to do something similar for `ord` - ie. allow setting configuration variables via environment variable

nevets963:
`ord` could be changed to accept environment variables for configuration. However, this is not needed inorder to provide Docker support.

An example service within a `docker-compose.yaml` file could look like this (in the case of Umbrel):
```
services:
  ord:
    image: casey/ord:0.5.0
    restart: on-failure
    command: "--rpc-url ${APP_BITCOIN_NODE_IP} --index-sats --chain ${APP_BITCOIN_NETWORK} server --http"
    volumes:
      - ${APP_DATA_DIR}/data:/var/lib/ord
      - ${APP_BITCOIN_DATA_DIR}:/var/lib/bitcoind:ro
```

magicniko:
hey @nevets963 , I foolishly did not search for Dockerfile PRs before making mine, and also submitted a PR for a dockerfile which generates a smaller docker image, (https://github.com/casey/ord/pull/1765) and was wondering what you thought. I am happy to close my PR if it's not interesting. My docker image is 28MB and yours is 104MB, but likely more usable.

mroxso:
I would prefer a smaller image. But your PR does not include any README @magicniko

PR #1752, opened 2023-02-15
Github workflow to build and push Docker images on tag

nevets963:
When a `Dockerfile` is successfully merged (hopefully PR #1751), we can automate the build and push of the Docker image to Docker Hub.

This PR adds a GitHub workflow to automatically build `amd64` and `arm64` Docker images and push them to the Docker registry (with a tag, and to also push/update the `latest` image tag).

This PR is largely based on this: https://github.com/getumbrel/umbrel-manager/blob/fc823490591ea55e26734d144a0527fe9618166d/.github/workflows/on-tag.yml

You need to create the necessary Github secrets (for the Docker registry credentials): https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository

These need to be created:
- `DOCKER_USERNAME`
- `DOCKER_PASSWORD`
- `DOCKER_HUB_USER`

You can create a tag easily by following these instructions: https://stackoverflow.com/a/18223354
The workflow is setup to look for tags using semver. versioning which looks inline with the existing tags: https://github.com/casey/ord/tags


nmfretz:
@casey is there any interest in merging this + https://github.com/casey/ord/pull/1751 to make it easy for people to use official casey/ord docker images to run ord in different environments, OSs, etc?

Issue #1747, opened 2023-02-15
Make `--fee-rate` a mandatory argument for `wallet send`


veryordinally:
Per suggestion from RarePepeMaximalist.

Lots of transaction getting stuck in mempool at  default 1sat/vByte, which we may never see clear again.

I think it'd be better to make `--fee-rate` a required parameter. Ord users will have to learn about fee market anyways.

raphjaph:
Merging #1755 which will make `--fee-rate` mandatory. For this issue to be closed we need it for `wallet inscribe` as well.

casey:
Partially fixed by #1755. Closing in favor of #1432, which has details on a full fix.

veryordinally:
Created follow up issue #1767 for the `wallet inscribe` part.

Issue #1432, opened 2023-01-31
`--fee-rate` not respected


casey:
We had a user report that the `--fee-rate` option was not being respected.

The command was:

`ord wallet inscribe --fee-rate 6.0 INSCRIPTION.png`

Transaction: [7acd66e6f673e82999cedd37de5a4cbe41f217ae6f1dbf26a1ddfd6ef5488051](https://mempool.space/tx/7acd66e6f673e82999cedd37de5a4cbe41f217ae6f1dbf26a1ddfd6ef5488051)

Which paid a fee rate of 16.0 sats/vB

I saw a screenshot of the inscribe command output, and it showed fee of 7392. The commit and reveal transactions paid 4928 + 2464 = 7392, so `ord wallet` was reporting the correct feerate.

casey:
This is not actionable at the moment, because we haven't been able to reproduce it. Please let us know if you run into this!

casey:
This is the same underlying problem as #1597, we don't pass a fee to the send RPC command. Easy fix:

1. Write a failing test that shows that the `ord wallet send <AMOUNT>` command doesn't respect `--fee-rate`
2. Fix it by passing the correct fee rate to `sendtoaddress`

casey:
I believe this is now fixed.

papiro:
In trying to get to the bottom of something, I came across this thread. Not sure if I did something wrong since I am fairly new at this, but I specified a fee-rate of 16 and it used 16 for the commit transaction, but went ahead and spent the rest of my UTXO in the reveal bringing the fee rate to 18.8 for the reveal.

The command I ran: `ord wallet --server-url=http://localhost:8080 --name=ord inscribe --fee-rate=16 --file=5.txt`

`5.txt` is 136 bytes. My wallet had 15000 sats in it in a single UTXO.

Commit & reveal transactions: https://mempool.space/address/bc1pp3ytujrdsv72aak3rytme3x90awujaz4rq8442msf5wdx5qzsjzqqzhs8a

When I do the math, I should have gotten at least something back (in addition to the ordinal with the postage), but more specifically about 533 satoshis back. Instead, all I got was the ordinal with postage (10000 sats).

casey:
[...]

PR #1755, opened 2023-02-15
Remove default for parameter `--fee-rate` for `wallet send`

veryordinally:
Implement issue #1747 
Given high fee environment many users run into issue with default fee rate of 1. 

Review checklist:
- [x] Does it have tests?
- [x] Does this work for wallet send and wallet inscribe? Any other places where we pick a fee rate? (Fixed by noting in #1747 that we need to also do this for wallet inscribe.)
- [x] Have the docs been updated?

### DIFF
--- a/docs/src/guides/inscriptions.md
+++ b/docs/src/guides/inscriptions.md
@@ -205,7 +205,7 @@ ord wallet receive
 Send the inscription by running:
 
 ```
-ord wallet send ADDRESS INSCRIPTION_ID
+ord wallet send --fee-rate FEE_RATE ADDRESS INSCRIPTION_ID
 ```
 
 See the pending transaction with:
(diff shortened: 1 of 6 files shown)


RarePepeMaximalist:
This looks good to me -- I think we should also make `--fee-rate` a required parameter on `ord inscribe` ?

Also we should update the guide to use `--fee-rate` in the example
https://github.com/casey/ord/blob/master/docs/src/guides/inscriptions.md

veryordinally:
Updated the guide, thanks @RarePepeMaximalist for the pointer.

Update for `ord inscribe` makes sense to me for consistency.

droplister:
Please include in your review this piece of code. If you are sending an amount of sats rather than an inscription, it will use an RPC command that does not take a --fee-rate at all, but relies on the default behavior of `send_to_address`.

I think this means that even if you require a fee rate, it will not be utilized in this instance.

https://github.com/casey/ord/blob/b0ca48658d1b5eb911e1a2e3afea675134f9ba4d/src/subcommand/wallet/send.rs#L51-L74

casey:
Great point @droplister! This is actually the root problem behind #1432. So extremely nice find!

PR #1759, opened 2023-02-15
Batch tx requests

andrewtoth:
Fixes https://github.com/casey/ord/issues/1364 by reverting #1357.

This PR uses `hyper` to fetch raw transactions in the background with batched JSON-RPC requests sent 12 at a time concurrently. This saturates the default 4 rpc threads but doesn't exceed the default rpcworkqueue of 16. I managed to sync to block 776774 in 41 minutes.

This does not require any configuration changes. It uses the json-rpc endpoint, but more efficiently. All missing inputs are sent to a separate thread before indexing, and the resulting input values are sent back via a channel in order.


veryordinally (commented):
@raphjaph @casey Reviewed conceptually and talked to @andrewtoth - this approach looks consistent with what we discussed yesterday. Does not introduce new communication protocol, and massively speeds up the indexing process for when we have inscriptions (from a day+ to less than an hour). 

Let's discuss and see if you also ACK this approach, then @andrewtoth can complete this PR.

raphjaph:
Concept ACK from @veryordinally, @casey, and @raphjaph (reviewing together on a Discord call).

This seems like the right approach. It doesn't require any config changes and doesn't change our security model. 

Would be great to get the MVP in! Should be as simple as possible, with no configuration or additional knobs. Do not worry about polishing too much. Just get full test coverage, undraft, and request review.

andrewtoth:
@veryordinally @raphjaph @casey fixed broken test, ready for review.

Also, #1637 and #1731 are relevant to this as well. Since we're using hyper, the json-rpc `index.client` that is created at the start times out and fails when trying to fetch block count again. Same with json-rpc clients created in subcommands before a long update.

src/index.rs
```diff
@@ -20,6 +20,7 @@ use {
 
 mod entry;
 mod rtx;
+mod tx_fetcher;
```
raphjaph, line 23:
```suggestion
mod fetcher;
```

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
+use {
+  anyhow::{anyhow, Result},
+  bitcoin::{consensus::deserialize, Transaction, Txid},
+  bitcoincore_rpc::Auth,
+  hyper::{client::HttpConnector, Body, Client, Method, Request},
+  serde::Deserialize,
+};
+
+pub(crate) struct TxFetcher {
```
raphjaph, line 9:
```suggestion
pub(crate) struct Fetcher {
```

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+  id: String,
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonError {
+  code: i32,
+  message: String,
+}
+
+impl TxFetcher {
```
raphjaph, line 28:
```suggestion
impl Fetcher {
```

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+  pub(crate) fn new(url: &str, auth: Auth) -> Result<Self> {
+    if auth == Auth::None {
+      return Err(anyhow!("No authentication provided"));
+    }
+
+    let client = Client::new();
+
+    let url = if url.starts_with("http://") {
```
raphjaph, line 36: Let's turn url into a proper URL type.

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+      url.to_string()
+    } else {
+      "http://".to_string() + url
+    };
+
+    let (user, password) = auth.get_user_pass()?;
+    let auth = format!("{}:{}", user.unwrap(), password.unwrap());
+    let auth = format!("Basic {}", &base64::encode(auth));
+    Ok(TxFetcher { client, url, auth })
```
raphjaph, line 45:
```suggestion
    Ok(Fetcher { client, url, auth })
```

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+
+  pub(crate) async fn get_transactions(&self, txids: Vec<Txid>) -> Result<Vec<Transaction>> {
+    if txids.is_empty() {
+      return Ok(Vec::new());
+    }
+
+    let mut reqs = Vec::with_capacity(txids.len());
+    for (i, txid) in txids.iter().enumerate() {
+      let req =
+        format!("{{\"jsonrpc\":\"2.0\",\"id\":\"{i}\",\"method\":\"getrawtransaction\",\"params\":[\"{txid:x}\"]}}");
```
raphjaph, line 56: use `json!` macro instead of string

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub(crate) struct TxFetcher {
+  client: Client<HttpConnector>,
+  url: String,
+  auth: String,
+}
+
+#[derive(Deserialize, Debug)]
+struct JsonResponse {
+  result: Option<String>,
```
raphjaph, line 17: make result a non-opaque type

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+  error: Option<JsonError>,
+  id: String,
```
raphjaph, line 19: should this be `String` or `usize`?

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    if let Some(err) = results.iter().find_map(|res| res.error.as_ref()) {
+      return Err(anyhow!(
+        "Failed to fetch raw transaction: code {} message {}",
+        err.code,
+        err.message
+      ));
+    }
+
+    results.sort_by(|a, b| {
```
raphjaph, line 82: why do we need to sort?

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+      a.id
+        .parse::<usize>()
+        .unwrap()
+        .cmp(&b.id.parse::<usize>().unwrap())
+    });
+
```
raphjaph, line 88: add an assert if it's critical that it's sorted

src/index/tx_fetcher.rs
```diff
@@ -0,0 +1,96 @@
...
+    Ok(
+      results
+        .into_iter()
+        .map(|res| deserialize(&hex::decode(res.result.unwrap()).unwrap()).unwrap())
```
raphjaph, line 92: add comment why `unwraps` are safe

raphjaph, line 92: use a more qualified import statement for `deserialize`.

src/index/updater.rs
```diff
@@ -92,6 +98,8 @@ impl Updater {
 
     let rx = Self::fetch_blocks_from(index, self.height, self.index_sats)?;
 
+    let (mut outpoint_sender, mut value_receiver) = Self::spawn_tx_fetcher(index)?;
```
raphjaph, line 101:
```suggestion
    let (mut outpoint_sender, mut value_receiver) = Self::spawn_fetcher(index)?;
```

raphjaph (changes requested):
Reviewed by the peanut gallery in Discord: We are confident this fixes the issue but we would love any comments that will help us understand and maintain this code. If you could do that in addition to our review comments (if they are easy), we can smash merge ❤️

src/index.rs
andrewtoth: I was actually looking at extending this struct to fetch blocks, hashes, headers, and block count. Maybe `HyperJsonRpc` is a better name?

src/index/tx_fetcher.rs
andrewtoth: It's a string of hex though...

Cargo.toml
```diff
@@ -17,6 +17,7 @@ members = [".", "test-bitcoincore-rpc"]
 anyhow = { version = "1.0.56", features = ["backtrace"] }
 axum = { version = "0.6.1", features = ["headers"] }
 axum-server = "0.4.0"
+base64 = "0.13.1"
```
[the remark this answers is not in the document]
andrewtoth: This is for the rpc basic authentication.

src/index/tx_fetcher.rs
andrewtoth: Batched response can be returned in any order, so we need to sort. Our ids are in order so we make sure we send in the same order to the channel.

src/index/tx_fetcher.rs
andrewtoth: Sorry, not sure what I would be asserting here? It needs to be sorted before we return, but this is sorting it so we know it's sorted?

src/index/tx_fetcher.rs
andrewtoth: Some might not necessarily be, if bitcoind is misbehaving. I'll break this out with error checking.

andrewtoth:
@raphjaph addresses most of your feedback, asked some questions for clarification on some things.
I added comments to parts of the code that I think could use them. Let me know if there are any other places that could benefit from comments.

kvnn:
>  I managed to sync to block 776774 in 41 minutes.
@andrewtoth Nice! Do you mind sharing the machine specs?

andrewtoth:
@kvnn running pop_os with Intel® Core™ i7-10870H CPU @ 2.20GHz × 16. Bitcoind was running locally on the same machine.

src/index/tx_fetcher.rs
raphjaph: Thanks for clarifying!

src/index.rs
```diff
@@ -1026,6 +1027,58 @@ mod tests {
     }
   }
 
+  #[test]
+  fn inscriptions_below_first_inscription_height_are_skipped() {
+    let inscription = inscription("text/plain", "hello");
```
raphjaph, line 1032:
```suggestion
    let inscription = inscription("text/plain;charset=utf-8", "hello");
```

raphjaph (approved):
Thanks for addressing our comments! Merging this now.

Issue #1668, opened 2023-02-11
Make ord available in nixpkgs


neunenak:
I wrote a [simple nix package](https://gitea.everydayimshuflin.com/greg/ord-nix-package/src/branch/master/default.nix) that builds and installs `ord`, so I could install it on a NixOS machine I have. I can manually import this package and confirm that it installs a working executable on that machine.

I'd like to see `ord` available in nixpkgs, perhaps using the above code as a starting point. If that's not currently feasible, the above code could at least be made available to people who want to set up a similar manual import for their own NixOS configurations (along with manually updating the version numbers and hashes as `ord` development continues).

raphjaph:
I'm not very familiar with packaging for nix. Maybe @kodylow can help with that. He wanted to package it for nix in use in replit.

Kodylow:
let's get this shipped @neunenak , hit me in the discord tomorrow I'm kodylow#5554

neunenak:
@Kodylow I don't have a phone number I'm comfortable with giving to discord for the phone verification, happy to do this over github issues or some other chat that doesn't require phone verification (e.g. https://gitter.im/ )

Kodylow:
Kk can you open a PR with the nix script you wrote and tag me we'll continue there?

neunenak:
> Kk can you open a PR with the nix script you wrote and tag me we'll continue there?

Created the PR, can move further discussion there.

PR #1768, opened 2023-02-16
Add default.nix file

neunenak:


neunenak:
This PR is intended to get ord into nixpkgs, and close out https://github.com/casey/ord/issues/1668 . 


Tagging @Kodylow

default.nix
```diff
@@ -0,0 +1,26 @@
+{ lib, fetchFromGitHub , rustPlatform , pkg-config, openssl }:
+
+rustPlatform.buildRustPackage rec {
+  pname = "ord";
+  version = "0.5.0";
```
casey, line 5: Can we avoid hardcoding a version here? It would be nice if we didn't need to keep this version in sync with the `Cargo.toml`

default.nix
```diff
@@ -0,0 +1,26 @@
...
+
+  src = fetchFromGitHub {
+    owner = "casey";
+    repo = pname;
+    rev = version;
+    sha256 = "sha256-0atd/zW4NsTAmnJYjjlnB9rWnvK7d/ap9RgFRPqNrpc=";
```
casey, line 11: How do we update this hash? Can we avoid updating it manually?

default.nix
```diff
@@ -0,0 +1,26 @@
...
+  };
+
+  cargoSha256 = "sha256-vuMPk7TDXeyL4w/igO0IG6KSG2uzJftartEOWoMUjO8";
```
casey, line 14: Ditto with this hash.

default.nix
```diff
@@ -0,0 +1,26 @@
...
+
+  nativeBuildInputs = [pkg-config];
+  buildInputs = [openssl];
+
+  meta = with lib; {
+    description = "ord is an index, block explorer, and command-line wallet. Ordinal theory imbues satoshis with numismatic value, allowing them to collected and traded as curios.";
```
casey, line 20: Where is this copy from?

casey (changes requested):
What's the difference between a `shell.nix` and a `default.nix` file? I guess a `shell.nix` is for development, and a `default.nix` is for creating a package?

Also, how do we avoid having to keep a bunch of values in sync in the default.nix file? I've dealt with keeping package hashes in sync before, and it sucks super bad. If we don't have a way of keeping them in sync manually, then I won't bother to update the hashes on every release, and the `default.nix` will be stale. If that were the case, then I would move this to `contrib/default.nix`, but it would be nicer to keep them in sync automatically, or be able to omit them from the `default.nix`.

default.nix
neunenak: I copy-pasted this from some other piece of Ordinals documentation, https://github.com/casey/ord/blob/master/docs/src/introduction.md has part of it. This description can get changed to whatever makes sense.

neunenak:
> What's the difference between a `shell.nix` and a `default.nix` file? I guess a `shell.nix` is for development, and a `default.nix` is for creating a package?
> 

`default.nix` is the standard filename convention for a nix build file in the Nixpkgs repo. I don't think nix tooling cares about the exact filename. 

> Also, how do we avoid having to keep a bunch of values in sync in the default.nix file? I've dealt with keeping package hashes in sync before, and it sucks super bad. If we don't have a way of keeping them in sync manually, then I won't bother to update the hashes on every release, and the `default.nix` will be stale. If that were the case, then I would move this to `contrib/default.nix`, but it would be nicer to keep them in sync automatically, or be able to omit them from the `default.nix`.

I'm honestly not sure what the standard best practices are here for a project that wants to maintain its own nix build file outside of nixpkgs. Maybe @Kodylow has some insight here? In any case, I mostly made this PR because this nix build file worked for me and should work for other people if they manually update the hash and version, so maybe it makes sense to just put this in `contrib` with instructions about how to manually adjust the version.

casey:
@neunenak If we can't auto-update the hashes, let's put this in contrib and merge it as is. Seems fine to include and should be useful, even if the hashes get out-of-date.

casey:
I think we need to figure out a way to keep the version and hashes in sync before merging this.

Issue #1403, opened 2023-01-27
Use minimal data-pushes


ottosch:
Hey, I was playing around, writing a parser for inscriptions.
According to [the docs](https://docs.ordinals.com/inscriptions.html) I should find:
```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_1
  OP_PUSH "text/plain;charset=utf-8"
  OP_0
  OP_PUSH "Hello, world!"
OP_ENDIF
```

After the parser kept failing, I saw that [my example](https://mempool.space/tx/72907215a4e32cdbd26dcc5707daaddcf4f6b98a9971fdb6129a46065559226e) wasn't really using OP_1 (0x51) but 0x0101. Is there a reason for not using the smaller, standard script?

casey:
Nice! I think you might be the first alternative implementation. Really glad to see it.

>  Is there a reason for not using the smaller, standard script?

No, no reason! this is an oversight on my part. The `ord` encoder should use minimal data pushes, so I'll change the title of the issue to reflect that. I'm not sure if the parser accepts them (it uses rust-bitcoin) so I'll check.

casey:
If the current parser accepts them, then we can switch to emitting them right away. If the current parser does not accept them, then we'll have to do a staged rollout, where we start accepting them, wait for everyone to upgrade, and then start emitting them.

ottosch:
> If the current parser accepts them, then we can switch to emitting them right away.
I think it should... AFAIK bitcoin core nodes don't relay transactions that use 0x0101 instead of OP_1 due to transaction malleability.

> Nice! I think you might be the first alternative implementation. Really glad to see it.
It's nothing serious... we were in a group yesterday trying to extract the image from the transaction and came up with a bunch of piped `sed`, then today tried to make something more generic :)

ottosch:
There it is: https://gist.github.com/ottosch/f699ec13d598efb8a27e30197d9c517f

It's pretty silly. Can be called as `bitcoin-cli getrawtransaction 946f23e7b858e0f5739f363cebb5493f6eb7b9dacd09a609ad325aa5b93d189b | ./otto-parser.py` or `./otto-parser.py <raw_tx_file>`.  

It's already prepared for 0x0101 or OP_1

casey:
> There it is: https://gist.github.com/ottosch/f699ec13d598efb8a27e30197d9c517f

[...]

PR #1769, opened 2023-02-16
Correctly parse inscriptions when using minimal opcodes

arik-so:
Currently, minimal opcodes like OP_1 and OP_2 don't get recognized because all opcode operations that aren't OP_ENDIF result in an invalid inscription error, necessitating that all numeric opcodes be pushed as slices instead. However, that is in violation of [BIP 62](https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki) motivation §3.


tyjvazum:
@arik-so, could you provide a short description of the logic used in your changes?

mintu2121:
give me many pliz sir my ETH address
0x4aAaa2D70582508B067035A2e53cbFE41E935960

mintu2121:
please give me same many sir my ETH address is
0x4aAaa2D70582508B067035A2e53cbFE41E935960

mintu2121:
give me same many places sir my ETH address
0x4aAaa2D70582508B067035A2e53cbFE41E935960

mintu2121:
0x4aAaa2D70582508B067035A2e53cbFE41E935960

arik-so:
Certainly! In the code, or as a comment on the PR?

The point is basically that the documentation suggests using OP_1, OP_2, etc. for the various content property tags. However, the way it's currently being handled, OP_1 would not actually be recognized, because it would match the `Instruction::Op` arm. The issue is that rust-bitcoin's `push_slice` does not do minimal opcode detection, so pushing `&[1]` actually results in the hex encoding of `0101` instead of the opcode hex encoding `51` (81 in decimal). The actual opcode would be interpreted as invalid.

The set of numeric opcodes ranges from 1-16 (it just so happens that `OP_0` and `OP_FALSE` and the 0-byte-slice are all encoded identically as (`00`), so that does not cause any breakages.

tyjvazum:
Thank you for the detailed, clear explanation. I don't think code comments are necessary. However, I do have a couple other questions: what's the logic for using `Vec` instead of `u8`, and shouldn't this same change be made (for building as well as parsing) in the `append_reveal_script_to_builder` function, which currently uses `push_slice`?

arik-so:
To answer your first question, the reason I had to make it a `Vec` instead of a `slice` is that in the `ord` case, I'm instantiating a new vector, and so merely using a reference would result in the vector getting dropped with the closure of the match arm, a problem which the match arm that receives a reference to a vec with a longer lifetime doesn't have.

Regarding your other question, I don't think `append_reveal_script_to_builder` should be modified yet, because people running older versions of `ord` would then not see inscriptions generated after this change gets merged. I think that change should happen in due time, maybe a couple weeks or a month after the parsing becomes more permissive. Or perhaps even sooner if the software updates propagates more quickly.

shesek:
Couldn't this be solved instead by using [`Builder::push_int()`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/script/struct.Builder.html#method.push_int) instead of `Builder::push_slice()` for pushing the tags and [`Script::instructions_minimal()`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/script/struct.Script.html#method.instructions_minimal) instead of `Script::instructions()`?

arik-so:
That would fix the script construction, but not its parsing. As I pointed out in my previous comment, the construction should not be modified until the parsing fix propagates to most `ord` clients such that the inscriptions don't break.

It's a followup PR that I hope to make in a couple weeks or so, if/when this one gets merged.

shesek:
Why won't `instructions_minimal()` fix parsing?

arik-so:
Sorry, `instructions_minimal` would improve parsing (though it obviously still requires the `::Ops` change), but I strongly believe that the parsing fix should go out several weeks prior to the inscription change. I think though that even with `instructions_minimal`, the slice match arm of the parser would remain necessary for any tags above 16.

brandonblack:
Ping! Relaxing this parsing in `ord` seems important for future interoperability, even though it will renumber inscriptions.

tyjvazum:
@arik-so, I'm working on [arb](https://github.com/tyjvazum/arb). Any improvements you have in mind would be very welcome 👍

veryordinally:
> Any chance CI might be permitted to run on this PR?

@arik-so if you could update this PR we'll run CI. We'd like to integrate this for completing the cursed inscriptions handling.

veryordinally:
> Probably not until this upcoming weekend. If you have pointers towards similar unit tests though, they'd be greatly appreciated! Thanks!

This would be a test on the full mainnet chain so outside our unit test and integration test framework. I also won't get to this before the weekend, happy to collaborate on this then.

veryordinally (commented):
Looks good - will format and resolve conflicts, then verify inscription number correctness.

veryordinally:
@arik-so Sorry this took me a while to get to. Reviewed the cursed changes, looks good so far. I fixed the formatting so our CI is happy, but noticed your two tests are failing. Could you look into that?

veryordinally:
@arik-so Looked like the tests weren't updated for the cursed logic - please check if my changes make sense. Will now run inscription number validation, which will take a day or so.

veryordinally:
This currently has a very significant negative performance impact. Indexing mainnet has not finished after 48 hours (was at about block 400k), and usually finishes in 12h on the identical machine, so it looks like this is introducing a performance penalty of 10x. Needs optimization. @arik-so could you look into this?

veryordinally:
See also #1403

raphjaph:
We merged this: https://github.com/ordinals/ord/pull/2497

PR #1783, opened 2023-02-17
Add 4 nodes that relay inscription transactions

1440000bytes:
Adding these nodes as manual connections for peers when running a node for creating and broadcasting inscription tx would ensure tx is relayed properly. Some nodes on bitcoin p2p network are running a patch that rejects inscription txs in their mempool.

### DIFF
--- a/docs/src/guides/inscriptions.md
+++ b/docs/src/guides/inscriptions.md
@@ -58,6 +58,14 @@ index, add the following to your `bitcoin.conf`:
 
 ```
 txindex=1
+
+#Manual connections for relaying transactions without any issues
+addnode=y1.ordinals.yachts
+addnode=y2.ordinals.yachts
+addnode=y3.ordinals.yachts
+addnode=y4.ordinals.yachts
+
+mempoolfullrbf=1
 ```
 
 Or, run `bitcoind` with `-txindex`:


vicariousdrama (changes requested):
It's a good idea. Only recommendation here would be to enhance with a comment above the block providing rationale to the user

petertodd:
NACK

1. You should use your own DNS alias for this purpose. I've never made any promises that those nodes will relay inscriptions in the future.
2. There's no evidence that ordinal censorship is even widespread enough to matter. To have a high chance of ordinal transactions not propagating for nodes that don't listen and just make the standard 8 outgoing connections, you'd have to have more than a majority of nodes censoring them. For nodes that do listen, the situation is even better as they'll be even better connected.

docs/src/guides/inscriptions.md
```diff
@@ -58,6 +58,12 @@ index, add the following to your `bitcoin.conf`:
 
 ```
 txindex=1
+
+#manual connections for relaying transactions without any issues
+addnode=full-rbf1.btc.petertodd.net
+addnode=full-rbf2.btc.petertodd.net
+addnode=full-rbf3.btc.petertodd.net
+addnode=full-rbf4.btc.petertodd.net
```
petertodd, line 66:
```suggestion
addnode=full-rbf4.btc.petertodd.net
mempoolfullrbf=1
```

If you're going to connect to these full-rbf nodes, at least have the decency to also turn on full-rbf so they're not wasting connection slots.

petertodd (changes requested):

Psifour:
Concept NACK

I am opposed to making mempoolfullrbf standard (would support including it in further documentation regarding node optimization/configuration) and I am STRONGLY opposed to providing static peers as a suggestion anywhere within the documentation.

The defaults (configuration documentation establishes defaults) should NEVER modify the peering configuration of a node unless a user explicitly chooses to do so (while being fully informed of why this is typically discouraged). 

Assuming that the concept was approved, the proposed change (with Peter's addendum) is an appropriate way to encourage users to add these four specific peers to their nodes.

docs/src/guides/inscriptions.md
1440000bytes: Added it in the last commit

1440000bytes:
@petertodd 

> You should use your own DNS alias for this purpose. I've never made any promises that those nodes will relay inscriptions in the future.

Fixed DNS aliases in last commit. Sure, docs can always be updated and I added this after confirming on [twitter](https://twitter.com/1440000bytes/status/1626582033639297031) once.

> There's no evidence that ordinal censorship is even widespread enough to matter. To have a high chance of ordinal transactions not propagating for nodes that don't listen and just make the standard 8 outgoing connections, you'd have to have more than a majority of nodes censoring them. For nodes that do listen, the situation is even better as they'll be even better connected.

Agree, its not an issue at the moment however it might be too late to act if most nodes censor them at some point. It would affect not just inscriptions but bitcoin as well. So, I wanted to use these nodes as a precaution or best practice. I could have added other nodes here although I am assuming these nodes would be better connected to mining pools compared to others.

I had even reached out to some mining pools to get node addresses that could help, however there was no response.

@Psifour 

> I am opposed to making mempoolfullrbf standard (would support including it in further documentation regarding node optimization/configuration) and I am STRONGLY opposed to providing static peers as a suggestion anywhere within the documentation.

I dont see any downside of using mempoolfullrbf particularly for nodes running to just create inscriptions.

Can you explain the reasons for STRONG opposition as I dont think there are any issues with 4 manual connections to these nodes apart from automatic connections that bitcoin core makes (outbound/inbound)

> The defaults (configuration documentation establishes defaults) should NEVER modify the peering configuration of a node unless a user explicitly chooses to do so (while being fully informed of why this is typically discouraged).

Manual connections are used in many bitcoin projects particularly the ones in alpha or beta phase. These aren't connections to some random nodes online. Peter Todd has been contributing to Bitcoin since years and also runs one of the DNS seeds used in bitcoin core.

petertodd:
> @petertodd
> 
> > You should use your own DNS alias for this purpose. I've never made any promises that those nodes will relay inscriptions in the future.
> 
> Fixed DNS aliases in last commit. Sure, docs can always be updated and I added this after confirming on [twitter](https://twitter.com/1440000bytes/status/1626582033639297031) once.

Thanks.

With this change, ACK.

1440000bytes:
https://twitter.com/provoost/status/1641866584426790912

Cc: @casey @rot13maxi @raphjaph @RarePepeMaximalist @cbspears @Psifour @p

PR #1785, opened 2023-02-17
Add remote rpc support

eddiewang:
Never written rust before, so might need some polish here before it's merge ready.
Mainly quality of life improvements for users that aren't using a localhost Bitcoin node.

- Adds rpc_user and rpc_pass arguments to the binary
- Uses rpc_user / rpc_pass if arguments exist, otherwise defaults back to cookie mode
- Ensures all wallet commands properly pass the /wallet/<name> even if the rpc_url is set

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -146,7 +146,7 @@ impl Index {
     let auth = if !rpc_pass.is_none() && !rpc_user.is_none() {
       let ruser = rpc_user.unwrap();
       let rpass = rpc_pass.unwrap();
-      log::info!("Using RPC credentials from command line {ruser} and {rpass}");
+      log::info!("Using RPC credentials from command line {ruser}");
 
       Auth::UserPass(ruser.clone(), rpass.clone())
     } else {
--- a/src/index.rs
+++ b/src/index.rs
@@ -133,14 +133,30 @@ impl<T> BitcoinCoreRpcResultExt<T> for Result<T, bitcoincore_rpc::Error> {
 impl Index {
   pub(crate) fn open(options: &Options) -> Result<Self> {
     let rpc_url = options.rpc_url();
+    let rpc_user = options.rpc_user();
+    let rpc_pass = options.rpc_pass();
     let cookie_file = options.cookie_file()?;
+    let cookie_file_path = cookie_file.clone();
 
-    log::info!(
-      "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
-      cookie_file.display()
-    );
 
+    // by default we use a cookie file
     let auth = Auth::CookieFile(cookie_file);
+
+    // if options.rpc_pass and rpc_user is not empty, use Auth::UserPass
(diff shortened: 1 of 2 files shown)


eddiewang:
Right now, the logic will default to cookie unless both arguments are provided.
My understanding of how `.is_none()` works is that if the value is set, even if it's to an empty string `""`, it should still return `False`. 

If my assumption is correct, I think the desired state is to explicitly pass in both the user and pass in order to activate this mode, defaulting to the cookie approach otherwise.

src/index.rs
```diff
@@ -133,15 +133,31 @@ impl<T> BitcoinCoreRpcResultExt<T> for Result<T, bitcoincore_rpc::Error> {
...
+      Auth::UserPass(ruser.clone(), rpass.clone())
+    } else {
+      log::info!(
+        "Connecting to Bitcoin Core RPC server at {rpc_url} using credentials from `{}`",
+        cookie_file_path.display()
+      );
+      auth
+    };
+
+
```
coconutpalms, line 160: I think you need to make similar changes around L161 of options.rs

src/index.rs
```diff
@@ -134,15 +134,31 @@ impl<T> BitcoinCoreRpcResultExt<T> for Result<T, bitcoincore_rpc::Error> {
...
-    );
 
+    // by default we use a cookie file
     let auth = Auth::CookieFile(cookie_file);
 
+    // if options.rpc_pass and rpc_user is not empty, use Auth::UserPass
+    let auth = if !rpc_pass.is_none() && !rpc_user.is_none() {
+      let ruser = rpc_user.unwrap();
+      let rpass = rpc_pass.unwrap();
+      log::info!("Using RPC credentials from command line {ruser}");
```
AngusP, line 150: `rpc_pass` and `rpc_user` are `Option<String>`, which you can get at in a nicer way than checking `! is_none()` then calling `unwrap()` (calling unwrap is generally (strongly) discouraged), by using [`if let`](https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html) and [destructuring](https://doc.rust-lang.org/rust-by-example/flow_control/match/destructuring/destructure_tuple.html):

```rust
let number = Some(7);
let another = Some(12);
let missing: Option<i32> = None;

if let (Some(i), Some(j)) = (number, another) {
    println!("Matched {:?} and {:?}!", i, j);
}
// ^ Prints "Matched 7 and 12!"

if let (Some(i), Some(j)) = (number, missing) {
    println!("Matched {:?} and {:?}!", i, j);
} else {
    println!("Didn't match");
}
// ^ Prints "Didn't match"
```

so something like

```suggestion
    if let (Some(rpass), Some(ruser))  = (rpc_pass, rpc_user) {
      log::info!("Using RPC credentials from command line {ruser}");
```

(Caveat I'm also a Rust noob so if the compiler complains, sorry!)

AngusP:
Also note #1527 was trying to add this as well

raphjaph:
We added this check out PR #1527

Issue #887, opened 2022-12-05
Inscription Metadata


casey:
This isn't really a single feature, but I want to collect current and future metadata fields in one place:

- [x] Type: MIME type that can be served as `Content-Type`
- [x] Resource: Payload bytes that can be served as HTTP resource
- [x] Magic number: Allow ord to skip non-ord containers, and allow protocol authors to utilize containers for other protocols
- [ ] Symbol: Human readable identifier for inscription #794 
- [ ] Encoding: String that can be served as `Content-Encoding` #891
- [ ] Link: Some kind of hash that can be used to look up off-chain content. #801 
- [ ] License: SHA-256 hash of content license. Not sure if this would be legally effective. #785

Issue #801, opened 2022-11-15
Off-chain content


casey:
Should we support off-chain content?

Pros:
- Cheaper to make inscriptions
- Possible to link to very large content

Cons:
- Increases confusion for users. Some inscriptions would have content on chain which would be nearly guaranteed not to be lost, whereas others would be off chain and would rely on economic incentives of inscription holders.
- Lots of complexity: Which hash function? Should we support arbitrary URLs? How store?

SHA-256, BLAKE3, or BitTorrent V2 file merkle root. If BitTorrent V2, will also need content length

lingmann:
If you decided to support off-chain content, it's probably worth looking at the [CID Specification](https://github.com/multiformats/cid). They have spent some time thinking about hash functions, codecs, etc. and it is being used by Bluesky (AT Protocol) as well as IPFS.

mullojo:
Seems smart to keep the main chain as light as possible for long term success.   IPFS is a very attractive solution for off-chain data storage.

casey:
Not to sound harsh, but I've looked at and worked with IPFS, and is a bloated, complex, and poorly-designed. BitTorrent V2 is a much better, simpler, sounder protocol.

Requirements for any off-chain storage:

- Content must be referenced with a standard hash function digest that is not protocol specific. SHA-256 is attractive, because it does not introduce any new security assumptions, since Bitcoin already relies on SHA-256. BitTorrent V2's SHA-256 merkle tree is nonstandard, but is so simple that I'll give it a pass. BLAKE3 is attractive for its extremely fast hashing, but it introduces a new security assumption, that BLAKE3 is secure. IPFS multihashes are eliminated by this requirement, because they are linked to that specific protocol, in addition to additional issues with IPFS.

- It must be possible to inform the user of the health of content stored off chain. How many copies are there? Are they complete? This can only be accomplished by actually attempting to fetch random chunks of the content. The approach of torrent sites is reasonable here, we could display seeders and leachers. Fetching random chunks of the content eliminates IPFS again, because the protocol is so complex that we cannot implement ourselves, and it's such a heavy dependency that we wouldn't want to pull it in.

- It must be easy for users to download content that they care about, and serve it for other users. "Easy" in this case doesn't mean that the user can do it, but it requires multiple manual steps and additional software. Users must be able to automatically, using the software they already have installed, namely `ord`, fetch, store, and replicate off-chain content.

mullojo:
[...]

Issue #624, opened 2022-10-06
Metadata Format


casey:
Anything we want to attach to ordinals needs to be serialized. Bitcoin doesn't have a native, general-purpose serialization format. I mean we could use Bitcoin script… ah ha ha, just kidding… unless 😳

| value | script | pushdata |
| --- | --- | --- |
| hash | `OP_0 PUSHDATA` | 32-byte SHA-256 hash | | 
| number | `OP_1 PUSHDATA` | little-endian bytes |
| byte string | `OP_2 PUSHDATA` | bytes |
| text string | `OP_3 PUSHDATA` | UTF-8 encoded text |
| array | `OP_4 PUSHDATA ELEMENT*` | number of elements |
| object | `OP_5 PUSHDATA (KEY VALUE)*` | number of fields |

The encoding is not the script, it's what it pushes, parsed starting at the bottom of the stack.

Notes:
- Byte and text strings are different types, so that text strings unambiguously map to human readable and writable strings.
- Arrays and objects are different types, so that object keys don't need a tag and can be raw byte strings, and for readability when rendered as JSON.
- When rendered to JSON, byte strings are represented as `0x…` literals. This is intentionally incompatible with JSON.
- Canonicality: Integers may not have leading zeroes, strings must be minimally encoded, and object keys must be unique and in ascending order.
- User defined types should be serialized as arrays or objects.
- Object keys are untagged byte arrays.
- Array and object lengths are pushed up front, so parser can allocate space up front.
- Use BitTorrent v2's tree hash as the hashing mode. This way, any subtree of an inscription can be retrieved via bittorrent. (Not that the hash is not the infohash of the torrent, it's the hash of the file within the torrent.) Could also not, and use SHA-256, and not try to predict off-chain storage of content.

Questions:
- How are negative and positive numbers represented?
- How is hashing mode defined? (Tag, length of 
- Can we support inclusion proofs?
- Display keys from 0 to 127 are displayed as ascii strings w/escapes, values from 128 to 255 are displayed as hex literals?
- Is WScriptHash an appropriate hash type for such a script?
- Do we need a chunked encoding for byte and text strings to avoid hitting 520 byte pushdata size limit?
- A much simpler encoding would be `OP_FALSE OP_IF <a bunch of data pushes> OP_ENDIF` and putting everything in those data pushes which are concatinated together to get the payload. That's kinda ugly though.
- What is the easiest to manipulate in script?

casey:
[...]

PR #1805, opened 2023-02-20
Add option to inscribe BitTorrent (V1) infohashes

shesek:
This adds a new `--torrent` CLI flag (plus some related `--torrent-xxx` options) to `ord wallet inscribe`, which automatically creates a torrent for the given file (by default, in the same directory with a `.torrent` extension) and inscribes its SHA-128 infohash alongside the file's (regular) SHA-256 using the `sha2,btih` content type.

The explorer was also adapted to render torrent inscriptions with the infohash/sha256 hex and a magnet link.

With this, files can be stored off-chain on the BitTorrent P2P network while the Bitcoin blockchain is used to commit to the file and to provide its `infohash`, which acts as a locator for obtaining the file from BitTorrent (using DHT or trackers).

---

Why not IPFS?

I haven't really found a compelling reason to prefer IPFS. It can work better for files that are duplicated across multiple torrents but it should rarely be the case here. Also bittorrent has a more mature ecosystem, better tooling, and more people already have a client installed.

Why not BitTorrent V2?

I couldn't find a Rust library that supports it. Also it appears that many bittorrent clients don't support V2 yet, including [webtorrent](https://github.com/webtorrent/webtorrent) (which is pretty neat and worth supporting imo).

Why inscribe the SHA-256 too?

Mainly to have an easy way for users to verify the file if they get it through some other non-torrent means (and therefore don't have the torrent metadata, which is necessary to confirm the infohash matches the file -- edit: [not quite true](https://github.com/casey/ord/pull/1805#issuecomment-1439371596)) and to avoid relying solely on the SHA-128 used by BitTorrent V1. This also enables the Nostr storage mentioned below.

---

It's possible to improve file availability and download speeds by providing some centrally managed services:

- A tracker. Having a tracker specifically for ordinal/inscription users could help improve peer discovery. The tracker could be set to only accept infohashes that were inscribed on the blockchain, so that it couldn't be used as a free-for-all tracker (with [webtorrent/bittorrent-tracker](https://github.com/webtorrent/bittorrent-tracker), the `filter` callback can be used to check with an external database/api).

- A DHT bootstrap node. Having ordinal users connect to it can similarly help improve connectivity. (some possible options are [bittorrent/bootstrap-dht](https://github.com/bittorrent/bootstrap-dht) and [webtorrent/bittorrent-dht](https://github.com/webtorrent/bittorrent-dht))

- A WebSeed mirror. [BEP 19](https://www.bittorrent.org/beps/bep_0019.html) defines a standard for adding HTTP mirrors to torrents, which can be used in addition to normal peers or as a fallback if there are none.

  The ordinals.com server (or whoever else is interested in providing a mirror) would need to run a daemon that: (1) Listens for infohash inscriptions, (2) Downloads the torrents associated with them (probably with some size limit), and (3) Makes them available for download over HTTP.

  It appears that [webtorrent](https://github.com/webtorrent/webtorrent) can be used to achieve this quite easily. It even includes a [built-in http server](https://github.com/webtorrent/webtorrent/blob/master/docs/api.md#clientcreateserveropts-force) that could be used as the webseed server [0].

This PR assumes that such services exists (on `tracker.ordinals.com`/`dht.ordinals.com`/`webseed.ordinals.com`), and embeds them into the generated `.torrent` files and magnet links. If you'd like me to, I can remove this.

[0] I submitted https://github.com/webtorrent/webtorrent/pull/2487 to add support for serving the `.torrent` file, which this PR assumes to be available on the webseed server at `<WEBSEED_URL>/<INFOHASH>.torrent`. This makes magnet links work even when there are no p2p seeds.

---

Using Nostr for storage

This can be achieved with this PR by inscribing an `event.json` file with the Nostr event in the [JSON serialization format](https://github.com/nostr-protocol/nips/blob/master/01.md#events-and-signatures) used for event IDs. For example:

```bash
echo -n '[0,"mypubkey",1676880792,1,[],"hello nostr!"]' > event.json
ord wallet inscribe --torrent event.json
```

This will inscribe the Nostr event ID as the committed SHA-256, which can then be looked up on Nostr relays. The inscribed torrent can be used as a backup, or ignored.

---

Possible future work:

- Display file previews directly in the browser, either via a webseed http mirror [1] or using an in-browser webtorrent client (its pretty neat that the torrent can be downloaded right in the browser! -- although with the caveat that it can only connect to other [WebRTC-enabled](https://github.com/webtorrent/webtorrent/issues/369) peers. See [instant.io](https://instant.io) for a cool demonstration of it.)

- Because the inscriptions here are pretty small, this could theoretically use `OP_RETURN`, which would avoid the complexities of the two commit/reveal transactions.

[1] I submitted https://github.com/webtorrent/webtorrent/pull/2488 to make embedding files from the webseed slightly easier.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ html-escaper = "0.2.0"
 http = "0.2.6"
 hyper = { version = "0.14.24", features = ["http1", "client"] }
 indicatif = "0.17.1"
+lava_torrent = { version = "0.8.1" }
 lazy_static = "1.4.0"
 log = "0.4.14"
 mime = "0.3.16"
@@ -56,6 +57,7 @@ tokio = { version = "1.17.0", features = ["rt-multi-thread"] }
 tokio-stream = "0.1.9"
 tokio-util = {version = "0.7.3", features = ["compat"] }
 tower-http = { version = "0.3.3", features = ["compression-br", "compression-gzip", "cors", "set-header"] }
+urlencoding = "2.1.2"
 
 [dev-dependencies]
 executable-path = "1.0.0"
--- a/src/inscription.rs
+++ b/src/inscription.rs
@@ -105,6 +105,14 @@ impl Inscription {
     str::from_utf8(self.content_type.as_ref()?).ok()
   }
 
+  pub(crate) fn torrent_data(&self) -> Option<super::torrent::TorrentHashData> {
+    if self.content_type() == Some("sha2,btih") {
+      super::torrent::parse_inscription_data(self.body()?)
+    } else {
+      None
+    }
+  }
+
   #[cfg(test)]
   pub(crate) fn to_witness(&self) -> Witness {
     let builder = script::Builder::new();
--- a/src/inscription.rs
+++ b/src/inscription.rs
@@ -23,7 +23,6 @@ pub(crate) struct Inscription {
 }
 
 impl Inscription {
-  #[cfg(test)]
   pub(crate) fn new(content_type: Option<Vec<u8>>, body: Option<Vec<u8>>) -> Self {
     Self { content_type, body }
   }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -119,6 +119,7 @@ mod sat_point;
 pub mod subcommand;
 mod tally;
 mod templates;
+mod torrent;
 mod wallet;
 
 type Result<T = (), E = Error> = std::result::Result<T, E>;
--- a/src/media.rs
+++ b/src/media.rs
@@ -13,6 +13,7 @@ pub(crate) enum Media {
   Text,
   Unknown,
   Video,
+  HashTorrent,
 }
 
 impl Media {
@@ -37,6 +38,7 @@ impl Media {
     ("text/plain;charset=utf-8", Media::Text, &["txt"]),
     ("video/mp4", Media::Video, &["mp4"]),
     ("video/webm", Media::Video, &["webm"]),
+    ("sha2,btih", Media::HashTorrent, &[]),
   ];
 
   pub(crate) fn content_type_for_path(path: &Path) -> Result<&'static str, Error> {
--- a/src/subcommand/preview.rs
+++ b/src/subcommand/preview.rs
@@ -86,6 +86,10 @@ impl Preview {
             dry_run: false,
             no_limit: false,
             destination: None,
+            torrent: false,
+            torrent_path: None,
+            torrent_tracker: "".to_string(),
+            torrent_peers: "".to_string(),
           },
         )),
       }
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -8,7 +8,8 @@ use {
   crate::templates::{
     BlockHtml, ClockSvg, HomeHtml, InputHtml, InscriptionHtml, InscriptionsHtml, OutputHtml,
     PageContent, PageHtml, PreviewAudioHtml, PreviewImageHtml, PreviewPdfHtml, PreviewTextHtml,
-    PreviewUnknownHtml, PreviewVideoHtml, RangeHtml, RareTxt, SatHtml, TransactionHtml,
+    PreviewTorrentHashHtml, PreviewUnknownHtml, PreviewVideoHtml, RangeHtml, RareTxt, SatHtml,
+    TransactionHtml,
   },
   axum::{
     body,
@@ -809,6 +810,13 @@ impl Server {
       }
       Media::Unknown => Ok(PreviewUnknownHtml.into_response()),
       Media::Video => Ok(PreviewVideoHtml { inscription_id }.into_response()),
+      Media::HashTorrent => {
+        let torrent_data = inscription
+          .torrent_data()
+          .ok_or_not_found(|| format!("inscription {inscription_id} torrent commitment data"))?;
+
+        Ok(PreviewTorrentHashHtml(torrent_data).into_response())
+      }
     };
   }
 
--- a/src/subcommand/wallet/inscribe.rs
+++ b/src/subcommand/wallet/inscribe.rs
@@ -1,6 +1,6 @@
 use {
   super::*,
-  crate::wallet::Wallet,
+  crate::{torrent, wallet::Wallet},
   bitcoin::{
     blockdata::{opcodes, script},
     policy::MAX_STANDARD_TX_WEIGHT,
@@ -54,11 +54,37 @@ pub(crate) struct Inscribe {
   pub(crate) dry_run: bool,
   #[clap(long, help = "Send inscription to <DESTINATION>.")]
   pub(crate) destination: Option<Address>,
+
+  #[clap(long, help = "Create a torrent for <FILE> and inscribe its infohash")]
+  pub(crate) torrent: bool,
+  #[clap(long, help = "Path to save .torrent file [default: <FILE>.torrent]")]
+  pub(crate) torrent_path: Option<PathBuf>,
+  #[clap(
+    long,
+    help = "Torrent tracker URL to include in .torrent file",
+    default_value = torrent::DEFAULT_TRACKER,
+  )]
+  pub(crate) torrent_tracker: String,
+  #[clap(
+    long,
+    help = "Torrent peers to include in .torrent file (space-separated <host>:<port>)",
+    default_value = torrent::DEFAULT_PEER,
+  )]
+  pub(crate) torrent_peers: String,
 }
 
 impl Inscribe {
   pub(crate) fn run(self, options: Options) -> Result {
-    let inscription = Inscription::from_file(options.chain(), &self.file)?;
+    let inscription = if !self.torrent {
+      Inscription::from_file(options.chain(), &self.file)?
+    } else {
+      torrent::make_torrent_inscription(
+        self.file,
+        self.torrent_path,
+        &self.torrent_tracker,
+        &self.torrent_peers,
+      )?
+    };
 
     let index = Index::open(&options)?;
     index.update()?;
--- a/src/templates.rs
+++ b/src/templates.rs
@@ -11,8 +11,8 @@ pub(crate) use {
   output::OutputHtml,
   page_config::PageConfig,
   preview::{
-    PreviewAudioHtml, PreviewImageHtml, PreviewPdfHtml, PreviewTextHtml, PreviewUnknownHtml,
-    PreviewVideoHtml,
+    PreviewAudioHtml, PreviewImageHtml, PreviewPdfHtml, PreviewTextHtml, PreviewTorrentHashHtml,
+    PreviewUnknownHtml, PreviewVideoHtml,
   },
   range::RangeHtml,
   rare::RareTxt,
--- a/src/templates/preview.rs
+++ b/src/templates/preview.rs
@@ -27,3 +27,6 @@ pub(crate) struct PreviewUnknownHtml;
 pub(crate) struct PreviewVideoHtml {
   pub(crate) inscription_id: InscriptionId,
 }
+
+#[derive(boilerplate::Boilerplate)]
+pub(crate) struct PreviewTorrentHashHtml(pub(crate) super::torrent::TorrentHashData);
--- a/src/torrent.rs
+++ b/src/torrent.rs
@@ -1,0 +1,95 @@
+use {
+  super::{inscription::Inscription, Error},
+  anyhow::Context,
+  bitcoin::hashes::{sha256, Hash},
+  lava_torrent::{bencode::BencodeElem, torrent::v1::TorrentBuilder},
+  std::path::{Path, PathBuf},
+  std::{ffi::OsString, fs},
+  urlencoding::encode,
+};
+
+use bitcoin::hashes::hex::ToHex;
+
+// Content type to identify Bittorent v1 hash commitments (v1 is implied by btih, v2 uses btmh)
+const CONTENT_TYPE: &str = "sha2,btih";
+
+// Bittorrent piece length (1MB)
+const PIECE_LENGTH: i64 = 1048576;
+
+// Default tracker URIs (included in .torrent & magnet links)
+// TODO support multiple, add wss tracker
+pub const DEFAULT_TRACKER: &str = "udp://tracker.ordinals.com";
+
+// Default webseed URL (included in .torrent & magnet links)
+// Expected to serve files at WEBSEED_URL/INFOHASH/FILENAME and the .torrent metadata file at WEBSEED_URL/INFOHASH.torrent
+pub const DEFAULT_WEBSEED: &str = "https://webseed.ordinals.com";
+
+// Bootstrap peers for DHT discovery (included in .torrent & magnet links)
+pub const DEFAULT_PEER: &str = "dht.ordinals.com:6885";
+
+pub(crate) fn make_torrent_inscription(
+  file_path: impl AsRef<Path>,
+  torrent_path: Option<impl AsRef<Path>>,
+  tracker_url: &str,
+  peer_addr: &str,
+) -> Result<Inscription, Error> {
+  // TorrentBuilder requires absolute paths
+  let file_path = fs::canonicalize(file_path)?;
+
+  // Create the torrent and gets its infohash
+  let torrent = TorrentBuilder::new(&file_path, PIECE_LENGTH)
+    .set_announce(Some(tracker_url.to_string()))
+    .add_extra_field("nodes".to_string(), bencode_nodes(peer_addr))
+    .build()
+    .with_context(|| "TorrentBuilder failed")?;
+  let infohash = torrent.info_hash_bytes();
+
+  // Write the .torrent file (by default, to <path>.torrent)
+  let torrent_path = get_torrent_path(&file_path, torrent_path);
+  log::info!(
+    "Writing torrent with infohash {} to {}",
+    infohash.to_hex(),
+    torrent_path.display()
+  );
+  torrent
+    .write_into_file(torrent_path)
+    .with_context(|| "failed writing .torrent file")?;
+
+  // Calculate the file's SHA256
+  // TODO streaming hash to avoid loading the entire file in memory
+  let contents =
+    fs::read(&file_path).with_context(|| format!("io error reading {}", file_path.display()))?;
+  let sha256hash = sha256::Hash::hash(&contents).into_inner().to_vec();
+
+  // Create an inscription for the SHA256+infohash
+  Ok(Inscription::new(
+    Some(CONTENT_TYPE.into()),
+    Some([sha256hash, infohash].concat()),
+  ))
+}
+
+fn get_torrent_path(file_path: &Path, torrent_path: Option<impl AsRef<Path>>) -> PathBuf {
+  if let Some(torrent_path) = torrent_path {
+    torrent_path.as_ref().to_owned()
+  } else {
+    let mut fileoss: OsString = file_path.into();
+    fileoss.push(".torrent");
+    fileoss.into()
+  }
+}
+
+fn bencode_nodes(nodes: &str) -> BencodeElem {
+  BencodeElem::List(
+    nodes
+      .trim()
+      .split(" ")
+      .filter_map(|node| {
+        let mut parts = node.split(":"); // host:port
+        Some(BencodeElem::List(vec![
+          parts.next()?.into(),
+          parts.next()?.parse::<i64>().ok()?.into(),
+        ]))
+      })
+      .collect(),
+  )
+}
--- a/src/torrent.rs
+++ b/src/torrent.rs
@@ -68,6 +68,50 @@ pub(crate) fn make_torrent_inscription(
   ))
 }
 
+pub fn parse_inscription_data(data: &[u8]) -> Option<TorrentHashData> {
+  if data.len() != 32 + 20 {
+    None
+  } else {
+    Some(TorrentHashData {
+      sha256: data[0..32].to_vec(),
+      infohash: data[32..].to_vec(),
+    })
+  }
+}
+
+pub struct TorrentHashData {
+  pub sha256: Vec<u8>,
+  pub infohash: Vec<u8>,
+}
+
+impl TorrentHashData {
+  pub fn magnet_uri(&self) -> String {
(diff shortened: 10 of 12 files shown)


1440000bytes:
> This adds a new --torrent CLI flag (plus some related --torrent-xxx options) to ord wallet inscribe, which automatically creates a torrent for the given file (by default, in the same directory with a .torrent extension) and inscribes its SHA-128 infohash alongside the file's (regular) SHA-256 using the sha2,btih content type.

This only makes the on-chain inscriptions more valuable and I am not sure if 'inscription' would be the write word for link to off-chain things.

> Why not BitTorrent V2?

> I couldn't find a Rust library that supports it. Also it appears that many bittorrent clients don't support V2 yet, including [webtorrent](https://github.com/webtorrent/webtorrent) (which is pretty neat and worth supporting imo).

Insecure because of SHA1: https://www.bittorrent.org/beps/bep_0052.html

> It's possible to improve file availability and download speeds by providing some centrally managed services:

Unnecessary complexity and dependencies.

> Using Nostr for storage

This makes no sense because:

- Free relays have limits
- Users would prefer to pay bitcoin miners instead of paid relays
- Nostr relays do not guarantee events to be stored forever

shesek:
> > Why not BitTorrent V2?

> Vulnerable because of SHA1: https://www.bittorrent.org/beps/bep_0052.html

As mentioned in the PR description, one of the reasons I included the SHA-256 alongside the infohash was to avoid relying solely on SHA-128.

Once the user finishes downloading the file from BitTorrent using the committed infohash, they can verify that it also matches the committed SHA-256.

> > It's possible to improve file availability and download speeds by providing some centrally managed services:

> Unnecessary complexity

Well I guess it depends on what your goals are :)

I would say that the tracker/dht are pretty straightforward to setup and require relatively little resources.

The web seed, however, is more complicated and resource-intensive (storage mainly, although its not really that different than full nodes storing files inscribed on the blockchain).

But it seems to me that enough people care about ordinals inscriptions so that at least a few would like to auto-seed inscribed infohashes and provide a public mirror/archive. It doesn't have to be official.

> > Using Nostr for storage

> This makes no sense because:

Using Nostr is just an option, which can be used alongside BitTorrent as a backup. Sort of like a web seed.

> Free relays have limits

For some use cases, the contents could be small enough to fit within the limits.

> Users would prefer to pay bitcoin miners instead of paid relays

Well, even if your assumptions are true, users might still want to inscribe something that was already published (and paid for) on a Nostr relay for a different purpose. For example some people were NFTing tweets awhile back.

shesek:
> [ .. SHA-256 is included ..] to have an easy way for users to verify the file if they get it through some other non-torrent means (and therefore don't have the torrent metadata, which is necessary to confirm the infohash matches the file)

Actually, thinking about this some more, assuming the user knows the filename and piece size used to create the torrent (in this PR the piece size is static), they can deterministically reproduce the `info` metadata from the file and compute its `infohash` to verify it.

This is however quite cumbersome, having the SHA-256 directly does make it much easier. And SHA-256 is still preferable over the SHA-128 infohash for security reasons.

petertodd:
NACK

The incentive model for people to seed BitTorrent content is that the data itself is valuable. This is not clearly the case with Ordinals, which represent ownership of something. Thus relying on BitTorrent is likely to result in data loss.

petertodd:
> Well to be fair there isn't really an incentive model for seeding on BitTorrent. You have an incentive to obtain the data for yourself, but not to make it available for others.

You're using the word "incentive" in a very narrow way here, only referring to selfish economic incentives. People have other incentives too, such as doing good for their community. In that sense of the term, there is a clear incentive to running BitTorrent seeds, as a wide community of people benefit from them. That incentive is missing for ordinal data.

> Even if your arguments are all correct, what's the harm in making this available as an option and have users choose?

The high chance it breaks of course. It does not make sense for Ordinals to ship features that are likely to break and harm the user experience. Additionally, every feature we add imposes a UI/UX cost, which further harms adoption.

A good question really is what is *your* incentive for proposing this feature? Are you actually part of the ordinals community? Or are you proposing it because you want that community to do something that benefits some other community?

petertodd:
> > The high chance it breaks of course. It does not make sense for Ordinals to ship features that are likely to break and harm the user experience. Additionally, every feature we add imposes a UI/UX cost, which further harms adoption.
> 
> This can be explained to the user and backups can be made. These arguments would preclude Bitcoin itself if applied to things like seed phrases. Users somehow manage to deal with it in those cases.

A clear advantage of on-chain inscriptions is that a one-time backup of a seed phrase is sufficient to recover all your assets in the future, with an appropriately designed wallet. That is a huge advantage over any scheme that requires on-going backups to avoid losing assets.

We already have systems like RGB's NFT's that have made the trade-off of requiring on-going backups in exchange for scalability. Similarly, Lightning's problems with backups are a notorious *problem* that is widely recognized as such.

> > A good question really is what is _your_ incentive for proposing this feature? Are you actually part of the ordinals community? Or are you proposing it because you want that community to do something that benefits some other community?
> 
> How would being able to make cheaper, larger NFTs that have a different, easily explainable, persistence tradeoff be something that isn't a valid topic? raised_eyebrow
> 
> It's even been discussed in prior issues: #801, #887, #624

It's a valid topic in general. But BitTorrent (and nostr) are clearly not very robust ways to do so. Actually writing up a quick patch to do that looks more like an attempt to shame Ordinals into getting rid of on-chain inscriptions rather than a bonafide new feature.

tyjvazum:
> It's a valid topic in general. But BitTorrent (and nostr) are clearly not very robust ways to do so. Actually writing up a quick patch to do that looks more like an attempt to shame Ordinals into getting rid of on-chain inscriptions rather than a bonafide new feature.

I see zero evidence in this thread of any shaming of on-chain inscriptions. In fact, supporting BitTorrent fundamentally requires on-chain inscriptions to work. It could be viewed as a "quick patch", because it's a simple feature to add, but that doesn't make it a bad feature.

shesek:
> You're using the word "incentive" in a very narrow way here, only referring to selfish economic incentives. People have other incentives too, such as doing good for their community. [..] clear incentive to running BitTorrent seeds [..] That incentive is missing for ordinal data.

Why doesn't the Ordinal community have that same incentive too?

And 'selfish economic incentives' aren't a bad thing, if there are platforms that benefit from Ordinal NFTs and seed them then everyone benefits from it.

> A good question really is what is your incentive for proposing this feature? Are you actually part of the ordinals community? Or are you proposing it because you want that community to do something that benefits some other community?
> ... an attempt to shame Ordinals into getting rid of on-chain inscriptions rather than a bonafide new feature.

Is this a good question really?

Making this about me and my motives rather than about the merits of the proposal/implementation seems unnecessarily hostile and unproductive. I expected better from you.

But to answer your question anyway:

I'm not hiding the fact that I personally find NFTs to be unexciting and prefer block space to be conserved for payment activities. But I'm not trying to stop anyone from using bitcoin in whatever way they want, and in fact I haven't even voiced an opinion or wrote anything on Twitter about Oridnals at all prior to tweeting about this PR.

I sent this contribution in good faith because I believe it can benefit both the Ordinal and Bitcoin communities -- or at least some parts of them, or of people that that may join the Ordinal community in the future because they feel more comfortable with block-space efficiency. It's an opt-in feature that no one is forced to use. I really don't get why you're reacting to it with such hostility.

shesek:
> We already have systems like RGB's NFT's that have made the trade-off of requiring on-going backups in exchange for scalability.

The tradeoffs with RGB/Taro are quite different.

They require off-chain storage and client-side validation for the token transfers themselves, not just for the inscription data. Also, the backup data needed by the user to restore funds is dynamic and has to be backed-up after every transfer, unlike the torrent data which is static and can be backed up more easily by anyone just once.

shesek:
Rebased on to top of master.

Thinking about this some more, it may be better to include the hashes under a new tag type rather than as the content type. The content type tag can then be used for the 'real' mime type of the file, which may be useful to have accessible without having to download the torrent (for example it makes it easier for the explorer to know how to embed external resources for display, using `` or `<video>` etc).

It can be a single tag with current format this PR is using (32 bytes SHA-256 + 20 bytes SHA-128 infohash, perhaps with the infohash made optional?), or as two tags for each hash type.

Another option is [multihash](https://github.com/multiformats/multihash). However, its standard only appears to cover cryptographic hash function types and not application-specific hashes (so it has no representation for BitTorrent infohashes and is probably unlikely to accept a PR for it), and it might be better to KISS and not over complicate this. It's also interesting to note that the multihash `<hash identifier><hash length varint><hash>` serialization format is very similar to that of Bitcoin Script for a `PUSH <hash tag identifier> PUSH <hash>` (the varint encoding is identical even), so its conceptually similar and perhaps even somewhat redundant to use both.

Happy to implement this change if that approach seems preferable.

jokie88:
seems like the whole point of ordinals is to avoid linkrot, and this introduces the possibility of this. if a file stops being 'torrented', then the link points to nothing. seems to ruin the simplicity and availability assurance that ordinals currently provides

shesek:
> seems like the whole point of ordinals

Different people can value different things about Ordinals. One could be excited about the ordinal numbers theory and digital token ownership on Bitcoin, but not feel too emotionally attached to the idea of on-chain storage for attached files. We can see that the vast majority of Ethereum NFTs aren't stored on-chain and people still seem to value them, so this is not some wild idea.

Even if on-chain storage is generally preferred, there are still some cases where you still wouldn't use it:

- It might be a big multi-gigabyte file that cannot realistically be stored on-chain.
- The inscribed file may be meant to be shared privately and not made available to the whole world.
- It may become too expensive to inscribe on-chain, at least for some use-cases and some people.

> ruin the simplicity and availability assurance

It's an opt-in feature for these that prefer the tradeoffs if offers, how does it ruin anything?

petertodd:
> > We already have systems like RGB's NFT's that have made the trade-off of requiring on-going backups in exchange for scalability.
> 
> The tradeoffs with RGB/Taro are quite different.
> 
> They require off-chain storage and client-side validation for the token transfers themselves, not just for the inscription data. Also, the backup data needed by the user to restore funds is dynamic and has to be backed-up after every transfer, unlike the torrent data which is static and can be backed up more easily by anyone just once.

Token transfers mean that tokens are moving from one party to another. So yes, in the inscriptions model if inscription data is not on chain every token transfer also requires a backup of that data by the new owner. That is fundamentally no different than RGB/Taro.

With RGB/Taro the community can also elect to backup transaction data. Though unlike ordinals, the owners are not forced to make that data public.

There's no reason for Ordinal's to implement a bad copy of RGB/Taro's approach. They should do their own thing.

shesek:
> So yes, in the inscriptions model if inscription data is not on chain every token transfer also requires a backup of that data by the new owner.

It requires backup *of a new piece of data* after each transfer, and if that fails to happen for any reason then the user can no longer prove ownership and cannot transfer the asset.

In the inscription case it can be archived just once initially after the inscription, for example to a slow cheap large disk that you can bury in a time capsule somewhere and revive years later, no matter how many times the token changed hands in the meanwhile.

petertodd:
On February 24, 2023 3:02:52 PM GMT+01:00, Nadav Ivgi ***@***.***> wrote:
>> So yes, in the inscriptions model if inscription data is not on chain every token transfer also requires a backup of that data by the new owner.
>
>It requires backup *of a new piece of data* after each transfer, and if that fails to happen for any reason then the user can no longer prove ownership and cannot transfer the asset.
>
>In the inscription case it can be archived just once initially after the inscription, for example to a slow cheap large disk that you can bury in a time capsule somewhere and revive years later, no matter how many times the token changed hands in the meanwhile.

Again you are missing my point... The new owner has to make a backup of their new asset. Pretty much exactly the same model as RGB.

petertodd:
On February 24, 2023 11:14:52 AM GMT+01:00, Nadav Ivgi ***@***.***> wrote:
>> seems like the whole point of ordinals
>
>Different people can value different things about Ordinals. One could be excited about the ordinal numbers theory and digital token ownership on Bitcoin, but not feel too emotionally attached to the idea of on-chain storage for attached files. We can see that the vast majority of Ethereum NFTs aren't stored on-chain and people still seem to value them, so this is not some wild idea.
>
>Even if on-chain storage is generally preferred, there are still some cases where you still wouldn't use it:
>
>- It might be a big multi-gigabyte file that cannot realistically be stored on-chain.
>- The inscribed file may be meant to be shared privately and not made available to the whole world.
>- It may become too expensive to inscribe on-chain, at least for some use-cases and some people.

If it becomes too expensive to inscribe on chain old inscriptions might become even more valuable than they otherwise would. Merging this pull-req requires hard thought about those economics.

>> ruin the simplicity and availability assurance
>
>It's an opt-in feature for these that prefer the tradeoffs if offers, how does it ruin anything?


Ordinals are basically an art project. Art is usually less than 100% rational and efficient.

1440000bytes (changes requested):
NACK

99% of users trying inscriptions are interested in it because they are on-chain (bitcoin) and simple. This option would add confusion, complexity and FUD. If someone wants to do it off-chain they already have other solutions.

AngusP:
NACK-ish

You can already inscribe a BitTorrent infohash if you want, and there will be people that understand that doing so isn't really the same as putting the content iteself on-chain as it's less durable, adds indirection, and the meaning of 'owning' it is weakened (you own a *pointer* to the content, not the content itself) -- but also I don't think it should be encouraged (to me adding native support to `ord` would count as encouragement). If lots of people used this, then there would likely follow pressure for ordinal explorers/viewers to alos go get the torrent files and treat them on-par with actually inscribed content. That would be bad IMO.

Appealing to authority, if widely adopted this would push Ordinal Inscriptions closer towards the janky shitcoin-style NFTs where almost all of them are just pointers to out-of-band content and not the actual asset, which is something Casey has explicitly given as motivation for making ord as a better way.

Also at the moment, if two people inscribe exactly the same file, then there are two copies of it that can be separately owned. The one that came first (or the one inscribed on a rarer sat) would be the more valuable, and it's easy to see "who owns what". If two people inscribe the same infohash, and we treat torrents as 'legit' out-of-band/off-chain storage, now you've got two pointers to the exact same thing, and they're both making a claim on that same thing -- this adds some new murkiness about "who owns what"

tyjvazum:
> NACK-ish
> 
> You can already inscribe a BitTorrent infohash if you want, and there will be people that understand that doing so isn't really the same as putting the content iteself on-chain as it's less durable, adds indirection, and the meaning of 'owning' it is weakened (you own a _pointer_ to the content, not the content itself) -- but also I don't think it should be encouraged (to me adding native support to `ord` would count as encouragement). If lots of people used this, then there would likely follow pressure for ordinal explorers/viewers to alos go get the torrent files and treat them on-par with actually inscribed content. That would be bad IMO.

Correct, if it isn't added here, then it's easy for alternative clients to add support for it.

> Appealing to authority, if widely adopted this would push Ordinal Inscriptions closer towards the janky shitcoin-style NFTs where almost all of them are just pointers to out-of-band content and not the actual asset, which is something Casey has explicitly given as motivation for making ord as a better way.

Actually, Casey has explicitly suggested, as recently as this year, adding support for off-chain content in multiple issues: https://github.com/casey/ord/issues/801, https://github.com/casey/ord/issues/887, https://github.com/casey/ord/issues/624

Regarding "janky shitcoin-style NFTs", I think the main issue is they don't even serve as pointers to their content, since it's uncommon for any hash-addressed content to be used. That difference allows the content to be changed, without any solid record of what it was originally supposed to be. There are no potential problems like that with BitTorrent-based content.

> Also at the moment, if two people inscribe exactly the same file, then there are two copies of it that can be separately owned. The one that came first (or the one inscribed on a rarer sat) would be the more valuable, and it's easy to see "who owns what". If two people inscribe the same infohash, and we treat torrents as 'legit' out-of-band/off-chain storage, now you've got two pointers to the exact same thing, and they're both making a claim on that same thing -- this adds some new murkiness about "who owns what"

It's not necessarily true the first, or "rarer sat", would be more valuable in all cases, and in both situations, on-chain and off-chain, the ordering would be the same so there would be no difference from the perspective of ownership. As far as on-chain content goes, it's very inefficient to replicate content, so when identical, the content of the second inscription should ideally be a pointer to the content of the first inscription anyway. These are all, of course, simply implementation details, leaving users to think about the concepts however they choose to.

1440000bytes:
?Has any reviewer tried to review https://github.com/bitcoin/bips/pull/1408?

Not that it matters for me however for others in this project maybe.

1440000bytes:
If not, maybe not waste your time getting dumb things merged in this repo although it belongs to @casey

hsjoberg:
Concept ACK.

I don't understand the weird gatekeeping.
Why are we trying to dictate what MIME types are allowed and not?
Seems totally unnecessary.

1440000bytes:
> Concept ACK.
> 
> I don't understand the weird gatekeeping.
> Why are we trying to dictate what MIME types are allowed and not?
> Seems totally unnecessary.

Gatekeeping is done by owners or maintainers of repository. Example: Luke Dashjr in bips repo. 

All comments in this PR are by reviewers who can have opinions same as you.

AngusP:
> I don't understand the weird gatekeeping.
> Why are we trying to dictate what MIME types are allowed and not?
> Seems totally unnecessary.

To be clear, I'm not at all opposed to `ord` accepting any MIME type, including torrents. It should handle commonly used types, but should also happily take user-set MIME types (with reasonable caveat that the ord explorer probably won't be able to render them)

I'm NACK on treating inscribed torrents as a _special case_, as-if the inscription is of the content pointed to by the torrent, and not the actual torrent file itself — but not against inscribed torrent files or other pointers to out-of-band storage being inscribed (e.g. IPFS, plain old URLs, pointers to things on other blockchains, etc)

shesek:
@AngusP 

> You can already inscribe a BitTorrent infohash if you want

How? `ord` doesn't let you do that.

> you own a _pointer_ to the content, not the content itself

It could be argued that with on-chain inscriptions you also own a pointer - the ordinal sat associated with the inscription - and not the data itself, which is readily and freely available to everyone. But I guess that's more of a philosophical question really.

> ordinal explorers/viewers to also go get the torrent files and treat them on-par with actually inscribed content. That would be bad IMO.

Why would it be bad for explorers to make a copy of torrents available for users, similarly to how they make a copy of on-chain inscribed files? I think it would actually be pretty neat (and also not too technically complicated, I can perhaps even help implement this).

But IMO it could also be sufficient if the explorer simply rendered a magnet link (as implemented in this PR). Users can then use [instant.io](https://instant.io) to download/preview it right in their browser.

> if widely adopted this would push Ordinal Inscriptions closer towards the janky shitcoin-style NFTs where almost all of them are just pointers to out-of-band content and not the actual asset

Well, if it does get widely adopted by the Ordinal community, wouldn't it be evidence that this feature has merit and is wanted by users?

> If two people inscribe the same infohash, and we treat torrents as 'legit' out-of-band/off-chain storage, now you've got two pointers to the exact same thing, and they're both making a claim on that same thing -- this adds some new murkiness about "who owns what"

Couldn't you simply also treat the first/rarest-satoshi-associated torrent inscription as the 'legit' one, exactly like how you described you would solve it with on-chain inscriptions?

@tyjvazum 

> Correct, if it isn't added here, then it's easy for alternative clients to add support for it.

Using an alternative client to make the torrent inscription could be alright and not too cumbersome, but it would be great if the block explorer at least rendered it by showing the hash in hex and with a magnet link. Users having to go to (and trust) a separate ordinal block explorer just for that doesn't seem very practical.

1440000bytes:
> How? ord doesn't let you do that.

You can write anything and inscribe a text file

yanagawak:
A thriving block space market, being necessary to the security of a decentralized Bitcoin, the right of the people to inscribe data on-chain, shall not be infringed.

tyjvazum:
> Using an alternative client to make the torrent inscription could be alright and not too cumbersome, but it would be great if the block explorer at least rendered it by showing the hash in hex and with a magnet link. Users having to go to (and trust) a separate ordinal block explorer just for that doesn't seem very practical.

Yes, it'd be great if the block explorer supported it, so hopefully that happens. However, I don't think separate block explorers should be avoided, and I don't think any should be treated as canonical. Imagine if that had happened with Bitcoin, and how much of a centralizing effect it could have become.

Ideally, users wouldn't even use a mere centralized website to begin with, but in practice that will likely require substantial work on some type of light client mode, or at the very least a better version of a local explorer.

AngusP:
> How? ord doesn't let you do that.

Fair point, current release doesn't, but PR #1482 adds arbitrary MIME type support, or as @1440000bytes says you can inscribe it as a text file though that's not ideal

AngusP:
> Why would it be bad for explorers to make a copy of torrents available for users, similarly to how they make a copy of on-chain inscribed files? I think it would actually be pretty neat (and also not too technically complicated, I can perhaps even help implement this).

Because pholosophically I don't see an inscribed torrent as being the *same thing* as an actually inscribed file. It lets you 'get away with' cheaping out and inscribing a small pointer rather than 'comitting' and paying to put your content on-chain. 

If it's treated as a special case on-par with actually inscribed content we end up closer at the current state of NFTs on altcoins where most people won't understand the distinction, caveats and detail, and then be upset and confused when one day their 'inscription' goes away because no-one seeds the torrent anymore.

nammaki:
I hope this gets merged 🙏

casey:
Also, thank you for the effort and time that you put into this PR! I appreciate the thought and care that went into it. However, this change is not something that is suitable for a PR. A large change like this should be discussed before code is written, otherwise contributors will waste time writing code that does not have a change to be merged.

I eventually hope to consider some form of BitTorrent support, but:

- It must be BitTorrent V2, because V1 uses SHA1, which is insecure.
- User education and UX *MUST* receive significant thought and attention. If a user trades bitcoin for a inscription where the content that they care about is not on-chain, and does understand that the content does not have the same guarantees, then they we are opening up an avenue for obvious and preventable harm to users.

Eventually, when other, higher priority issues have been fixed, I hope that this receives the attention that it deserves, but there are too many high priority bugs, issues, and features that need to be addressed, and a large change like this requires much more bandwidth.

tyjvazum:
@shesek, thank you for your work and discussion on this. I used it as a basis to implement support in [arb](https://github.com/tyjvazum/arb). Any improvements you have in mind would be very welcome 👍

PR #1820, opened 2023-02-22
Push Docker image on git tag via Github Actions

tiero:
This commits adds a GitHub Action that automatically catches git tags, checkout the project at the tag, and builds a docker image to be pushed to Github Container Registry.

The docker image supports a multi-architecture Docker manifest, which means that will work on both arm64 and amd64 architectures, with just a docker pull.

Please @casey review


Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+USER root
+RUN mkdir builder
+WORKDIR /builder
+
+RUN set -ex \
+	&& apt update \
+	&& apt install -y libssl-dev
+
+
+RUN git clone https://github.com/tiero/ord.git
```
kvnn, line 14: Hi, did you mean to use your personal fork url here?

casey:
I want to first merge a working dockerfile before worrying about how to publish docker images, and before doing that, I need to understand why a Dockerfile would be useful. Please comment in #2032!

PR #1823, opened 2023-02-23
added eth address checksum

rot13maxi:


rot13maxi:
going to add more tests and server support and then re-open

PR #1825, opened 2023-02-23
Added checksum to eth address, added teleburn to inscriptions page

rot13maxi:
In this PR:
- eth teleburn addresses now have checksums in them. Had to take a dependency on rust-crypto, but didn't take any eth-specific dependencies
- added the ethereum teleburn address to the inscription detail page
- tests!

### DIFF
--- a/src/subcommand/teleburn.rs
+++ b/src/subcommand/teleburn.rs
@@ -1,5 +1,4 @@
-use super::*;
-use crate::teleburn_address::EthereumTeleburnAddress;
+use {super::*, crate::teleburn_address::EthereumTeleburnAddress};
 
 #[derive(Debug, Parser)]
 pub(crate) struct Teleburn {
--- a/src/subcommand/teleburn.rs
+++ b/src/subcommand/teleburn.rs
@@ -1,3 +1,5 @@
+use crypto::digest::Digest;
+use crypto::sha3::Sha3;
 use {super::*, crate::index::entry::Entry};
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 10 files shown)


cbspears:
This looks pretty fun!

src/subcommand/teleburn.rs
```diff
@@ -10,35 +11,10 @@ pub struct Output {
   ethereum: EthereumTeleburnAddress,
 }
 
-#[derive(Debug, PartialEq)]
```
cbspears, line 13: test comment plz ignore

Psifour: Review ACK

This reviews suggested changes are to "plz ignore" and that is what I do with code reviews anyways.

src/teleburn_address.rs
```diff
@@ -0,0 +1,93 @@
...
+#[derive(Debug, PartialEq, Serialize)]
+pub(crate) struct EthereumTeleburnAddress {
+  pub(crate) address: String,
+}
+
+impl From<InscriptionId> for EthereumTeleburnAddress {
+  fn from(inscription_id: InscriptionId) -> Self {
+    let digest = bitcoin::hashes::sha256::Hash::hash(&inscription_id.store());
+    Self {
+      address: create_address_with_checksum(&digest[0..20].to_hex()),
```
cryptoquick, line 16: Wouldn't it make more sense to just use md160 as the hash function here? It's the same one that hashes public keys in P2WPKH addresses, and it also produces 20 bytes.

cryptoquick (commented):
I really love this work, but just one question.

src/teleburn_address.rs
rot13maxi: There’s an implementation that @casey and rob did thats already been used to teleburn some NFTs, so i stuck with this to not invalidate that burn

cryptoquick: Gotcha, well, this works too!

raphjaph (approved):
I like it. Just gonna merge this into the `teleburn` branch so we can continue discussing there.

PR #1828, opened 2023-02-23
Skip indexing inscriptions when below first inscription

andrewtoth:
After #1759 we skip indexing transactions when below the first inscription height, and fetch missing inputs afterwards. However, we still try and index inscriptions below the first inscription block during `--index-sats`.

This PR skips transactions below the first inscription block for `--index-sats` as well. With this method combined with #1812  I was able to sync `--index-sats` in 23 hours 42 minutes.

### DIFF
--- a/src/index/updater.rs
+++ b/src/index/updater.rs
@@ -344,7 +344,9 @@ impl Updater {
 
     let mut outpoint_to_value = wtx.open_table(OUTPOINT_TO_VALUE)?;
 
-    if !self.index_sats {
+    let index_inscriptions = self.height >= index.first_inscription_height;
+
+    if index_inscriptions {
       // Send all missing input outpoints to be fetched right away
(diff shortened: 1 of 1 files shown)


jorge-j1m:
Probaly unrelated, but I've noticed that block committing takes a big chunk of the time needed to build an index, I don't know enough Rust to fiddle with that myself, but I feel like being able to manipulate the "checkpoints" (i.e. specifying how many blocks to allow between index commits) from the settings would be a good improvement, for me and most of use cases I know this is just a process in a server, fairly stable, so it's not super crazy to increase that threshold. I also don't know how easy/complicated a change it is, but being able to allocate more memory to the `ord` process should help as well. I can't help with the development, but if you make a PR I'm happy to test and report results @andrewtoth!

raphjaph (approved):

PR #1847, opened 2023-02-27
Fix the ord wallet descriptors

### LINKED ISSUE #1801: Use change addresses for change
When ord creates a wallet it uses internal = true for the receiving addresses and internal = false for change addresses. This is exactly backwards.

To see this, look at the change address in the commit tx. It is a receiving address not a change address. And look at the output of 'wallet receive'. It is a change address. 

To fix this delete the ! before 'change' about 6 lines from the end of wallet.rs.

gmart7t2:
Mark change addresses as change addresses and receiving addresses as receiving addresses rather than the other way around.

Closes #1801.


nammaki:
Merge this already 😓

raphjaph (approved):

PR #1854, opened 2023-02-28
Makes bitcoind a forking daemon

kvnn:
Unless I'm mistaken, we'd like bitcoind to run as a daemon, and doing so requires it to be a forking process under systemd . 

See:
- https://github.com/bitcoin/bitcoin/blob/master/contrib/init/bitcoind.service#L33
- https://manpages.ubuntu.com/manpages/impish/man5/systemd.service.5.html
- https://bitcoin.stackexchange.com/questions/74116/bitcoind-daemon-mode

### DIFF
--- a/deploy/bitcoin.conf
+++ b/deploy/bitcoin.conf
@@ -1,3 +1,4 @@
 datadir=/var/lib/bitcoind
 maxmempool=1024
 txindex=1
+daemon=1
--- a/deploy/bitcoind.service
+++ b/deploy/bitcoind.service
@@ -26,7 +26,7 @@ StateDirectory=bitcoind
 StateDirectoryMode=0710
 TimeoutStartSec=infinity
 TimeoutStopSec=600
-Type=simple
+Type=forking
 User=bitcoin
 
 [Install]


raphjaph:
This seems reasonable to me but I'm not a systemd expert. What do you think @casey?

casey (changes requested):
Thank you for the PR! This seems reasonable, but I don't understand the motivation. Are you running into a problem that this solves? For the ord instances that we run, the current configuration seems fine. Also, this is a change in how the ord service is run, so to confirm that it's correct, we would need to do a local test deploy. We've wanted to be able to spin up a local test instance, but IIRC, we don't have a way to do that.

kvnn:
> Thank you for the PR! This seems reasonable, but I don't understand the motivation. Are you running into a problem that this solves? For the ord instances that we run, the current configuration seems fine. Also, this is a change in how the ord service is run, so to confirm that it's correct, we would need to do a local test deploy. We've wanted to be able to spin up a local test instance, but IIRC, we don't have a way to do that.

> I don't understand the motivation
If I remember correctly I was running into issues with my own setup of bitcoind , ord and a python app when `forking` was not set correctly. This PR was made in hyper-vigilance more than practicality. I also do not have a local setup to test it, so, I'm going to close it since you haven't heard / seen any issues with the `bitcoind` daemon in your setup. Cheers

PR #1856, opened 2023-02-28
Fix bitcoin wiki link

gmart7t2:
The colored coins link was broken.


raphjaph (approved):
good catch!

raphjaph (approved):
good catch!

raphjaph (approved):
good catch!

Issue #783, opened 2022-11-15
Collections and Provenance


casey:
See this comment for an update: https://github.com/casey/ord/issues/783#issuecomment-1495169403

Provenance is important for art, and a large part of how it derives its value. We should have some notion of the provenance of an inscription:

- For an inscription, see who made it
- For a creator, see which inscriptions they've made
- Allow referencing an inscription within an artist's body of work by number or name (maybe just name, so they aren't confused with symbols) `/ARTIST_SYMBOL/NUMBER`

This will likely be accomplished as follows:

- If the inscription reveal transaction contains more than one input, the first ordinal of the first input is the "prime" ordinal
- If inscription has a prime ordinal, the first output of the inscription is ignored for the purposes of figuring out which ordinals are inscribed
- If an inscription has a prime ordinal, the inscription page should link to the prime ordinal's page
- If an ordinal has been used as a prime ordinal, the ordinal page should list all inscriptions it's made

It should be possible to make a series of related inscriptions:

- If an inscription is part of a collection, it should link to the collection page
- The collection page should link to all inscriptions in the collection
- The collection page should have some description about the collection
- It should be possible to nest collections. E.g. create a collection that contains multiple collections.
- Collections should have provenance. You should be able to see the scribe who created a collection, and see the collections issued by a scribe.
- It should be possible to make a collection closed-ended, e.g. guarantee that it isn't possible to issue more items.
- It should be possible link to an item in a collection by number or possibly name, e.g. - `/COLLECTION_SYMBOL/NUMBER` or `/COLLECTION_SYMBOL/NAME`.

Design sketch:

- Reuse prime ordinals from #783. A collection is just a prime ordinal, and all the inscriptions that it's made. 
- If you want to add a description to the collection, make an inscription on the prime ordinal used to issue the collection.
- If you want to close the collection, send the prime ordinal to an OP_RETURN. (I don't actually like this. It seems bad to force burning the prime ordinal of a collection. But, it's very simple.)

rot13maxi:
this sounds awesome

dannydeezy:
i'm a bit confused by this line
> If inscription has a prime ordinal, the first output of the inscription is ignored for the purposes of figuring out which ordinals are inscribed

let me see if i get this right:
Say I want to make a collection Astral Babes with good provenance. 
So I create an inscription w/ info about the collection like normal onto the first satoshi of Outpoint-A, which a value of 10,000 sats

[...]

PR #1869, opened 2023-03-04
Enable collections/provenance through parent inscriptions method

raphjaph:
Implementation work on #783, look [there](https://github.com/casey/ord/issues/783#issuecomment-1454314946) for more information.

### DIFF
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-pub(super) trait Entry: Sized {
+pub(crate) trait Entry: Sized {
   type Value;
 
   fn load(value: Self::Value) -> Self;
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -26,10 +26,12 @@ pub(crate) struct InscriptionEntry {
   pub(crate) fee: u64,
   pub(crate) height: u64,
   pub(crate) number: u64,
+  // pub(crate) parent: Option<InsciptionId>,
   pub(crate) sat: Option<Sat>,
   pub(crate) timestamp: u32,
 }
 
+// pub(crate) type InscriptionEntryValue = (u64, u64, u64, (u128, u128), u64, u32);
 pub(crate) type InscriptionEntryValue = (u64, u64, u64, u64, u32);
 
 impl Entry for InscriptionEntry {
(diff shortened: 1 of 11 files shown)


src/inscription.rs
```diff
@@ -106,6 +130,16 @@ impl Inscription {
...
 
+  pub(crate) fn get_parent_id(&self) -> Option<InscriptionId> {
+    if let Some(vec) = &self.parent {
+      Some(InscriptionId::load(
+        vec.clone().try_into().expect("expected a [u8; 36]"),
+      ))
+    } else {
+      None
+    }
+  }
```
ericatallah, line 141: The `expect` here seems like it can break the index too easily in the case of any malformed parent data. Maybe return `None` instead of system fail?

```suggestion
  pub(crate) fn get_parent_id(&self) -> Option<InscriptionId> {
    if let Some(vec) = &self.parent {
      if let Some(vec2) = vec.clone().try_into().ok() {
        Some(InscriptionId::load(vec2))
      } else {
        None
      }
    } else {
      None
    }
  }
```

raphjaph: Thanks, good catch!

Could you please create a test that demonstrates the issue and verifies it is solved? If you can, also think about test cases around different malformed inputs.

raphjaph: Thanks, good catch!

Could you please create a test that demonstrates the issue and verifies it is solved? If you can, also think about test cases around different malformed inputs.

veryordinally:
`ord wallet inscribe` is still broken with this PR:
```
error: Failed to send reveal transaction
because: JSON-RPC error: RPC error response: RpcError { code: -26, message: "non-mandatory-script-verify-flag (Invalid Schnorr signature)", data: None }
```

src/inscription.rs
```diff
@@ -15,24 +16,42 @@ const PROTOCOL_ID: &[u8] = b"ord";
 
 const BODY_TAG: &[u8] = &[];
 const CONTENT_TYPE_TAG: &[u8] = &[1];
+const PARENT_TAG: &[u8] = &[3];
 
 #[derive(Debug, PartialEq, Clone)]
 pub(crate) struct Inscription {
-  body: Option<Vec<u8>>,
+  parent: Option<InscriptionId>,
```
wagmiwiz, line 23: This is a more general point, but I was curious if there is a technical reason behind having only a single parent (vs. a list of 1 or more)?

I understand that in most cases people may want just one but there are also lots of usecases where you may want more than one. A basic example is a mashup (I take one inscription and another inscription and create a new mashup art). A more advanced example is "crafting", where I may take some set of inscriptions as parents, create a child, and burn the parents (all in one tx).

veryordinally:
After some deep-diving on decoding the partially signed and fully signed reveal TXs when inscribing with a parent, I found that `sign_raw_transaction_with_wallet` apparently removes our inscription data from the witness data. Need to dig deeper to find out why. 

@raphjaph @casey it would be great to catch this in tests, but I wasn't sure how to best fit this kind of test into the testing framework. Happy to discuss.

src/subcommand/wallet/inscribe.rs
```diff
@@ -278,7 +382,7 @@ impl Inscribe {
     );
 
     let witness = sighash_cache
-      .witness_mut(0)
+      .witness_mut(commit_input_offset)
       .expect("getting mutable witness reference should work");
     witness.push(signature.as_ref());
```
ericatallah, line 387: When inscribing with a `parent`, do we need to encode the `signature` with the proper SigHashType before adding to the witness?
```
let encoded_sig = SchnorrSig { sig: signature, hash_ty: SchnorrSighashType::AllPlusAnyoneCanPay };
witness.push(encoded_sig.serialize());
```

src/inscription.rs
```diff
@@ -231,7 +260,19 @@ impl<'a> InscriptionParser<'a> {
...
 
-      return Ok(Some(Inscription { body, content_type }));
+      return Ok(Some(Inscription {
+        body,
+        content_type,
+        parent: match parent {
+          None => None,
+          Some(bytes) => {
+            if bytes.len() != 36 {
+              return Err(InscriptionError::InvalidInscription)
```
ericatallah, line 270: Might just ignore and return `None` here instead of throwing Error.

src/inscription.rs
```diff
@@ -231,7 +260,19 @@ impl<'a> InscriptionParser<'a> {
...
+            }
+            Some(InscriptionId::load(bytes.as_slice().try_into().unwrap()))
+          }
+        },
```
clrke, line 274:
```suggestion
        parent: parent.and_then(|parent| {
          Some(InscriptionId::load(parent.as_slice().try_into().ok()?))
        }),
```

src/index/entry.rs
```diff
@@ -85,6 +89,74 @@ impl Entry for InscriptionId {
...
+        tail_array[11],
+        tail_array[12],
+        tail_array[13],
+        tail_array[14],
+        tail_array[15],
+        index_array[0],
+        index_array[1],
+        index_array[2],
+        index_array[3],
+      ];
```
felipelincoln, line 142: A suggestion to improve code readability:

```suggestion
      let array = vec![
        &head_array[0..=15],
        &tail_array[0..=15],
        &index_array[0..=3]
      ].concat().try_into().unwrap();
```

src/inscription.rs
veryordinally: Yep, this part unfinished, but you're right, we should gracefully ignore malformed inscriptions.

ericatallah:
I've been having trouble getting a child inscription to render properly in the explorer.  Looks like a few inscription indexes were hard coded to 0.  Not sure if there are more though.  Here is a minor fix, hope it helps:
https://github.com/raphjaph/ord/pull/3

veryordinally:
Making progress. For anyone watching, this code does NOT meet our usual quality criteria yet, by far - lots of tests missing and some places are just held together by string. Anyone tempted to try this, this code may work, but be aware that using this on live wallets may lead to loss of inscriptions or other unpredictable outcomes.

tests/lib.rs
```diff
@@ -43,6 +43,7 @@ struct Inscribe {
   inscription: String,
   reveal: Txid,
   fees: u64,
+  parent: Option<String>,
```
clrke, line 46:
```suggestion
  #[allow(dead_code)]
  parent: Option<String>,
```

clrke: This is just so that the IDE doesn't show warnings that `inscribe.parent` isn't being used.

PR #1871, opened 2023-03-05
README.md: less brittle install command

RandyMcMillan:
fixes issue when installing on macOS x86_64:
Malformed Archive: url


RandyMcMillan:
macos_x86

PR #1874, opened 2023-03-05
Better interrupt message

neunenak:
Print a message saying that Ord has responded to a Ctrl-C event and will shut down immeidately with 5 Ctrl-C's.


raphjaph (approved):
Looks good to me!

PR #1897, opened 2023-03-08
Mandatory fee rate for inscribe

gmart7t2:
Too many people are inscribing with the default 1 sat/vB fee rate.

This should help.


raphjaph:
Looks good!

Issue #943, opened 2022-12-11
Traits and rarity


casey:
Lots of generative collections have traits of different rarity. This is very interesting to collectors. We should consider supporting a way to add this to a collection.

One way to do this is to add a field for arbitrary textual metadata. This will get expensive fast, and will be limited to 520 bytes, but it would do the trick and be easy.

A better way to do this would be to allow multiple inscriptions on the same sat, and store traits and metadata in an inscription after the main inscription. We could use a tag or special title to indicate that it contains the traits. HTML inscriptions should be able to access this metadata.

coding-hunk:
Hi @casey ,
I think your approach to adding traits in inscriptions is a great way to encourage new communities and projects to join the ecosystem. I've been working in the direction you mentioned and came up with given solution & raised a PR [here](https://github.com/casey/ord/pull/1906/files).

Solution:
1. While running ord inscribe command, user can add **'--metadata'** flag (optional) and pass on traits in JSON form. The JSON will be added in the inscribing envelop as another segment and broadcasted on-chain. eg.
`ord wallet inscribe --metadata "data"`
An example of data passed:
```
{
  "traits": [
    {
      "trait_type": "hair",
      "value": "Short Black"
    },
    {
      "trait_type": "cap",
      "value": "Blue Cap"
    }
  ]
}
```

2. While showing the inscription details on explorer, we will show **traits** section if metadata is NOT NULL and has whitelisted key "**traits**" in it. Showing ss demonstrating the content passed in that key.
[image_link](https://ibb.co/DQGMgsY)


Once collections and provenance gets established, we can easily compute rarity % for traits in items & collections. 
This can encourage more community-based NFT projects to join Ordinals and create new collections. Happy to incorporate changes if community has any thoughts.

Note: Validating JSON in metadata input and parsing key-value pair on explorer UI is pending. Will add it soon :)

p0stcapone:
I'd like to propose leveraging BIP-322 for metadata (and many other things).

Off chain messaging which is provably associated to the creator and/or some later collector would allow for definitions of collections, metadata, artist information, and even evolving narratives as collectors may contribute along the way.  Participants would have similar freedoms to storage options as they do with PSBT (public, private, centralized, distributed, etc).

It appears that for most of the extant NFT markets, metadata is critical for indexing/sorting features, and thus marketplaces, explorers, etc would all share some mutual incentives to ensure availability of the data (and the data could be trivially re-produced if necessary).

[...]

PR #1899, opened 2023-03-08
metadata string added for inscribe file

MBenz12:
1. In casey ord github repo (https://github.com/casey/ord). There is a inscribe command situated at this path (https://github.com/casey/ord/blob/master/src/subcommand/wallet/inscribe.rs). We need add to start accepting a new parameter in this file.
Purpose of this param - this param will let anyone pass an additional string param and will attach this param to the meta information of the file. 

Example, 
ord wallet inscribe FILE --metadata "metada_string"

2. After the above, your inscription struct (https://github.com/casey/ord/blob/master/src/inscription.rs) will have the meta information that we attached to this. This struct powers this template (https://github.com/casey/ord/blob/master/src/templates/inscription.rs). So what we need to do is, start displaying meta information in this view. 
Currently this view looks like this (https://ordinals.com/inscription/431fea141cfc2e6f8dda9c209fbe9ce411e9e158d48661e86cbd0cc5e0bc33f5i0)
It should show meta information somwhere that we attached

### DIFF
--- a/src/inscription.rs
+++ b/src/inscription.rs
@@ -20,19 +20,20 @@ const CONTENT_TYPE_TAG: &[u8] = &[1];
 pub(crate) struct Inscription {
   body: Option<Vec<u8>>,
   content_type: Option<Vec<u8>>,
+  pub metadata: Option<&'static str>,
 }
 
 impl Inscription {
   #[cfg(test)]
   pub(crate) fn new(content_type: Option<Vec<u8>>, body: Option<Vec<u8>>) -> Self {
-    Self { content_type, body }
+    Self { content_type, body, metadata: None }
   }
 
   pub(crate) fn from_transaction(tx: &Transaction) -> Option<Inscription> {
     InscriptionParser::parse(&tx.input.get(0)?.witness).ok()
   }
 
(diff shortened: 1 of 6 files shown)


tyjvazum:
With the code you've provided, nothing would actually be done with the metadata. It wouldn't be written in the inscription transaction and it wouldn't be read when an inscription is parsed from the blockchain. It'd need a tag and correct builder/parser logic for that. Take a look at https://github.com/casey/ord/pull/1713 for an example of how that can be implemented.

MBenz12:
thanks for your advice @tyjvazum 
I've added that features, 
could you please review my code again?

coding-hunk:
Hey mate @MBenz12 ,
Glad to see you raised a PR with same technical solution which I suggested you in discord. Would have been better if you could have told me that you also want to contribute here. 
Reference - [Issue](https://github.com/casey/ord/issues/943)
We have already raised PR to Casey's open issue for traits. Would request to help us in getting that PR merged or you can further contribute in the PR to improve it further. Let's work together to make Ordinals more strong ecosystem :)

MBenz12:
sorry @coding-hunk 
I didn't check you already created pr for this issue,
I will close my pr.
I hope to work together to make Oridinals more strong.
thanks for your contribution

PR #1904, opened 2023-03-09
Add "ord wallet cardinals" command to list the cardinal outputs.

gmart7t2:
There's no easy way to find a list of cardinal outputs. You have to list all the outputs and remove the inscriptions. 

Hence: `ord wallet cardinals`.


BTCAlchemist:
This would really be helpful. I'm overwhelmed with the list of outputs and have difficulty discerning how many cardinal UTXOs and the balances that remain. @raphjaph Would you be the right person to tag for review?

PR #1906, opened 2023-03-09
add traits for inscriptions

coding-hunk:


tyjvazum:
This is very similar to https://github.com/casey/ord/pull/1899...

raphjaph:
Could you elaborate what specific use case you have in mind for this?

casey:
This is a large feature which has many implications that I haven't yet considered, so closing for now.

PR #1922, opened 2023-03-12
Use the --fee-rate when sending an amount

### LINKED ISSUE #1842: sending cardinals requires but ignores --fee-rate
I run:

    ord wallet send --fee-rate <rate> <destination> 99999sats

It sends the 999999 sats but doesn't use the specified fee rate. Instead it uses the default fee rate that I set with:

    bitcoin-cli settxfee

The send is done using:

    client.send_to_address()

which only takes 8 arguments. Bitcoin's `sendtoaddress` RPC takes a `fee_rate` as its 10th argument but the rust client doesn't allow access to it. 

Workaround: use `bitcoin-cli settxfee <btc-per-kvb>`. So to get the effect of `--fee-rate 7` use `bitcoin-cli settxfee 0.00007000` (with the necessary `-rpcwallet=<wallet>`).

gmart7t2:
"ord wallet send" requires the --fee-rate flag but ignores it when sending cardinals. This PR fixes that and also adds 2 new Outgoing types: 'all', and 'max':

  * 'wallet send all' sends all available cardinals

  * 'wallet send max' maximizes the output by sending all cardinals that are worth more than they cost in fees to send

Users want to be able to consolidate all their cardinals into a single utxo. These new modes make that easier.

Closes #1842 but there is still the issue that "wallet send" leaves the user's inscriptions locked, and fails to run a 2nd time because it fails when it tries to lock the already-locked inscriptions.


BTCAlchemist:
I think the workaround until this is merged is to set the default fee-rate in Bitcoin Core and to manually unlock UTXOs between 'ord wallet send' commands. A fix would be helpful since users may not know the workaround and to save time for those who do. @raphjaph Do you have a moment to review this?

src/outgoing.rs
```diff
@@ -3,6 +3,8 @@ use super::*;
 #[derive(Debug, PartialEq)]
 pub(crate) enum Outgoing {
   Amount(Amount),
+  All, // send all cardinals
+  Max, // only send cardinals that can pay for their own fees to maximize the output amount
```
raphjaph, line 7: I don't see a difference between how these two cases are handled in the logic.

Just call it AllCardinal?

raphjaph (changes requested):
Definitely still needs a test as well.

PR #1923, opened 2023-03-12
Fix errors found by clippy since the rust 1.68.0 update.

gmart7t2:


casey (approved):
This is a great PR, obviously correct and would be an ez merge. However it looks like we already made these changes in another PR, so closing. Sorry for not looking at this sooner!

PR #1926, opened 2023-03-13
features/compression

staccDOTsol:
if we can store arbitrary bytes on btc chain is there no technical reason to not build a machine on an executable layer?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,6 +612,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "brotlic"
+version = "0.8.1"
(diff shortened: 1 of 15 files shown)


tyjvazum:
> if we can store arbitrary bytes on btc chain is there no technical reason to not build a machine on an executable layer?

This should be a separate issue since it's not related to compression, but to answer your question, there's no technical reason a state machine couldn't be built on top of inscriptions, other than the sat-tracking functionality likely being undesirable for such a use case.

casey:
This PR would take a lot of work and review to get into shape. Since we're bandwidth constrained, I'm going to close it.

PR #1927, opened 2023-03-13
Allow splitting merged inscriptions

gmart7t2:
I see a lot of users with multiple inscriptions merged into a single utxo.

Currently ord refuses to do anything with such utxos.

This pull request relaxes the condition on sending inscriptions and satpoints:

It is OK to send an inscription or satpoint from an output with multiple inscriptions in it as long as the satpoint being sent has no inscriptions after it, and also none within the dust limit before it. 

As long as those conditions are met we can safely send the inscription or satpoint and the user can separate their merged inscriptions one by one, starting from the last in the output. 

The error message when sending is not allowed has been modified to mention the last inscription in the output rather than the first to help guide the user to the one they need to send first.


gmart7t2:
The lint failures go away after #1923.

raphjaph (approved):

PR #1928, opened 2023-03-13
Add "amount" field to "wallet inscriptions" output.

gmart7t2:
Allow people to check the postage on each of their inscriptions.


rot13maxi:
this is always going to be 10000 until we have an option to change the output postage amount, or am I missing something?

raphjaph:
When sending around inscriptions the postage will decrease over time. It will only be replenished if it would go below the dust limit. So this change makes sense.

raphjaph (changes requested):
Reviewing with Casey. This is missing a few things that he will address in his review.

src/subcommand/wallet/inscriptions.rs
```diff
@@ -5,6 +5,7 @@ pub struct Output {
   pub inscription: InscriptionId,
   pub location: SatPoint,
   pub explorer: String,
+  pub amount: u64,
```
casey, line 8: I think `postage` is clearer than `amount` here:

```suggestion
  pub postage: u64,
```

casey, line 8: Struct fields should be sorted.

src/subcommand/wallet/inscriptions.rs
```diff
@@ -29,6 +30,7 @@ pub(crate) fn run(options: Options) -> Result {
         location,
         inscription,
         explorer: format!("{explorer}{inscription}"),
+        amount: unspent_outputs.get(&location.outpoint).unwrap().to_sat(),
```
casey, line 33: Refactor to use `if let …` and avoid the unwrap, like so:

```
    if let Some(output) = unspent_outputs.get(&location.outpoint) {
      output.push(Output {
        location,
        inscription,
        explorer: format!("{explorer}{inscription}"),
        amount: output.to_sat(),
```

casey (changes requested):
Nice! This is a good addition.

This requires a test! We should have an integration test that the returned `postage` field is correct. Add one in `tests/wallet/inscriptions.rs`.

raphjaph (approved):

PR #1938, opened 2023-03-16
Bug fix: Only return blocks once indexed

apezord:
**Background**

There are two block-related endpoints:

```
let router = Router::new()
   .route("/", get(Self::home))
   .route("/block-count", get(Self::block_count))
   .route("/block/:query", get(Self::block))
   .route("/bounties", get(Self::bounties))
```

`/block-count` returns how many blocks have been indexed by ord

`/block` returns blocks the node knows about regardless of whether ord indexed them

This difference around indexing could possibly lead to users accidentally sending away their ordinals if their wallet uses these APIs to query user artifacts and the wallet believes a block was indexed when it really wasn't.

I discovered this issue after a reorg which stops ord indexing.

**Change**

This PR queries the database of indexed blocks to determine if a block is truly indexed before returning from /block. Blocks not indexed by ord are effectively hidden.

**Potential issues**

When /block is called with a block hash, the database currently doesn't have an efficient index to check whether that hash is indexed. The code walks the block hashes known from newest to oldest to check. We could add an index but I wonder whether that is worth doing because at most it will read a few 10s of MBs and only for early blocks.

**Testing**

Manual

1. Queried /block with both height and hash during a reorg when node and ord are out of sync. Checked that blocks were filtered.
2. Queried again when not in a reorg and blocks appear normally.


apezord:
Raising priority on this. Users might lose their artifacts. Since BTC is experiencing a reorg and ordinals.com is reindexing, but if any wallet is using /block to determine if ord is indexed, they may think a UTXO doesn't have artifacts when it does.

apezord:
On second look, I believe this affects /tx endpoint as well. This could be bad.

DanielHook15:
If I use ord in the "normal" way described in the handbook (i.e. no sparrow wallet or similar), is it possible that I lose artifacts or does this problem only apply to other situations?
What can I do to ensure that I do not lose artifacts until the problem is solved by a new version of ord?

raphjaph:
The /block endpoint only return a block and no inscriptions so I don't see how this affects acccidentally sending those.

Closing this since it is kind of old. Please reopen if still relevant.

Issue #1609, opened 2023-02-08
Package ord for as many package managers as possible


casey:
Ideally, we could get `ord` into as many package repositories as possible. Homebrew should be a high priority. I think we already meet their notability and popularity guidelines. We cannot maintain these packages, so interested community members, ideally people who are already familiar with Rust and packaging Rust binaries for the repo they're working with.

You don't need our permission! Feel free to package `ord` and open an issue to add the install steps to the readme. Consult the readme of [casey/just](https://github.com/casey/just) for ideas of how to list packages in the readme. I like the two tables there.

Here's my priority list so far:

1. Nix: https://nixos.org/ - I think we can create a nix file which allows a user to start bitcoind, and ord, in one command, and it will work on many platforms.
2. Codespaces: Getting ord working in codespaces, so one-lick to get started.
4. mynode86 - node-in-a-box solution that supports x86
5. regtest replit
6. raspi blitz - (most?) open source node-in-box solution
7. docker - commonly used by node-in-boxes like start9, so having an example docker container for them to work off of would be awesome
8. homebrew - very common, macs are easy, super common dev platform

earthgohan:
Could you kindly provide a cdylib version of this library, as it would greatly assist in the building of wasm and the creation of web wallets.

casey:
@earthgohan Off topic for this issue! I noticed you opened an issue, which is great, but try not to comment on unrelated issues: #1621

casey:
@btcviking expressed an interest in packaging this for homebrew. I know him afk, and it would be great if someone we knew and trusted were managing such a popular package.

casey:
I think this is best left up to individual contributors, and not done by the project.

PR #1942, opened 2023-03-17
add deb package

ericatallah:
Add deb package, along with instructions in README to build one.

https://github.com/casey/ord/issues/1609

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,10 @@ repository = "https://github.com/casey/ord"
 autobins = false
 rust-version = "1.67"
 
+[package.metadata.deb]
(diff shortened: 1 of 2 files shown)


casey:
Thanks for the PR! A few issues:

- The current maintainer is Raph, but this may change in the future. Can we use something generic, like "The ord maintainers"?
- We'd rather list an email address or receive emails. Can we omit the email addresses?
- For the purpose of copyright, there have been many contributions aside from mine. Would it be better to put something like "The ord contributors"?

casey:
Added in #2725 with generic info for maintainer and copyright.

Issue #1894, opened 2023-03-08
Use Improved Memory Allocators When Available


bonedaddy:
# Overview

By default rust uses the system's default memory allocator which tends to be `malloc`. `malloc` is sufficient for light workloads, however under heavy workloads, and lots of memory allocations like those done by `ord` it can be extremely useful to use a different allocator.

# Solution

Add optional support for usage of `jemalloc`

# Notes

This paper does a thorough analysis of different allocators, and recommends the usage of `jemalloc` over `malloc` for things like RDMS servers.

https://adms-conf.org/2019-camera-ready/durner_adms19.pdf

veryordinally:
Sounds interesting - would be great if someone could try this out and add some comparative benchmarks/

nammaki:
Benchmark provided in https://github.com/casey/ord/issues/1893

Issue #1893, opened 2023-03-08
Improve Heap Allocation In Index::blocks


bonedaddy:
# Overview

Heap allocation is inefficient in `index::blocks`, as it initializes a vector with an empty capacity and pushes elements into the vec. This will result in numerous allocations being performed, when the function can be optimized to perform only one allocation:

# Solution

Initialize the vector with a fixed capacity after retrieving `block_count`. If `block_count` is `100` this would avoid 6 total allocations in favour of one allocation (note: rust reallocs vector size by 2x). With the current bitcoin blockchain size, this would reduce hundreds of allocations into one allocation leading to drastically improved performance.

```rust
    let rtx = self.begin_read()?;

    let block_count = rtx.block_count()?;

    let height_to_block_hash = rtx.0.open_table(HEIGHT_TO_BLOCK_HASH)?;

    let mut blocks = Vec::with_capacity(block_count);

    for next in height_to_block_hash.range(0..block_count)?.rev().take(take) {
      blocks.push((next.0.value(), Entry::load(*next.1.value())));
    }
```

veryordinally:
Makes sense. Have you tried this and benchmarked?

bonedaddy:
This is from using both improvements introduced in https://github.com/casey/ord/pull/1959 and https://github.com/casey/ord/pull/1960 but pretty fast speed boost
![2023-03-21_18-19](https://user-images.githubusercontent.com/17089485/226777383-87e7fac8-1f44-467a-8664-4a9cae15621c.png)

veryordinally:
Sounds great - just to make sure I understand your benchmark data: this is timing for an index run with your optimizations? What was the time without the optimizations? Would be great to get a rough idea of speedup you see. Absolute times are hard to compare across setups.

cryptoni9n:
This appears to have been merged in #2409 and should therefore be closed.

PR #1959, opened 2023-03-22
Enable Usage of JemAllocator

bonedaddy:
Fixes the issue highlighted in https://github.com/casey/ord/issues/1894, however likely needs to be adjusted to support multiple operating systems as at present it will only support linux + mac osx

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2204,6 +2204,7 @@ dependencies = [
  "sys-info",
  "tempfile",
  "test-bitcoincore-rpc",
+ "tikv-jemallocator",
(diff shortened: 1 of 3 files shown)


schlunsen:
Nice! Could be awesome with some benchmark data to show the improved performance

nammaki:
> Nice! Could be awesome with some benchmark data to show the improved performance

Benchmark already provided https://github.com/casey/ord/issues/1893

casey:
Optimization PRs require a benchmark before they can be reviewed.

PR #1966, opened 2023-03-23
Update miniscript dependency to 9.0.1

soenkehahn:
Version 9.0.0 was yanked, see here:
- https://github.com/rust-bitcoin/rust-miniscript/blob/9.x-branch/CHANGELOG.md#901---march-8-2023
- https://crates.io/crates/miniscript/versions

I noticed this when running:
`cargo install --offline --frozen --path .`

Which gives a warning about using yanked versions from crates.io.


casey:
Should we also change the Cargo.toml file? Since the lock file is sometimes ignored, maybe we should. Not sure what best practices are here.

soenkehahn:
> Should we also change the Cargo.toml file? Since the lock file is sometimes ignored, maybe we should. Not sure what best practices are here.

I *think* `cargo` would already prefer `9.0.1` over `9.0.0` without changing the `Cargo.toml` file, since `9.0.0` is yanked. I also think bumping the version in `Cargo.toml` doesn't hurt. If anything it prevents people from using a yanked version. So I changed it.

(Also not 100% sure about how this works and what best practices are.)

casey:
> > Should we also change the Cargo.toml file? Since the lock file is sometimes ignored, maybe we should. Not sure what best practices are here.
> 
> I _think_ `cargo` would already prefer `9.0.1` over `9.0.0` without changing the `Cargo.toml` file, since `9.0.0` is yanked. I also think bumping the version in `Cargo.toml` doesn't hurt. If anything it prevents people from using a yanked version. So I changed it.
> 
> (Also not 100% sure about how this works and what best practices are.)

Seems reasonable to me! Merging.

PR #1970, opened 2023-03-24
Toggle between latest inscriptions and blocks

raphjaph:
This is a proposal to change the layout of the landing page. Specifically to hide the list of recent blocks, since they usually aren't that interesting.

Screenshot 2023-04-05 at 16 55 08
Screenshot 2023-04-05 at 16 55 19

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -1626,7 +1626,8 @@ mod tests {
     test_server.assert_response_regex(
     "/",
     StatusCode::OK,
-    ".*<h2>Latest Blocks</h2>
+    ".*
+  <h2>Latest Blocks</h2>
   <ol start=1 reversed class=blocks>
     <li><a href=/block/[[:xdigit:]]{64}>[[:xdigit:]]{64}</a></li>
     <li><a href=/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f>000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f</a></li>
@@ -1650,11 +1651,10 @@ mod tests {
     test_server.mine_blocks(101);
 
     test_server.assert_response_regex(
-      "/",
-      StatusCode::OK,
-      // ".*<ol start=101 reversed class=blocks>\n(    <li><a href=/block/[[:xdigit:]]{64}>[[:xdigit:]]{64}</a></li>\n){100}</ol>.*"
-      ".*<ol start=101 reversed class=blocks>.*</ol>.*",
-    );
+    "/",
+    StatusCode::OK,
+    ".*(  <ol start=101 reversed class=blocks>)\n(    <li><a href=/block/[[:xdigit:]]{64}>[[:xdigit:]]{64}</a></li>\n){100}(  </ol>).*"
+  );
   }
 
   #[test]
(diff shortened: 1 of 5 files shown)


soenkehahn:
Hey @raphjaph!

Could you provide a bit more info on what this feature is/does? Maybe even add some screenshots? (I think a good place would be the PR description.) That'd make it easier for more people to review the PR.

Also, can you make the tests pass?

And should we add tests for this? Generally, testing UIs like this is hard and sometimes not thaaat useful, so I'm not convinced either way. I think once the existing tests pass it'd be easier to see whether we want to add more tests. (Or are those flaky tests?)

raphjaph:
This PR is from a session with Caroline (web designer) and @casey. 

I don't really like the UI tests because they are a pain to get right (indentation always wrong and difficult to see what's wrong) and aren't very useful. Casey just always had them so I add them as well.

src/subcommand/server.rs
```diff
@@ -1629,10 +1628,11 @@ mod tests {
...
 
     test_server.assert_response_regex(
-    "/",
-    StatusCode::OK,
-    ".*<ol start=101 reversed class=blocks>\n(  <li><a href=/block/[[:xdigit:]]{64}>[[:xdigit:]]{64}</a></li>\n){100}</ol>.*"
-  );
+      "/",
+      StatusCode::OK,
+      // ".*<ol start=101 reversed class=blocks>\n(    <li><a href=/block/[[:xdigit:]]{64}>[[:xdigit:]]{64}</a></li>\n){100}</ol>.*"
+      ".*<ol start=101 reversed class=blocks>.*</ol>.*",
```
soenkehahn, line 1634: Hmm, yeah, these tests seem a bit frustrating to maintain. Still, this is probably not a change we want to make:

- It looks like the test is supposed to test that there's a limit to the number of shown blocks. The test doesn't test that anymore. If we don't want to test that anymore we should remove the test. If we do, we need to fix it.
- We probably don't want old regexes in comments like this.

soenkehahn:
@raphjaph: When we talked about this PR last we thought that we should revamp the tests to be able to test the template struct values directly. This was a bigger refactoring. When I talked to @casey, we thought that

1. We should probably not implement the more direct template struct testing system, but instead finish up https://github.com/casey/ord/pull/1662 at some point. And then -- instead of testing on the template struct values -- we could write very similar tests on the returned JSON. So for example the [home_block_limit](https://github.com/casey/ord/blob/adc4d7405824e8703129f5da60e7173d8636e284/src/subcommand/server.rs#L1624-L1636) won't have to do use regexes anymore, but -- since it's still testing the server through http network calls -- it also wouldn't have to deal with calling the server internals in an `async` way.
2. So -- to not block this PR -- I think we should just try to fix the regex things for now. I'll try to do come up with a commit.

soenkehahn:
@raphjaph: I came up with https://github.com/casey/ord/pull/1970/commits/b3b7b5847792a1aa02d49eb4092ffda1224b60ab. It fixes the `home_block_limit` test, and also tweaks some of the regexes, mainly to make the diffs in this PR read nicer. From my side this PR is ready to be merged with that change. Let me know what you think!

(I pushed to this PR branch, hope that's good with you!)

raphjaph:
I'm going to close this because not a priority right now

PR #1971, opened 2023-03-25
Subtract output values from offset when inscription spent as fee

### LINKED ISSUE #1841: Internal Server Error for ordinal inscriptions that were spent as miner fees
Some of the URLs on `/inscription` endpoint are throwing `Internal server error `, even though the inscriptions are visible on the `/tx` endpoint

For example:
- https://ordinals.com/inscription/ebfc364d43e6f921bc9aff3831340e1b06f8c813cb6fbb8a6a54aa26a8f80497i0
- https://ordinals.com/tx/ebfc364d43e6f921bc9aff3831340e1b06f8c813cb6fbb8a6a54aa26a8f80497

or

-  https://ordinals.com/inscription/b6bad18d2b08574f14d84342f722356ea73c19969bd241259443bbd43369428bi0
- https://ordinals.com/tx/b6bad18d2b08574f14d84342f722356ea73c19969bd241259443bbd43369428b

gmart7t2:
Fixes #1841.

Indexes will need rebuilding.


veryordinally:
@raphjaph this looks good to me - should prioritize for review.

unisky:
I have found this bug too, it made a big confusion to my current data. hope prioritize for review.

unisky:
hello guys,  when to merge this commit and fix history data? I think this bug is very fatal

src/index/updater/inscription_updater.rs
```diff
@@ -178,7 +178,7 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
       Ok(self.reward - output_value)
     } else {
       self.flotsam.extend(inscriptions.map(|flotsam| Flotsam {
-        offset: self.reward + flotsam.offset,
+        offset: self.reward + flotsam.offset - output_value,
```
casey, line 181: This is a bug fix, so it needs a test to ensure that we don't change it and regress. The test should fail without this change, and pass with this change. Please see other tests for an idea for how to do this.

casey (changes requested):
I think this change is correct. In addition to a test, can you please describe what the bug is, why this fixes it, and the observable change in behavior? There was a lot of discussion in #1841, but I think it also discusses a different issue, so it's a bit confused. If there's already a different issue describing this, you can also link that issue.

gmart7t2:
> I think this change is correct. In addition to a test, can you please describe what the bug is, why this fixes it, and the observable change in behavior? There was a lot of discussion in #1841, but I think it also discusses a different issue, so it's a bit confused. If there's already a different issue describing this, you can also link that issue.

The bug is #1841. 

https://github.com/casey/ord/issues/1841#issuecomment-1483950070 explains the bug and why this fixes it.

The observable change in behavior is that ord tracks inscriptions correctly when they are spent as change rather than tracking them incorrectly.

I don't understand the confusion here. Sorry if I'm being slow.

src/index/updater/inscription_updater.rs
gmart7t2: I modified the existing test for inscriptions spent as fees. Bug #1841 only happens when there are no outputs in the tx that spends the inscription as fee, so adding any input triggers the bug.

raphjaph:
I think the bug is very well explained! I just added the changes you made to the test in a new test so it is very clear what the bug was. This is ready for you to have a look @casey!

raphjaph (approved):

casey (approved):
Looks good! For bug fixes, existing tests should not be changed if they aren't broken, new tests should be added, which gives confidence that old and correct behavior, covered by the old tests, isn't being changed.

PR #1987, opened 2023-03-30
Make output created after strip_value cover tx fee

harutyunaraci:


gmart7t2:
The existing code looks correct to me. This PR effectively adds two fees together and ends up with an assertion error:

    thread 'main' panicked at 'assertion failed: `(left == right)`
      left: `Amount(0.00000265 BTC)`,
     right: `Amount(0.00000154 BTC)`: invariant: fee estimation is correct'

PR #1991, opened 2023-03-31
Don't try to create the default data directory if --index overrides it

gmart7t2:
Create the specified directory instead.


casey:
Can you describe what problem this PR is solving? Is the issue `ord` is creating the data dir when it isn't needed? Or is the issue that `ord` isn't creating the parent directory of the index when `--index` is passed?

raphjaph (approved):

PR #1996, opened 2023-04-02
Add `inscriptions` subcommand to list all inscriptions in order of in…

gmart7t2:
…scribed sat number.


soenkehahn:
Thanks for the PR!

I'm not the best person to review this PR, since I'm new here. In particular I'm not sure if this will be merged. `ord` tries to be somewhat conservative with adding new subcommands, since every user facing change / addition needs to be maintained in a backwards compatible way (within reason). So I can't promise anything, but here's my 2 cents on the current PR:

I think it'd be much easier for more people to review this PR if it were fleshed out a bit more:

- Maybe add some concrete use-case for when this would be useful in the PR description.
- Maybe add some (possibly truncated) example output in the PR description.
- Add some tests for the new command? Both as documentation and for preventing regressions in the future.
- Make all github checks pass.

raphjaph:
We now have `ord index export` command that exports this and more information as a tsv file. If you want to export other things have a look there. Going to close this for that reason.

PR #1998, opened 2023-04-03
Modify `ord list` output to include the end of each range and the off…

gmart7t2:
Modify `ord list` output to include the end of each range and the offset into the output of the start of each range.

Sometimes when sat hunting I will find a sat that I want to isolate and it will be in the 1000th range in its output. I want to use `ord wallet send` to send the sat to a new address but to do that i need to know its satpoint, which means its `txid`, `vout`, and `offset`. 

The `txid` and `vout` are obvious. They define the output I am looking at. But to know the offset I have to sum the sizes of the 999 ranges before the sat I am interested in.

I figure it is better to have `ord` do that summing than to expect the user to do it for themselves., so that's what this change does.


soenkehahn:
Thanks for the PR! Overall this looks like a useful feature.

Here's a few comments:
- Could you add a description of your motivation/use-case for this feature, please? It's easy to imagine use-cases, but it might help review this issue -- especially for people less familiar with all the details of `ord` -- if we have a very concrete example for when this would be good to have. (I think the first comment aka the issue description is a good place for that.)
- Could you make all the checks pass? Currently for example, the tests don't compile.
- Could you add some tests for this feature, please? Generally we are aiming for very comprehensive test coverage for `ord`. Additionally, adding a test may make it much easier for more people to review the PR, since the tests document what is being changed. (And prove that the changes work.)

I'll turn this PR into a draft PR. Please, undraft when you think it's ready for another look.

Thanks again for contributing!

gmart7t2:
> * Could you add a description of your motivation/use-case for this feature, please? 

Done.

> * Could you make all the checks pass? Currently for example, the tests don't compile.

I can't reproduce that. Everything seems to be passing. Do you have details of what's failing?

> * Could you add some tests for this feature, please?

I modified the existing test to expect the new output fields. 

> I'll turn this PR into a draft PR. Please, undraft when you think it's ready for another look.

Ok.

casey:
@soenkehahn Want to give this another look?

soenkehahn:
I'd love to. This feature needs a `--index-sats` index, I'm in the process of acquiring one. I felt I couldn't really review this without being able to try it out myself.

LiHuanLH:
> This feature needs a `--index-sats`

can anyone tell me what is `--index-sats`? how to write it?

soenkehahn:
@LiHuanLH: It's a command line flag when building `ord`'s database (with `ord index`) that means that the database tracks all satoshis. [Here](https://docs.ordinals.com/guides/sat-hunting.html?highlight=index-sats#sat-hunting)'s some sparse information. Also, `ord --help` says:

```
--index-sats
            Track location of all satoshis.
```

But I think this question is a bit off-topic and doesn't belong in this particular issue. A good place for asking might be the [discord](https://discord.gg/ordinals).

src/subcommand/list.rs
```diff
@@ -78,23 +87,29 @@ mod tests {
...
         (
           OutPoint::from_str("4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b:5")
             .unwrap(),
           1050000000000000,
+          1150000000000000,
           100000000000000,
+          5 * COIN_VALUE + 90,
           Rarity::Epic,
           "gkjbdrhkfqf".to_string()
         )
```
soenkehahn, line 115: In addition to fixing the existing tests, I think we should have tests that are dedicated to the new fields `end` and `offset`. I think ideally their `assert_eq` would only work on the tested fields, so that it looks something like this:
```rust
assert_eq!(outputs.into_iter().map(|output| output.offset).collect(), vec![0, 10, 30]);
```
And they should have speaking names, like `list_calculates_output_offsets`.

In general the idea is to have very good test coverage, but also to use tests to document the code. For example if someone in the future breaks the offset calculation, it'd be really nice if one of the failing tests was very clearly testing offset calculation. Hope this makes sense.

These tests could also be added in `tests/list.rs`. Not sure what I'd prefer.

soenkehahn (changes requested):
Sorry it took so long to review!

soenkehahn:
> > * Could you make all the checks pass? Currently for example, the tests don't compile.
> 
> I can't reproduce that. Everything seems to be passing. Do you have details of what's failing?

Here's the failing log: https://github.com/casey/ord/actions/runs/4591718231/jobs/8108123525. But you fixed it in later commits.

raphjaph (approved):
Made the changes Casey requested.

raphjaph (approved):

casey (approved):

PR #2002, opened 2023-04-07
Implement 'export' subcommand to write inscriptions to the file system

soenkehahn:
The `export` subcommand will write out inscriptions into the filesystem. This provides an easy but fairly powerful (read-only) interface to all inscriptions. For example it makes it easy to `grep` through all inscriptions, or filter by filetype or size, etc.

This has some security implications: `ord export` will download files from completely untrusted sources to user's disks. And they may use programs that treat local files as more than completely untrusted. (E.g. when opening the `inscriptions` directory in a file browser, it may create thumbnails or previews for all inscription files. And -- while this is a very fun way to explore inscriptions -- the file browser may not be safe to run on untrusted files in this way.)

Here's the `ord export --help` output:
```
ord-export 
Write all inscriptions out into files. This will write files from untrusted sources into your local
filesystem. This may be a security risk, if programs treat local files not as completely untrusted.

USAGE:
    ord export [OPTIONS]

OPTIONS:
    -h, --help
            Print help information

        --output-dir <OUTPUT_DIR>
            Write to directory <OUTPUT_DIR>
            
            [default: inscriptions]
```

Here's the output of `ord export ; ls inscriptions/numbers | sort -h | head`:

```
0.png
1.png
2.gif
3.png
4.png
5.txt
6.png
7.html
8.png
9.webp
```

Currently it writes all inscriptions into a `numbers` subdirectory. This is done this way, in case we ever want to also write (the same) inscriptions but with different file names, e.g. by inscription id or date. There's lots of design questions to think about before implementing that, but I would like to punt on that for now to avoid scope creep.

Up for bikeshedding:

- Is `ord export` a good name for the subcommand?
- Are the directory names (`inscriptions/numbers`) good?
- Should there be a `--please-write-untrusted-files-to-my-working-directory` flag to force users to acknowledge the security risk?
- Should the help text be different?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ rust-version = "1.67"
 members = [".", "test-bitcoincore-rpc"]
 
 [dependencies]
+accept-header = "0.2.3"
 anyhow = { version = "1.0.56", features = ["backtrace"] }
 axum = { version = "0.6.1", features = ["headers"] }
 axum-server = "0.4.0"
--- a/src/index.rs
+++ b/src/index.rs
@@ -18,7 +18,7 @@ use {
   std::sync::atomic::{self, AtomicBool},
 };
 
-mod entry;
+pub(crate) mod entry;
 mod fetcher;
 mod rtx;
 mod updater;
@@ -902,19 +902,19 @@ impl Index {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
   use {
     super::*,
     bitcoin::secp256k1::rand::{self, RngCore},
   };
 
-  struct ContextBuilder {
+  pub(crate) struct ContextBuilder {
     args: Vec<OsString>,
     tempdir: Option<TempDir>,
   }
 
   impl ContextBuilder {
-    fn build(self) -> Context {
+    pub(crate) fn build(self) -> Context {
       self.try_build().unwrap()
     }
 
@@ -966,23 +966,23 @@ mod tests {
     }
   }
 
-  struct Context {
+  pub(crate) struct Context {
     options: Options,
-    rpc_server: test_bitcoincore_rpc::Handle,
+    pub(crate) rpc_server: test_bitcoincore_rpc::Handle,
     #[allow(unused)]
-    tempdir: TempDir,
-    index: Index,
+    pub(crate) tempdir: TempDir,
+    pub(crate) index: Index,
   }
 
   impl Context {
-    fn builder() -> ContextBuilder {
+    pub(crate) fn builder() -> ContextBuilder {
       ContextBuilder {
         args: Vec::new(),
         tempdir: None,
       }
     }
 
-    fn mine_blocks(&self, n: u64) -> Vec<Block> {
+    pub(crate) fn mine_blocks(&self, n: u64) -> Vec<Block> {
       let blocks = self.rpc_server.mine_blocks(n);
       self.index.update().unwrap();
       blocks
--- a/src/index.rs
+++ b/src/index.rs
@@ -939,7 +939,7 @@ pub(crate) mod tests {
       ];
 
       let options = Options::try_parse_from(command.into_iter().chain(self.args)).unwrap();
-      let index = Index::open(&options)?;
+      let index = Arc::new(Index::open(&options)?);
       index.update().unwrap();
 
       Ok(Context {
@@ -971,7 +971,7 @@ pub(crate) mod tests {
     pub(crate) rpc_server: test_bitcoincore_rpc::Handle,
     #[allow(unused)]
     pub(crate) tempdir: TempDir,
-    pub(crate) index: Index,
+    index: Arc<Index>,
   }
 
   impl Context {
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version = "1.67"
 members = [".", "test-bitcoincore-rpc"]
 
 [dependencies]
+accept-header = "0.2.3"
```
soenkehahn, line 17: See `src/media.rs` for why I added this.

src/media.rs
```diff
@@ -132,6 +146,34 @@ mod tests {
...
+  fn extension_for_content_type() -> Result {
+    let test_cases = vec![
+      ("application/json", "json"),
+      ("image/jpeg", "jpg"),
+      ("image/png", "png"),
+      ("image/svg+xml", "svg"),
+      ("text/html;charset=utf-8", "html"),
+      ("text/plain;charset=utf-8", "txt"),
+      ("text/html", "html"),
+      ("text/plain", "txt"),
```
soenkehahn, line 159: There are inscriptions in `mainnet` with content type `text/plain;charset=utf-8` and others with just `text/plain` (even though `ord` will not create those, at least in its currrent version). This PR also pulls in `accept-header` for lenient content type parsing.

src/index.rs
```diff
@@ -966,23 +966,27 @@ mod tests {
     }
   }
 
-  struct Context {
+  pub(crate) struct Context {
```
soenkehahn, line 969: This is the test `Context` that before this PR was only used in the tests here in `index`. Now the tests in `export` also use this `Context`, so I made it `pub(crate)` (and changed it a little). Note that for production code, this is still not visible or exported, since the whole `test` module is guarded through `#[cfg(test)]`.

It feels a bit weird to import test helpers from another test module, so I would be up for moving this to a dedicated test helpers module. Although it does work fine and I actually like it, now that I'm used to it. I *do* think these common test helpers should not be duplicated. Any opinions?

src/subcommand/export.rs
```diff
@@ -0,0 +1,407 @@
...
+  fn run_with_index(&self, index: Arc<Index>) -> Result {
+    self.prepare_directory_tree()?;
+    let written_numbers = self.get_written_numbers()?;
+    let ids = index.get_inscriptions(None)?;
+    let index = index.as_ref();
+    for id in Export::add_progress_bar(ids.into_values()) {
+      Export::retry(|| {
+        let entry = index
+          .get_inscription_entry(id)?
+          .ok_or_else(|| anyhow!("inscription entry not found: {id}"))?;
```
soenkehahn, line 35: The `entry` is only looked up to get the inscription number. Is there a better way to do that?

src/subcommand/export.rs
```diff
@@ -0,0 +1,338 @@
...
+
+#[cfg(test)]
+mod test {
+  use {
+    super::*,
+    crate::{index::tests::Context, test::inscription},
+    pretty_assertions::assert_eq,
+  };
+
+  impl Context {
```
soenkehahn, line 154: These are additions to `index::test::Context` that are only relevant for `export` tests. Personally I like this structure, but I could see how someone might find this confusing. I'd be happy to move `Context` into a common test helper module.

casey:
Sorry for never actually reviewing this. I don't know if you have any inclination to ressurect this, but if you want to, please rebase and reopen.

Jrbiltmore:
import osimport argparsedef parse_arguments():    parser = argparse.ArgumentParser(        description='Safely export inscriptions from untrusted sources to your local filesystem. Includes security measures to mitigate risks.')    parser.add_argument('--output-dir', default='inscriptions', help='Output directory [default: inscriptions]')    parser.add_argument('--acknowledge-risk', action='store_true',                        help='Acknowledge the security risk of writing untrusted files to the filesystem')    return parser.parse_args()def is_executable(file_name):    executable_extensions = ['.exe', '.bat', '.sh', '.js', '.cmd', '.scr', '.ps1', '.dll', '.so', '.app', '.jar']    return any(file_name.lower().endswith(ext) for ext in executable_extensions)def export_inscriptions(output_dir, acknowledge_risk):    if not acknowledge_risk:        print("Security risk not acknowledged. Use --acknowledge-risk to proceed. Aborting.")        return    numbers_dir = os.path.join(output_dir, "numbers")    os.makedirs(numbers_dir, exist_ok=True)    print(f"Exporting inscriptions to {numbers_dir}...")    for i in range(10):  # Simulated loop to represent exporting files        file_name = f"{i}.txt"  # Simulated file name        file_path = os.path.join(numbers_dir, file_name)                if is_executable(file_name):            print(f"Skipping potential executable file: {file_name}")            continue        with open(file_path, 'w') as file:            file.write(f"Dummy inscription content for {i}")            # Implement malware scanning here            # if malware_detected(file_path):            #     print(f"Malware detected in {file_name}. File removed.")            #     os.remove(file_path)            #     continue    print("Export completed. Remember to handle the exported files with care.")def main():    args = parse_arguments()    export_inscriptions(args.output_dir, args.acknowledge_risk)if __name__ == "__main__":    main()Sent from my iPhoneOn 12. 2. 2024., at 12:38, Casey Rodarmor ***@***.***> wrote:﻿
Sorry for never actually reviewing this. I don't know if you have any inclination to ressurect this, but if you want to, please rebase and reopen.

Issue #2000, opened 2023-04-05
Indexer ignores inscriptions that are not on first input of TX, but otherwise valid causing missing inscription numbers


veryordinally:
While working on #783 @raphjaph and @veryordinally discovered a bug in the indexer that affects versions of `ord` up to and including 0.5.1: inscriptions that are not made on the first input are ignored during indexing. This was due to a simplifying assumptions in `src/inscription.rs` where only the first input was parsed for inscriptions. We fixed that with this [commit](https://github.com/casey/ord/pull/1963/commits/fbc4cdec56393136cd9995b71972a70537b529b7) as part of draft PR #1963 

Discussed with @raphjaph and we suggest to disentangle this fix from the larger collections and provenance topic, and also to address this sooner, as it will cause inscription numbers to change since indexers with this fix applied will index those inscriptions that the previous indexer skipped. I will compile and add a list of inscriptions that are currently ignored to this issue.

veryordinally:
That would seem like a good mechanism to do these kinds of updates/fixes to me. Would also be an option for collections/provenance.

gmart7t2:
> While working on #783 @raphjaph and @veryordinally discovered a bug in the indexer that affects versions of `ord` up to and including 0.5.1: inscriptions that are not made on the first input are ignored during indexing

That is how the docs say it should be:

https://docs.ordinals.com/inscriptions.html

![Screenshot_20230406_093102_Presearch](https://user-images.githubusercontent.com/49558347/230379492-50dfeaae-f2c5-4b4a-963c-260750a60d74.jpg)

The docs and the code agree that only the first input/output is an inscription

> that add in some sense arbitrary constraints on what gets parsed as an inscription

The whole thing is already arbitrary. Adding an extra constraint that allows multiple inscriptions per reveal tx after a certain block height is no different.

timechainz:
> The docs and the code agree that only the first input/output is an inscription

@gmart7t2  If I'm not mistaken, the docs you highlight refer to the first output on which the first sat contains the inscription, but the aforementioned bug relates to the INPUT of the reveal transaction.  The code was only looking in the first input.  That's not consistent with the doc which doesn't specify it's the first input.  I believe that is why @veryordinally is referring to it as a bug.
![docs](https://user-images.githubusercontent.com/126361786/230498272-9df56065-2a36-47e7-b197-79d2d6be025f.jpeg)
[...]

PR #2006, opened 2023-04-11
Identifying inscriptions

raphjaph:
At the moment `ord` only parses new inscriptions that are in the first input of a transaction. This PR rectifies this mistake by also parsing inscriptions on other transaction inputs. This will make future improvements like batch inscriptions and parent-child relationships easier. Out of this follows that there can now be `inscription_id`s that end with `...i1`,`...i54`/, etc. (not only `...i0`). The number at the end is just the index of the inscription in the transaction. Another question is which sat an inscription is inscribed on. At the moment it is the first sat of the first output but this is implicit. If there is multiple inscription inputs but only one output, which sat is the second inscription inscribed on? Does there have to be a seperate output for the inscription to be valid or is it just discarded if a corresponding output is missing? Through an optional (uneven) tag we could build pointers to the sat the inscription is supposed to be on. It could specify the sat offset in the transaction as the sat this inscription is inscribed on. Here with offset=1000:
```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_3
  OP_PUSH 1000
  OP_1
  OP_0
  OP_PUSH "first inscription"
OP_ENDIF
```

Thinking even further a more general consensus question arises as well. Should it be possible to have more than one inscription *per input*? Practically, this would be achieved by putting multiple of our envelope constructions in the input's witness, like so:
```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_1
  OP_0
  OP_PUSH "first inscription"
OP_ENDIF
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_1
  OP_0
  OP_PUSH "second inscription"
OP_ENDIF
```
Having multiple inscriptions per input would make batch inscriptions way more space efficient since an inscription wouldn't need an extra input.

#2000 
#1769

### DIFF
--- a/src/index/updater/inscription_updater.rs
+++ b/src/index/updater/inscription_updater.rs
@@ -106,7 +106,7 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
             index: id_index,
           };
 
-          id_index += 1; 
+          id_index += 1;
 
           floating_inscriptions.push(Flotsam {
             inscription_id,
--- a/src/index/updater/inscription_updater.rs
+++ b/src/index/updater/inscription_updater.rs
@@ -76,6 +76,7 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
     let mut floating_inscriptions = Vec::new();
     let mut inscribed_offsets = BTreeSet::new();
     let mut input_value = 0;
+    let mut id_index = 0;
     for tx_in in &tx.input {
       // skip subsidy since no inscriptions possible
       if tx_in.previous_output.is_null() {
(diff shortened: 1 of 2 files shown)


veryordinally:
This PR addresses #2000 

I would suggest to move the broader protocol questions regarding multiple inscriptions per transaction and multiple inscriptions per input to separate issues to facilitate discussion and review.

casey:
This is not ready for a PR, because we haven't decided how to proceed. Changing it to a draft to reflect this.

casey:
I'm on mobile, and I couldn't figure out how to make this PR a draft, so I closed it instead.

raphjaph:
How do we decide how to proceed?

casey:
This is a big change, and we've started tackling smaller pieces independently, so this can be closed.

PR #2008, opened 2023-04-12
Handle interrupts better

terror:
This PR changes the amount of times the user has to hit Ctrl+C in order to kill the process from 7 to 2, this makes it easier for the user to immediately shutdown the process.

### DIFF
--- a/src/index/updater.rs
+++ b/src/index/updater.rs
@@ -159,7 +159,7 @@ impl Updater {
           )?;
       }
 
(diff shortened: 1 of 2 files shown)


casey:
Could you add a description? Without a description a maintainer will have to look at the code to figure out what the PR does.

casey (changes requested):

casey (approved):
LGTM!

casey:
I'm merging without tests because testing this would be a huge pain, and we didn't have any tests before, so in this case, YOLO.

PR #2028, opened 2023-04-18
Tweak --config and --config-dir help text

soenkehahn:
I thought that these options would also be used when running other subcommands, e.g. `ord index`. I think this PR clarifies that they're only used for `ord server`.

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -16,9 +16,9 @@ pub(crate) struct Options {
     help = "Use <CHAIN>."
   )]
   pub(crate) chain_argument: Chain,
-  #[clap(long, help = "Load configuration from <CONFIG>.")]
+  #[clap(long, help = "Load server configuration from <CONFIG>.")]
   pub(crate) config: Option<PathBuf>,
-  #[clap(long, help = "Load configuration from <CONFIG_DIR>.")]
+  #[clap(long, help = "Load server configuration from <CONFIG_DIR>.")]
   pub(crate) config_dir: Option<PathBuf>,
   #[clap(long, help = "Load Bitcoin Core RPC cookie file from <COOKIE_FILE>.")]
   pub(crate) cookie_file: Option<PathBuf>,


soenkehahn:
@raphjaph: Will #1527 mean that the config file will also be used for other commands, e.g. `ord index`?

casey:
Might this be something that we just wind up undoing in the future? I.e. we start using the config file for other commands, and now the help text is out of date. I'm not opposed though, especially if it's something you found confusing.

https://www.youtube.com/watch?v=CxK_nA2iVXw

casey (changes requested):
Adding review so it shows up as having changes requested.

soenkehahn:
> Might this be something that we just wind up undoing in the future? I.e. we start using the config file for other commands, and now the help text is out of date.

Very much yes. It might. But we can't predict the future and I strongly feel we should try to have the best help output for the present behavior. To put it differently: If we change the behavior of `ord`, of course we have to make sure that the help texts get updated.

> I'm not opposed though, especially if it's something you found confusing.

Yeah, I found it very confusing. I was hoping to be able to put the `--data-dir` setting into the config file. I ended up looking at the code and using `rust-analyzer` to figure out that the config is only used for `ord server`.

casey:
One issue I see is that really, the config file only supports a subset of the CLI configuration. It isn't just specific to the server, it's also specific to hiding inscriptions, since that's the only config option. So this change would make someone not look at the config file when trying to configure non-server commands, but wouldn't help someone who was trying to use the config file to configure options other than hiding things.

This makes me think that perhaps the best solution to this problem is actually just documenting which options are supported in the config file, and pointing to that documentation. That documentation (probably in the form of an example config file with all options) would then make it obvious that the only supported config option is hiding inscriptions when using the server.

soenkehahn:
@casey: I think what you say is true and makes sense. I still think it shouldn't stop this PR from being merged. This PR *does* make the help texts more accurate and helpful, while not solving *all* problems you point out. So I would still merge this. And -- if this is a high priority -- further improve documentation of the config file in other issues and PRs.

soenkehahn:
Closing in favor of #2044.

aguinaldok4:
'sever deleted _0fd079902825c00ed761eab06ff2f5cd8c544d57#*

PR #2033, opened 2023-04-19
Copy + Format Changes to Gifts Page

Isabelfoxenduke:
There has been some confusion within the Ordinals community about how to share gifts with the Ordinals Project core team. Specifically, people are confused about which wallet addresses they can safely send gifts to - and whether or not they can send Bitcoin and/or inscriptions to those wallet addresses safely. To address this confusion, I recommend making the following changes to the “Gifts” page at https://docs.ordinals.com/donate.html.

PR #2034, opened 2023-04-20
Improve donation page

casey:
I looked at #2033, but realized I'm just too particular, and wanted to write my own prose.

This attempts to solve the following problems:

- People don't realize that the donation addresses correspond to physical opendimes
- People don't realize that inscriptions can be sent to the donation addresses.


soenkehahn (approved):
LGTM. Made sure the video is embedded ok, too, at least on firefox.

raphjaph (approved):
LGTM! Let's get the money

PR #2037, opened 2023-04-22
Switch to clippy nightly

### LINKED ISSUE #2035: Fix clippy
Clippy is currently failing, I think because of this: https://github.com/rust-lang/rust-clippy/pull/10499

The failure is in clap, a dependency. I don't know how to suppress this error. Also, I think maybe it doesn't happen in clap 4.0, but we're blocked on upgrading to 4.0 on them supporting color in help text.

soenkehahn:
This PR switches `clippy` to `nightly`, both in ci scripts and for `just clippy` and friends. This is to workaround [this issue](https://github.com/clap-rs/clap/issues/4733), which I believe is fixed [here](https://github.com/rust-lang/rust-clippy/pull/10502), but hasn't landed in the stable version of `clippy`. This PR should be reverted once the fix lands.

Fixes #2035.

Of course this has some downsides:
- Nightly `clippy` is not stable, so we may have some CI issues introduced in newer nightly versions. (We could try to pin down `clippy` to an exact nightly version, which would avoid this. Any opinions?)
- CI caches get bigger.
- Before locally running `just clippy` (or e.g. `just ci`) you have to run `rustup install nightly`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,17 +49,25 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Install Rust Toolchain Components
+    - name: Install rustfmt
       uses: actions-rs/toolchain@v1
       with:
-        components: clippy, rustfmt
+        components: rustfmt
         override: true
         toolchain: stable
 
+    - name: Install clippy nightly
+      uses: actions-rs/toolchain@v1
+      with:
+        components: clippy
+        toolchain: nightly
+
     - uses: Swatinem/rust-cache@v2
+      with:
+        key: "clippy-nightly"
 
     - name: Clippy
-      run: cargo clippy --all --all-targets
+      run: cargo +nightly clippy --all --all-targets
 
     - name: Format
       run: cargo fmt --all -- --check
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -49,20 +49,30 @@ jobs:
...
     - uses: actions/checkout@v2
 
-    - name: Install Rust Toolchain Components
+    - name: Install rustfmt
       uses: actions-rs/toolchain@v1
       with:
-        components: clippy, rustfmt
+        components: rustfmt
         override: true
         toolchain: stable
```
soenkehahn, line 57: `rustfmt` is still installed as the stable version.

.github/workflows/ci.yaml
```diff
@@ -49,20 +49,30 @@ jobs:
...
 
+    - name: Install clippy nightly
+      uses: actions-rs/toolchain@v1
+      with:
+        components: clippy
+        toolchain: nightly
+
     - uses: Swatinem/rust-cache@v2
+      with:
+        key: "clippy-nightly"
```
soenkehahn, line 67: Hopefully this will mean that -- once we revert this PR -- we will not suffer from bigger caches anymore.

PR #2064, opened 2023-05-05
ord wallet inscriptions Can't get the inscription link correctly

daidean:
The link in the code is still ordinals.com in btc, it should be changed to ordinalslite.com


daidean:
Sorry, wrong place to mention, please delete

PR #2107, opened 2023-05-23
Inscriptions with no sat are unbound

casey:


raphjaph:
Is this ready for review?

casey:
@raphjaph Not yet, I tried it out, and i'm getting an error when I visit the `/inscription` page.

src/index/updater/inscription_updater.rs
```diff
@@ -111,13 +114,27 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
     if inscriptions.iter().all(|flotsam| flotsam.offset != 0)
       && Inscription::from_transaction(tx).is_some()
     {
-      inscriptions.push(Flotsam {
+      let flotsam = Flotsam {
         inscription_id: txid.into(),
         offset: 0,
         origin: Origin::New {
           fee: input_value - tx.output.iter().map(|txout| txout.value).sum::<u64>(),
```
raphjaph, line 121: I know this is probably not possible but should this be `.checked_sub()`?

raphjaph (commented):
lgtm

raphjaph:
> @raphjaph Not yet, I tried it out, and i'm getting an error when I visit the `/inscription` page.

Probably because you're returning a Option<Sat> instead of a Sat and the boilerplate template expects it. But going to bed now. I'll look tomorrow

casey:
@raphjaph Should be ready for review! When requesting `/inscription`, the server tries to look up the txout, i.e. the output value and script, for an inscription. Unbound inscriptions are on the 0…0:0 output, which doesn't exist, so this was failing. I changed the template to accept an `Option<TxOut>`, and return `None` in the case of unbound inscriptions.

raphjaph (approved):
Breaker of JPEG no more!

PR #2112, opened 2023-05-25
[WIP] Add manually cursed tag OP_66

dannydeezy:
Lots of people went too degen and deliberately tried to front-run the cursed inscriptions concept by pushing OP_66 (a previously unrecognized even tag). However we did not realize this would make the inscriptions unbound! Lots of people will feel rugged if these OP_66 inscriptions are unbound, so I think we should recognize and ignore OP_66, allowing these to be cursed but _not_ unbound (i.e. they should be bound like a normal inscription).

### DIFF
--- a/src/inscription.rs
+++ b/src/inscription.rs
@@ -15,6 +15,7 @@ const PROTOCOL_ID: &[u8] = b"ord";
 
 const BODY_TAG: &[u8] = &[];
 const CONTENT_TYPE_TAG: &[u8] = &[1];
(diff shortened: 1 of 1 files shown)


dannydeezy:
closing in favor of one off of raphjaphs existing branch

dannydeezy:
the PR to recognize OP_66 as cursed (negative) but not unbound is here: https://github.com/raphjaph/ord/pull/10

PR #2125, opened 2023-05-29
Update donate.md for inscriptions donations.

veryordinally:
Added separate donation address for inscriptions to make managing the wallet easier.

@casey @raphjaph please review and carefully check address.

### DIFF
--- a/docs/src/donate.md
+++ b/docs/src/donate.md
@@ -5,16 +5,16 @@ Ordinals is open-source and community funded. The current lead maintainer of
 `ord` is [raphjaph](https://github.com/raphjaph/). Raph's work on `ord` is
 entirely funded by donations. If you can, please consider donating!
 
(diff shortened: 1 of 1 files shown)


casey:
Looks good to me, although what do you think about using the second address in the wallet? It seems very slightly more natural and easy to remember.

veryordinally:
> Looks good to me, although what do you think about using the second address in the wallet? It seems very slightly more natural and easy to remember.

We already have one donation there which maybe is easier to keep separately.

casey:
> > Looks good to me, although what do you think about using the second address in the wallet? It seems very slightly more natural and easy to remember.
> 
> We already have one donation there which maybe is easier to keep separately.

Fine by me!

casey (approved):

raphjaph (approved):

PR #2135, opened 2023-05-31
`OP_PUSH` instead of `OP_1`

raphjaph:
Technically should be a `OP_PUSHDATA1` but this is simplified overview.

### DIFF
--- a/docs/src/inscriptions.md
+++ b/docs/src/inscriptions.md
@@ -43,9 +43,9 @@ follows:
 OP_FALSE
 OP_IF
   OP_PUSH "ord"
-  OP_1
+  OP_PUSH 1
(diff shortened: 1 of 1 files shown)


veryordinally (approved):
This looks good to me.

One point to consider: do we want to mention that earlier versions of the docs had OP_1 instead of OP_PUSH 1? Bit of a philosophical question, if we just want up to date docs, or also include historical info. I'd probably gravitate to not include changes explicitly - that's what we have Github for, after all.

raphjaph:
This is a simplified overview anyway, so I don't think we need to include historical changes.

kevinfaveri:
I guess it is '5' tho if you are using bitcoinjs lib to compile

PR #2141, opened 2023-06-02
Update redb from 0.13.0 to 1.0.2

raphjaph:
Still have to run some benchmarks.


veryordinally:
Converting to draft until we resolve the performance issues.

veryordinally:
Seeing somewhat of an improvement by setting 8 GiB cache like this:

```
        let database = Database::builder()
        .set_cache_size(8589934592)
        .create(&path)
        .unwrap();
```

Benchmarks in https://github.com/cberner/redb/issues/603

veryordinally:
See also https://github.com/ordinals/ord/issues/2183

raphjaph:
I updated to 1.0.1. It's the second stable release for redb

src/index.rs
```diff
@@ -185,23 +185,17 @@ impl Index {
...
-              WriteStrategy::Checksum
-            } else {
-              WriteStrategy::TwoPhase
-            })
-            .create_mmapped(&path)?
-        };
-        let tx = database.begin_write()?;
+      Err(_) => {
+        let database = Database::builder()
+          .set_cache_size(17179869184)
```
veryordinally, line 190: Make configurable

src/index.rs
```diff
@@ -185,23 +185,17 @@ impl Index {
...
 
-        #[cfg(test)]
-        let tx = {
-          let mut tx = tx;
-          tx.set_durability(redb::Durability::None);
-          tx
+        if cfg!(test) {
+          tx.set_durability(redb::Durability::Eventual);
+        } else {
+          tx.set_durability(redb::Durability::Immediate);
```
veryordinally, line 198: Before we had None for tests, now we have immediate - was that intended?

src/index.rs
```diff
@@ -523,6 +522,7 @@ impl Index {
     self.client.get_block(&hash).into_option()
   }
 
+  /// This gets the first inscription on a sat, with reinscriptions enabled it would return a Vec
```
veryordinally, line 525: The comment is obsolete, since we now have reinscriptions. We should probably make sure here to always return the chronologically first inscription for a given sat using our ordering table.

veryordinally (changes requested):
- parameter for cache with 8GiB default as you already mentioned
- should think through which inscription to return for sat - currently not defined I think, would be better to use ordering IMO

src/index.rs
raphjaph: pulling out and making new PR

src/index.rs
raphjaph: None for tests and Immediate for everything else

raphjaph:
Waiting for https://github.com/cberner/redb/issues/626 to resolve.

veryordinally (approved):
LGTM!

Issue #2139, opened 2023-06-01
Support of inscription reveal through P2wsh


lgalabru:
I see a few instances of recent transactions including what appears to be a valid inscription, not being accepted by `ord`.
[de7401dcef9c694b7326051263b8d5342dada9fdb0eb732e04022e324f6d9b36i0](https://blockstream.info/tx/de7401dcef9c694b7326051263b8d5342dada9fdb0eb732e04022e324f6d9b36?input:0) being one of them.

314159265359879:
Here is another one
https://ordinals.hiro.so/inscription/de7401dcef9c694b7326051263b8d5342dada9fdb0eb732e04022e324f6d9b36i0

Leonidas pointed this one out 
_There were 3 weird inscriptions made last night that none of the other indexers count but that Hiro does. This is resulting in all inscription numbers after '10366012' being off in Hiro's Ordinals API. '10366013' which is de7401dcef9c694b7326051263b8d5342dada9fdb0eb732e04022e324f6d9b36i0 is not a valid inscription and along with the other two invalid inscriptions it should be removed and the numbers recalculated after 10366012. Also to prevent future weird inscriptions like this from being included BobBodily on twitter can share how he did them._

Mentioned twitter thread? https://twitter.com/BobBodily/status/1664304114782838786

lgalabru:
Thanks @314159265359879 for helping me connecting the dots. 
I think Hiro will unfortunately have to disable P2wsh inscription indexing until it is officially supported by `ord` - in a backward compatible fashion.

314159265359879:
This is one of the other inscriptions mentioned by Leonidas:

59dedca28c59e01aa394d002bf13ae2f0dd627353b27825264367aa87c444eaci0
https://ordinals.hiro.so/inscription/59dedca28c59e01aa394d002bf13ae2f0dd627353b27825264367aa87c444eaci0

They are checking with Bob for others.

lgalabru:
Thanks @314159265359879 ! We'll remove these entries from our current db and repair the sequence hopefully tomorrow.
We are in the process of building a new index with `hord`, taking into account cursed inscriptions. I just added the `W2psh` curse, so we that we can retrieve these inscriptions in the negative inscription number sequence.

bodily11:
Sorry to break some indexers. I just really want support for P2WSH in ordinals indexers. I have a dev working on a PR to support them in Ord right now. Then we can see at what block height it makes sense to remove the P2WSH "curse".

[...]

PR #2109, opened 2023-05-24
Unrecognized even tags are unbound and assigned negative inscription numbers

raphjaph:

Issue #2045, opened 2023-04-25
Cursed Inscription Tracking Issue


casey:
It would be desirable to modify `ord` to recognize the following:

- Multiple inscriptions per transaction, for efficient batching
- Inscriptions on inputs after the first, which is useful for collections
- Multiple inscriptions on the same sat, so that the entire history of a sat doesn't need to be checked to determine if a new inscription is valid
- Inscriptions with unrecognized even headers, so that new even headers don't cause upgraded clients to disagree about inscription numbers

An idea for how to do this is:

- Modify ord to recognize the above currently invalid inscriptions, including retroactively in old blocks, but consider these new inscriptions "cursed" and assign them negative inscription numbers. Since they would not receive positive inscription numbers, they would not disrupt existing inscription numbers
- At a future, pre-announced block height X, new inscriptions that would have been cursed before X are now not cursed, or "blessed", and receive a positive inscription number instead of a negative one.

The reasoning for this strange scheme is as follows:

- I don't like changing old inscription numbers by fiat.
- I want to make previously invalid inscriptions appear on [ordinals.com](http://ordinals.com/) as soon as possible, and giving them a negative inscription number allows that with minimal changes to the code. Hopefully making them visible will help further the conversation, let people see how many exist, and make everything more concrete.
- Having a pre-announced block height at which previously invalid inscriptions would now be valid allows ample time for everyone to upgrade.
- Negative inscription numbers would always be perma-unstable.
- We could use this same scheme again in the future. Some new way of making inscriptions is desirable, they initially receive negative numbers, and at a pre-announced flag day they enter the main sequence.

dannydeezy:
let's curse these babays!!!

lgalabru:
Are you saying that batched inscriptions - which should actually be the go to / most used pattern - will be "cursed"?
How do you envision handling bugs within "cursed" inscriptions?

paradisd:
Yeahhhh we curse!!!!

yueliulin:
How to identify which inscriptions are cursed

dannydeezy:
i'm concerned about https://github.com/casey/ord/pull/2109 which makes it appear that inscriptions with unrecognized even tags will be "unbound" which i think means "not assigned to a sat"

what is the reasoning for this decision? I would think they should be assigned to the top sat as if the even tags weren't there?

DrJingLee:
Now the cursed inscriptions are on fire....

penguin4238:
its fun

[...]

PR #2145, opened 2023-06-03
Cursed Inscriptions [1/n]

raphjaph:
Recognizes multiple inscriptions per transaction:
1. in different inputs
2. in the same input

All but the first inscription in the first input are cursed (for now). In addition all but the first inscription in an input are made unbound (they will be bound in the future as reinscriptions).

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -69,6 +69,7 @@ pub(crate) enum Statistic {
   LostSats = 2,
   OutputsTraversed = 3,
   SatRanges = 4,
+  UnboundInscriptions = 5,
 }
 
 impl Statistic {
@@ -801,7 +802,7 @@ impl Index {
     &self,
     inscription_id: InscriptionId,
     satpoint: SatPoint,
-    sat: u64,
+    sat: Option<u64>,
   ) {
     let rtx = self.database.begin_read().unwrap();
 
@@ -836,32 +837,34 @@ impl Index {
       inscription_id,
     );
 
-    if self.has_sat_index().unwrap() {
-      assert_eq!(
-        InscriptionId::load(
-          *rtx
-            .open_table(SAT_TO_INSCRIPTION_ID)
-            .unwrap()
-            .get(&sat)
-            .unwrap()
-            .unwrap()
-            .value()
-        ),
-        inscription_id,
-      );
+    if let Some(sat) = sat {
+      if self.has_sat_index().unwrap() {
+        assert_eq!(
+          InscriptionId::load(
+            *rtx
+              .open_table(SAT_TO_INSCRIPTION_ID)
+              .unwrap()
+              .get(&sat)
+              .unwrap()
+              .unwrap()
+              .value()
+          ),
+          inscription_id,
+        );
 
-      assert_eq!(
-        SatPoint::load(
-          *rtx
-            .open_table(SAT_TO_SATPOINT)
-            .unwrap()
-            .get(&sat)
-            .unwrap()
-            .unwrap()
-            .value()
-        ),
-        satpoint,
-      );
+        assert_eq!(
+          SatPoint::load(
+            *rtx
+              .open_table(SAT_TO_SATPOINT)
+              .unwrap()
+              .get(&sat)
+              .unwrap()
+              .unwrap()
+              .value()
+          ),
+          satpoint,
+        );
+      }
     }
   }
 
@@ -1378,7 +1381,70 @@ mod tests {
           outpoint: OutPoint { txid, vout: 0 },
           offset: 0,
         },
-        50 * COIN_VALUE,
+        Some(50 * COIN_VALUE),
+      );
+    }
+  }
+
+  #[test]
+  fn inscriptions_without_sats_are_unbound() {
+    for context in Context::configurations() {
+      context.mine_blocks(1);
+
+      context.rpc_server.broadcast_tx(TransactionTemplate {
+        inputs: &[(1, 0, 0)],
+        fee: 50 * 100_000_000,
+        ..Default::default()
+      });
+
+      context.mine_blocks(1);
+
+      let txid = context.rpc_server.broadcast_tx(TransactionTemplate {
+        inputs: &[(2, 1, 0)],
+        witness: inscription("text/plain", "hello").to_witness(),
+        ..Default::default()
+      });
+
+      let inscription_id = InscriptionId::from(txid);
+
+      context.mine_blocks(1);
+
+      context.index.assert_inscription_location(
+        inscription_id,
+        SatPoint {
+          outpoint: unbound_outpoint(),
+          offset: 0,
+        },
+        None,
+      );
+
+      context.mine_blocks(1);
+
+      context.rpc_server.broadcast_tx(TransactionTemplate {
+        inputs: &[(4, 0, 0)],
+        fee: 50 * 100_000_000,
+        ..Default::default()
+      });
+
+      context.mine_blocks(1);
+
+      let txid = context.rpc_server.broadcast_tx(TransactionTemplate {
+        inputs: &[(5, 1, 0)],
+        witness: inscription("text/plain", "hello").to_witness(),
+        ..Default::default()
+      });
+
+      let inscription_id = InscriptionId::from(txid);
+
+      context.mine_blocks(1);
+
+      context.index.assert_inscription_location(
+        inscription_id,
+        SatPoint {
+          outpoint: unbound_outpoint(),
+          offset: 1,
+        },
+        None,
       );
     }
   }
@@ -1403,7 +1469,7 @@ mod tests {
           outpoint: OutPoint { txid, vout: 0 },
           offset: 0,
         },
-        50 * COIN_VALUE,
+        Some(50 * COIN_VALUE),
       );
 
       let send_txid = context.rpc_server.broadcast_tx(TransactionTemplate {
@@ -1422,7 +1488,7 @@ mod tests {
           },
           offset: 50 * COIN_VALUE,
         },
-        50 * COIN_VALUE,
+        Some(50 * COIN_VALUE),
       );
     }
   }
@@ -1465,7 +1531,7 @@ mod tests {
           },
           offset: 0,
         },
-        50 * COIN_VALUE,
+        Some(50 * COIN_VALUE),
       );
 
       context.index.assert_inscription_location(
@@ -1477,7 +1543,7 @@ mod tests {
           },
           offset: 50 * COIN_VALUE,
         },
-        100 * COIN_VALUE,
+        Some(100 * COIN_VALUE),
       );
     }
   }
@@ -1502,7 +1568,7 @@ mod tests {
           outpoint: OutPoint { txid, vout: 0 },
           offset: 0,
         },
-        50 * COIN_VALUE,
+        Some(50 * COIN_VALUE),
       );
 
       let send_txid = context.rpc_server.broadcast_tx(TransactionTemplate {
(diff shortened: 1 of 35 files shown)


src/index.rs
```diff
@@ -2279,4 +2280,340 @@ mod tests {
...
+        .push_slice(&[])
+        .push_slice(b"qix")
+        .push_opcode(opcodes::all::OP_ENDIF)
+        .into_script();
+
+      let witness = Witness::from_vec(vec![script.into_bytes(), Vec::new()]);
+
+      let txid = context.rpc_server.broadcast_tx(TransactionTemplate {
+        inputs: &[(1, 0, 0), (2, 0, 0), (3, 0, 0)],
+        witness,
```
veryordinally, line 2551: Witness is replicated for all inputs - either comment or make explicit by using vector of witnesses

src/index/updater/inscription_updater.rs
```diff
@@ -76,66 +92,118 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
...
-    {
-      let flotsam = Flotsam {
-        inscription_id: txid.into(),
-        offset: 0,
-        origin: Origin::New {
-          fee: input_value - tx.output.iter().map(|txout| txout.value).sum::<u64>(),
-        },
+      let offset = input_value;
+
+      // different ways to get the utxo set (input amount)
```
veryordinally, line 124: Let's call it a multi-level cache

veryordinally, line 124:
```suggestion
      //  multi-level cache to get to input amount from UTXO set
```

src/index/updater/inscription_updater.rs
```diff
@@ -76,66 +92,118 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
...
-          SatPoint {
-            outpoint: unbound_outpoint(),
-            offset: self.unbound_inscriptions,
+      // go through all inscriptions in this input
+      while let Some(inscription) = new_inscriptions.peek() {
+        if inscription.tx_in_index != u32::try_from(input_index).unwrap() {
+          break;
+        }
+
+        // ignore reinscriptions on already inscribed offset (sats)
```
veryordinally, line 147:
```suggestion
        // In this first part of the cursed inscriptions implementation we ignore reinscriptions. This will change once we implement reinscriptions.
```

src/index/updater/inscription_updater.rs
```diff
@@ -76,66 +92,118 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
...
-        inscriptions.push(flotsam);
+        });
+
+        new_inscriptions.next();
+        id_counter += 1;
       }
-    };
+    }
+
+    // still have to normalize over multiple inscriptions per tx and inscription size
```
veryordinally, line 177: We should at least keep the invariant that the sum of genesis fees corresponds to the fee paid for the transaction - i.e. divide genesis fee by number of inscriptions in input for multiple inscriptions per input. That's not yet a fair allocation, but at least total fees would add up correctly.

src/inscription.rs
```diff
@@ -11,8 +11,12 @@ use {
   std::{iter::Peekable, str},
 };
 
+const INSCRIPTION_ENVELOPE: [bitcoin::blockdata::script::Instruction<'static>; 3] = [
```
veryordinally, line 14:
```suggestion
const INSCRIPTION_ENVELOPE_HEADER: [bitcoin::blockdata::script::Instruction<'static>; 3] = [
```

veryordinally: Technically just the header of the envelope

src/subcommand/server.rs
```diff
@@ -1637,17 +1629,17 @@ mod tests {
...
   <dd class=monospace>{inscription_id}</dd>
   <dt>preview</dt>.*<dt>output</dt>
-  <dd><a class=monospace href=/output/0000000000000000000000000000000000000000000000000000000000000000:0>0000000000000000000000000000000000000000000000000000000000000000:0</a></dd>.*"
+  <dd><a class=monospace href=/output/0000000000000000000000000000000000000000000000000000000000000000:0>0000000000000000000000000000000000000000000000000000000000000000:0 \\(unbound\\)</a></dd>.*"
       ),
     );
   }
 
   #[test]
   fn unknown_output_returns_404() {
```
veryordinally, line 1638:
```suggestion
  fn unbound_output_returns_404() {
```

veryordinally: Maybe delete, no more unknown outputs.

veryordinally (changes requested):

BatBushRacks:
@raphjaph @veryordinally first off thank you for being open source devs... irregardless of no open discussion... any update on https://github.com/ordinals/ord/pull/2109... why is it being straight up disregarded?

veryordinally:
> @raphjaph @veryordinally first off thank you for being open source devs... irregardless of no open discussion... any update on #2109... why is it being straight up disregarded?

We're working through the many different types of cursed inscriptions step by step, and are starting with what looks like the most-straightforward cases. Please be a little patient. Current plan is to get the uncontroversial cases of cursed inscriptions out first, so that we can make forward progress, and then tackle the other cases from #2045 like cursed inscriptions with unrecognized even tags (including OP66 etc), and other cases (like inscriptions on P2WSH) as in case #2139 after that.

BennyTheDev:
what are reinscriptions?

templates/inscription.html
raphjaph, line 1: Add (unstable) after negative inscription numbers

raphjaph:
> reinscriptions

inscribing a sat multiple times. Still figuring out if we should allow that and how exactly it should work.

BatBushRacks:
> > @raphjaph @veryordinally first off thank you for being open source devs... irregardless of no open discussion... any update on #2109... why is it being straight up disregarded?
> 
> We're working through the many different types of cursed inscriptions step by step, and are starting with what looks like the most-straightforward cases. Please be a little patient. Current plan is to get the uncontroversial cases of cursed inscriptions out first, so that we can make forward progress, and then tackle the other cases from #2045 like cursed inscriptions with unrecognized even tags (including OP66 etc), and other cases (like inscriptions on P2WSH) as in case #2139 after that.

@veryordinally thank you so much for responding... that all sounds great... patience is possible when we all have honest discourse... though what seems to be missing is any mention of the bound or unbound treatment of https://github.com/ordinals/ord/pull/2109... can we get a commitment that they will be bound?

BennyTheDev:
> > reinscriptions
> 
> inscribing a sat multiple times. Still figuring out if we should allow that and how exactly it should work.

Thank you. 

In regards to, all but but the first inscriptions being unbound: you mean multiple inscriptions within _one_ input, not inscriptions across multiple inputs, correct?

DanteScott:
> > reinscriptions
> 
> inscribing a sat multiple times. Still figuring out if we should allow that and how exactly it should work.

I think allowing reinscriptions would be useful in making rare sat / sub 100k inscriptions “recycleable”. Assuming the inscription number remains after reinscription, we could see new sub 100k collections emerge and rare sat inscriptions repurposed.

veryordinally (approved):

BatBushRacks:
> > > @raphjaph @veryordinally first off thank you for being open source devs... irregardless of no open discussion... any update on #2109... why is it being straight up disregarded?
> > 
> > 
> > We're working through the many different types of cursed inscriptions step by step, and are starting with what looks like the most-straightforward cases. Please be a little patient. Current plan is to get the uncontroversial cases of cursed inscriptions out first, so that we can make forward progress, and then tackle the other cases from #2045 like cursed inscriptions with unrecognized even tags (including OP66 etc), and other cases (like inscriptions on P2WSH) as in case #2139 after that.
> 
> @veryordinally thank you so much for responding... that all sounds great... patience is possible when we all have honest discourse... though what seems to be missing is any mention of the bound or unbound treatment of #2109... can we get a commitment that they will be bound?

Conveniently speedily merged without comment on https://github.com/ordinals/ord/pull/2109... seems like a pattern... avoid avoid avoid...

BennyTheDev:
I think there is an edge case not covered:

if there is just one inscription in a tx but not in input 1, it is cursed but treated like a "collection" and the new satpoint calculations are applied, which leads them to be unbound under certain circumstances. Imho they should generally get on offset = 0 and be bound.

You can cross-test this with ord 0.5.1 and inscribe for example from input index 1. It will treat it as it should, while ord 0.6.0 completely differs.

jotapea:
> Recognizes multiple inscriptions per transaction:
> 
>    1. in different inputs
>    2. in the same input

I believe the most sound way to move forward is to only count the FIRST inscription per input. Or even simpler, only consider the inscription valid if it has ONE inscription per input.

Seems like over-complicating the protocol when there is no real need. Is not like multi-file inscriptions are not possible now, because the data envelope can itself be a multi-part file (like a zip file or just multiple files directly encoded in a json).

And then there is the consistency of the assigned sats... it is supposed to be assigned as the first sat of the output, correct? How are the sats determined when an inscription has multiple inputs, and multiple inscriptions per input, and a single output?

cvasqxz:
> Sorry for coming off in a harsh way, but over-complicating the protocol is not the best design choice.

There's an economical incentive to over-complicate the protocol, people will try to break the inscriptions parsing to say they have a rare inscription and sell it at a premium

This is why bitcoin has the BIP system, to prevent people adding stuff to the code for personal profits or benefits

BatBushRacks:
> I think there is an edge case not covered:
> 
> if there is just one inscription in a tx but not in input 0, it is cursed but treated like a "collection" and the new satpoint calculations are applied, which leads them to be unbound under certain circumstances. Imho they should generally get offset = 0 and be bound.
> 
> You can cross-test this with ord 0.5.1, modify it to inscribe for example from input index 1. It will treat it as it should, while ord 0.6.0 completely differs.

@BennyTheDev not a dev so not certain... but looks like a fix is being made for the bug you found... https://github.com/ordinals/ord/pull/2154... sure would be nice if the devs understood that open source includes open discourse... hopefully with time...

concernedordinal:
> > reinscriptions
> 
> inscribing a sat multiple times. Still figuring out if we should allow that and how exactly it should work.

Hey @raphjaph this directly contradicts what you wrote in the first message. That said "All but the first inscription in the first input are cursed (for now). In addition all but the first inscription in an input are made unbound (they will be bound in the future as reinscriptions)." Can you clarify things? Super confusing as it currently stands.

creamplanet:
> > > Sorry for coming off in a harsh way, but over-complicating the protocol is not the best design choice.
> > 
> > 
> > There's an economical incentive to over-complicate the protocol, people will try to break the inscriptions parsing to say they have a rare inscription and sell it at a premium
> > This is why bitcoin has the BIP system, to prevent people adding stuff to the code for personal profits or benefits
> 
> This is a very important and under-appreciated risk. The long term health and longevity of the Ordinals protocol is the most important thing for the ecosystem. Once you add something in, it is hard to take it out.
> 
> If Ordinals as a protocol becomes too bloated or unmaintainable, people will move to a new standard. (this sort of standards obsolescence happens quite often in tech)
> 
> If people move on to a new standard of course, no inscriptions will hold any value and the entire ecosystem will disappear.
> 
> I encourage everyone to focus on the big picture let the team make the best decision they can.

My understanding was recognizing both "1. in different inputs" and "2. in the same input" is bloated and unnecessary. Seems to be in line with your thinking. Maybe I don't understand what you mean. Contradiction on reinscription is concerning from this angle too.

PR #2148, opened 2023-06-04
Tweak docs

raphjaph:


veryordinally (approved):
This change is backwards compatible with all all previously-recognized inscriptions, and provides clarity for future extensions (for example, inscriptions that are not on the first input).

huuep:
This is a great PR. It clearly defines the simple and elegant Ordinals Protocol, one that will last for a long time. The Ordinals Protocol can be simply defined in 3 rules:

1) Ordinal Theory - how individual satoshis can be uniquely identified and tracked across transactions using FIFO.
2) Inscription - data written to the input of a transaction (Inscriptions are not ownable yet.)
3) Define the Ordinal onto which an Inscription is inscribed: the first satoshi in the input of the Inscription. This is a simple definition and the first satoshi in the input is the closest point to the Inscription on the blockchain. (Now inscriptions are ownable)

PR #2155, opened 2023-06-05
Updated install.sh for new repo name

LightRider5:
The install.sh was still using casey/ord repo.


raphjaph (approved):
Thanks, for catching this!

PR #2161, opened 2023-06-06
allow ajax requests from 'self' same origin within inscriptions

ericatallah:
Relating to discussion here:
https://github.com/ordinals/ord/issues/1082

Allow browser ajax requests to same origin `/content/` path for flexibility and composability in the ord protocol.


casey:
This would allow inscriptions to request arbitrary content from the server, which is not guaranteed to be stable, and thus introduce mutability, which we don't want.

huuep:
good point, we need to limit it to /content/ which are liimted to on-chain inscription data

PR #2192, opened 2023-06-18
add indexing for p2wsh inscriptions

bdemann:
Note: This does not index them as cursed inscriptions
Update: It indexes them as cursed inscriptions now!

### DIFF
--- a/src/index/updater.rs
+++ b/src/index/updater.rs
@@ -435,6 +435,7 @@ impl Updater {
       block.header.time,
       unbound_inscriptions,
       value_cache,
+      index
     )?;
 
     if self.index_sats {
--- a/src/index/updater/inscription_updater.rs
+++ b/src/index/updater/inscription_updater.rs
@@ -1,3 +1,5 @@
+use bitcoin::{util::address::WitnessVersion};
+
 use super::*;
 
 #[derive(Debug, Clone)]
@@ -36,6 +38,7 @@ pub(super) struct InscriptionUpdater<'a, 'db, 'tx> {
   timestamp: u32,
   pub(super) unbound_inscriptions: u64,
   value_cache: &'a mut HashMap<OutPoint, u64>,
+  index: &'a Index,
 }
 
 impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
@@ -52,6 +55,7 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
     timestamp: u32,
     unbound_inscriptions: u64,
     value_cache: &'a mut HashMap<OutPoint, u64>,
+    index: &'a Index,
   ) -> Result<Self> {
     let next_cursed_number = number_to_id
       .iter()?
@@ -83,6 +87,7 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
       timestamp,
       unbound_inscriptions,
       value_cache,
+      index,
     })
   }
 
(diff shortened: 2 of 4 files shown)


bodily11:
@veryordinally @raphjaph This is a first draft pull request that adds P2WSH as cursed inscriptions. This gets quite technical distinguishing between P2TR and P2WSH, so I figured a text explanation would be helpful here:

1. P2TR (assuming no annex) always grabs the second to last section of witness data. This doesn't work for P2WSH because the ord envelope for P2WSH has to always be on the last section of witness data.
2. To index these as inscriptions, we had a few options. Ultimately we decided on the simplest approach. The simplest approach is check for inscriptions using the existing P2TR logic, and if no inscriptions are found, then check the last section of witness data for inscriptions. If you find any, add them as inscriptions. These inscriptions will be P2WSH inscriptions.
3. Next we needed to add them as cursed. To do this we need to distinguish between P2TR inscriptions and P2WSH inscriptions. The easiest and most efficient way to do this is check the witness version of the scriptPubKey on the previous_output. So in the inscription updater we check the scriptPubKey witness version. If it is 1 that means it is taproot and continue on as normal. If it is not 1 (meaning it is 0) it is P2WSH, and it should be cursed.
4. Then we added a few tests. :)

There were a few decisions we made along the way to try to simplify this implementation. Would love to hear your thoughts as we are anxious to get this added as a cursed inscription category.

Let me know if you have any questions/suggestions on our implementation.

raphjaph:
What is the benefit of using P2WSH? They seem strictly worse from a data carrying capacity. I'm a bit reluctant to add any kind of script only because it is possible.

bodily11:
Hey @raphjaph great question.

P2WSH inscriptions are worse from a data carrying capacity (I believe 10KB is the limitation for witness data serialization in P2WSH, can't remember exactly), but otherwise P2WSH inscriptions function in the same way as P2TR transactions for inscriptions.

The goal here is not to add P2WSH just because it is possible, but to broaden the types of applications that can inscribe on Bitcoin.

I do a lot of building on ICP. ICP has threshold ECDSA (you can sign ECDSA transactions directly from an ICP smart contract), but doesn't have threshold Schnorr. This means I can't inscribe directly from an ICP smart contract. The Dfinity foundation is looking into adding threshold Schnorr at the protocol level, but that might not ever happen.

Let me give you two examples of applications that I want to build on ICP that I can't without recognized and blessed P2WSH inscriptions:

1. **Decentralized inscription service**. Using ICP, I could build a community controlled or DAO controlled inscription service that does not rely on a centralized server for inscribing. You would be able to verify the code running in the smart contract, prove nobody can alter that code, pay to inscribe in BTC, and know that the inscription will be created exactly as you specified. The smart contract in this case would receive Bitcoin to a Bitcoin address controlled by the smart contract (uses tECDSA on ICP), then create the transaction in the smart contract, sign it using tECDSA, and broadcast the transaction, all entirely within the smart contract itself.

2. **Fully decentralized BRC-20 wrapper**. Sending BRC-20 to a wrapper smart contract Bitcoin address is easy as it only requires a Bitcoin transaction. But from a wrapper smart contract, you have to inscribe a transfer to send the balance back out. So right now, only centralized BRC-20 wrappers are possible. If P2WSH were recognized and blessed, then you could inscribe directly from an ICP smart contract to send a BRC-20 balance back out in the event that someone wanted to unwrap back to Bitcoin. 

These two solutions would provide interesting and valuable functionality to the Ordinals ecosystem, but neither is currently possible without recognized and blessed P2WSH inscriptions.

So this is why we submitted the PR, in the hopes that we could get P2WSH inscriptions recognized as cursed inscriptions, and then hopefully added as a blessed inscription type relatively soon.

Happy to chat offline on this too if you want additional context or info.

casey:
Please see my comment [here](https://github.com/ordinals/ord/issues/2139#issuecomment-1701523234).

PR #2206, opened 2023-06-22
add Markdown media type

elocremarc:
Markdown is a desirable media type to add now that we are inscribing Libraries. 
This would be a good way to document code onchain. 
You could keep the documentation in the same UTXO as the library or create a child parent using the library inscription
Clients would probably need to create a recursive markdown parser to add links and images for instance.
`[link](<inscriptionId>)`
`Image`
People could even make clients like obsidian using the same recursion above.
This would also allow people to easily write a blog on inscriptions.


raphjaph (approved):
With some markdown renderer as recursive inscription this could work nicely.

PR #2208, opened 2023-06-22
Add export command for number to id table

raphjaph:


src/subcommand/index.rs
```diff
@@ -1,5 +1,43 @@
 use super::*;
 
+#[derive(Debug, Parser)]
+pub(crate) enum IndexSubcommand {
+  #[clap(about = "Write inscription number and id to a file")]
```
veryordinally, line 5:
```suggestion
  #[clap(about = "Write inscription numbers and ids to a tab-separated file")]
```

veryordinally (approved):

PR #2222, opened 2023-06-26
Shadow inscriptions

onchainguy-eth:
I wasn't sure if creating a PR right away was the best approach, but I figured it would allow me to present my idea as simply as possible.

**Problem**
Recursive inscriptions are great. However:

1. Every inscription that assembles HTML (or svg) traits from a collection (e.g. [this one](https://ordinals.com/content/181be382241e90692b2dff3d02f926e4b40a254fd62823e9181ff5a2280f0ebfi0)) has overhead.
3. There is no clean way to let a recursive inscription determine traits based on e.g. block height or other inputs. With recursive endpoints an inscription can fetch that info for itself. But then again, every inscription needs to have that logic.

**Solution**
The proposed concept of shadow inscriptions allows users to inscribe a shadow of an existing (recursive) inscription by using the `text/shadow` content type and the shadowed inscription ID as the content. Internally, ord will resolve the content of the shadow inscription with the shadowed inscription and pass some arguments in the query string.

**Potential positive side effects**
This may also reduce the number of internal HTTP requests because ord can pass info directly to the shadowed inscription.

### DIFF
--- a/src/media.rs
+++ b/src/media.rs
@@ -37,6 +37,7 @@ impl Media {
     ("text/html;charset=utf-8", Media::Iframe, &["html"]),
     ("text/javascript", Media::Text, &["js"]),
     ("text/plain;charset=utf-8", Media::Text, &["txt"]),
+    ("text/shadow", Media::Text, &["shadow"]),
     ("text/markdown;charset=utf-8", Media::Text, &["md"]),
     ("video/mp4", Media::Video, &["mp4"]),
     ("video/webm", Media::Video, &["webm"]),
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -768,6 +768,29 @@ impl Server {
       .get_inscription_by_id(inscription_id)?
       .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
 
+    if inscription.content_type() == Some("text/shadow") {
+      let shadowed_inscription_id_raw = String::from_utf8(inscription.body().unwrap().to_vec()).unwrap();
+      let shadowed_inscription_id_parsed: String = shadowed_inscription_id_raw.chars().filter(|c| !c.is_whitespace()).collect();
+
+      let entry = index
+        .get_inscription_entry(inscription_id)?
+        .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
+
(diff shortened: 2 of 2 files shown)


bodily11:
You should look at the way we handle this with BitGen.

https://bioniq-1.gitbook.io/bitgen-bitcoin-generative-ordinals-standard-alpha/

Each inscription fits nicely (and succinctly) in a <script> tag. You put all the collection logic in the collection.js. And you put all the renderer logic in the renderer.js. Works pretty well and is extremely flexible for tons of use cases.

Thoughts?

onchainguy-eth:
@bodily11 I like your approach. It definitely keeps overhead low and works with the current state of recursive inscriptions. However, the following would only work if you get the inscription ID inserted from ord right?

```
<script t="0,1,2" src="/content/{collection-js-inscription-id-here}"></script>
```

Otherwise you would need to parse it from `window.location.pathname` which would blow up the inscription again. In any case, your approach will always consume more bytes then just a plain text with the inscription ID.

Also my PR proposes to redirect to the referenced inscription passing multiple arguments via query string only if the content type is `text/shadow`. This avoids any find/replace logic on inscription content.

bodily11:
Yeah I get it. We use the Collection JS inscription to handle anything like that. You inscribe it once per collection and it includes all of the logic needed to render the inscription. This is where you would put block height logic, trait logic, etc. Of course there is a bit of overhead here in defining it, but not that much. The rendering itself is further abstracted into a Renderer JS to keep the Collection JS lightweight. And it is very fast. The Collection JS and Renderer JS (both components of BitGen) are small files that don't impact loading performance.

So I agree with you. If there can be a more compact way of referencing other inscriptions (text/shadow like you suggested, or even text/uri) then we could shave a few bytes off of these inscriptions. But for all practical purposes, we can essentially do it all now already right?

casey:
I don't think this provides much benefit, given that a compact generative inscription is already just:

```
<script src=/content/ID></script>
```

The referenced script can use the inscription ID, from window.location, as a seed to instantiate the inscription. This is an additional 31 bytes (`<script src=/content/></script>`) vs just inscribing the inscription ID, so adding shadow inscriptions doesn't seem like a big win.

PR #2223, opened 2023-06-27
Change Recursion to Composition

piesrtasty:
We should change the name of "recursive" inscriptions as referenced in the docs. Nothing about them is self-referential. A better and more appropriate name would be "composite" inscriptions, as they are made by composing multiple inscriptions (layers) together to create a single image. (I realize this is not the only use case. Regardless other ways folks are using them are still not recursive, e.g. importing JS libs from other inscriptions)


casey:
I think this is reasonable, bit given that the term recursion is already in common use, I think that ship has sailed. Plus recursion sounds cool!

PR #2224, opened 2023-06-28
Set db cache size

raphjaph:
Depends on https://github.com/ordinals/ord/pull/2141


src/options.rs
```diff
@@ -28,6 +28,12 @@ pub(crate) struct Options {
   pub(crate) cookie_file: Option<PathBuf>,
   #[clap(long, help = "Store index in <DATA_DIR>.")]
   pub(crate) data_dir: Option<PathBuf>,
+  #[clap(
+    long,
+    default_value = "8589934592",
```
veryordinally, line 33: Make depend on total available RAM. A good default might be a quarter (or an eighth) of available total RAM on machine. Can use sys-info crate and then call like this: sys_info::mem_info().unwrap().total

veryordinally (approved):
Looks good!

PR #2228, opened 2023-06-28
Ord dev deploy

raphjaph:
I hate bash scripting


deploy/deploy-ord-dev
```diff
@@ -0,0 +1,33 @@
...
+# This script deploys whatever is currently 
+# in working directory to our dev server
+DOMAIN="ordinals-dev.com"
+
+# Current local working state 
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+COMMIT=$(git rev-parse --short HEAD)
+DIR="ord-${BRANCH}-${COMMIT}"
+
+$EDITOR ./deploy/ord-dev.service
```
veryordinally, line 13: Why do we want to do this?

veryordinally: To set --index-sats etc?

veryordinally (approved):

deploy/deploy-ord-dev
raphjaph: jup

PR #2231, opened 2023-06-30
brotli except iframe

staccDOTsol:
A much better brotli commit, designed to compress / decompress any text.

A little more research is required.

Todo: fix remaining tests (aware of iframe).

done: backwards compatible with noncompressed text ordinals, and forwards everything that is Media::Text will be compressed( and rendered) including brc-20 or whatever the cool kids are doing now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -613,6 +613,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "brotlic"
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


casey:
There are some issues with this approach:

- Compression is not part of the content type of an HTTP response, it's expressed in the Content-Encoding header
- If an HTTP client doesn't specify that a `Content-Encoding` is a acceptable using the `Accept-Encoding` header, then the web server shouldn't return it.

So, the right approach is:

- Allow inscriptions to contain a new field, which is the `Content-Encoding` of the body
- When serving /content, check the `Accept-Encoding` header to see if the `Content-Encoding` of the inscription is acceptable. If it is, return the content as is, otherwise decode it.

PR #2236, opened 2023-07-01
Update server.rs to fix recursive inscriptions on iOS/Safari

bodily11:
Recursive inscriptions on ordinals.com are currently broken on iOS and Safari. Moving 'self' further down fixes this so recursive inscriptions can render on iOS and Safari on Ordinals.com.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -790,7 +790,7 @@ impl Server {
     );
     headers.insert(
       header::CONTENT_SECURITY_POLICY,
-      HeaderValue::from_static("default-src 'self' 'unsafe-eval' 'unsafe-inline' data: blob:"),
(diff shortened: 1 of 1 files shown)


bodily11:
I'm unsure on this, but the second set of CSPs might need to be adjusted as well (moving each whitelisted endpoint after the unsafe eval and unsafe inline params)

raphjaph:
Unfortunately I'm unable to make this work on Safari. How did you test this?

devlz303:
I think it is not working because you still use two CSP headers (insert and then append, which doesn't work for Safari). I've made a fix for this in this pull request: https://github.com/ordinals/ord/pull/2251

bodily11:
Yeah, I can confirm that #2251 fixes it. Great work @devlz303

raphjaph:
Ok, I will close this in favor of #2251. Will write there.

Issue #2232, opened 2023-06-30
[Bug] Easy way to break all new /content and /preview responses


revofusion:
If you request an inscription ID before its confirmed on the chain, ordinals.com caches the failed response, and does not return correct /content and /preview even after inscription is confirmed

Quick reproduction:
https://ordinals.com/preview/084e7e4ce6875c025bb25caafd508f91380dc86bc415978154123edaf5398a55i0
https://ordinals.com/content/084e7e4ce6875c025bb25caafd508f91380dc86bc415978154123edaf5398a55i0

Can see it exists here:
https://ordinals.com/inscription/084e7e4ce6875c025bb25caafd508f91380dc86bc415978154123edaf5398a55i0

Correct image on an ord mirror:
https://ord-mirror.magiceden.dev/content/084e7e4ce6875c025bb25caafd508f91380dc86bc415978154123edaf5398a55i0

Fix: Cache not found for something more reasonable like 5 minutes

Note: After 2 days, seems like cache finally expired on above, or was purged, and is finally viewable

raphjaph:
Closed by https://github.com/ordinals/ord/pull/2240.

casey:
PR #2240 didn't actually fix this, so re-opening. #2240 set the cache control header not in the case that the inscription was not found, but in the case that the inscription had no body. So presumably this is still an issue.

casey:
This has been fxied.

PR #2240, opened 2023-07-01
Set lower max age for not found

revofusion:
Lower cache time to 600 (10 minutes) when an inscription is not found

Resolve issue #2232


veryordinally (approved):
@raphjaph this looks good to me

revofusion:
@veryordinally have made a slight change to not set `immutable` on not found

raphjaph:
Thanks, this is a very good catch!

PR #2250, opened 2023-07-05
Read only Json API

Mathieu-Be:
Opening a new PR based on the work of @ynohtna92 on https://github.com/ordinals/ord/pull/2036.

I'm not super experienced in Rust so I'm opening the PR to get early feedback, only the `sat` endpoint is implemented as of now.

I'm wondering if it is really necessary to have the exact same fields as the HTML response, is the `_links` section really necessary for example ?

Also, I feel like there could be a better way to build the Json response, but if I change the `Serialize` impl of `Sat` it is breaking other stuff, so not sure what to do here (maybe create a `SatApiResponse` struct based on `Sat` ?)


ynohtna92:
_links aka HATEOAS make the API crawlable, it is better to have them then not have them.

raphjaph:
@Mathieu-Be just implemented the small changes I requested. Have a look if you agree with them.

Mathieu-Be:
> @Mathieu-Be just implemented the small changes I requested. Have a look if you agree with them.

Cool ! I haven't seen the changes you requested, did they get lost somewhere ?

Want me to take a stab at the other endpoints ?

raphjaph:
> Cool ! I haven't seen the changes you requested, did they get lost somewhere ? The Json struct looks nicer indeed.

I reviewed them above. Maybe they are collapsed (grey box)? 

> Want me to take a stab at the other endpoints ?

That would be great! But before you start I think we should make the JSON endpoint configurable in this PR. Such that you have to pass a flag to activate it. Something like `--with-json-api`

devords:
Before adding the `/sat/N` JSON endpoint, it might be worth getting some clarity on #2277 as it includes a proposal for the same path. While the server can return different responses based on the `Accept` header, it'd be nice to use different paths for returning JSON and inscriptions on the sat.

src/options.rs
```diff
@@ -54,6 +54,12 @@ pub(crate) struct Options {
   pub(crate) testnet: bool,
   #[clap(long, default_value = "ord", help = "Use wallet named <WALLET>.")]
   pub(crate) wallet: String,
+  #[clap(
+    long,
+    short,
+    help = "Enable JSON API to get JSON responses from the server instead of HTML."
```
raphjaph, line 60:
```suggestion
    help = "Enable JSON API."
```

raphjaph (approved):
Let's go!

Issue #2180, opened 2023-06-14
Ord instance frozen after restart


victorkirov:
I have an instance of Ord running with a bitcoin full node on the side. Everything runs smoothly until there is a power outage or Ord is shutdown for some reason. When I start it up again, I try running `ord index` and it just hangs. Using RUST_LOG=trace, I get the following logs and then it just hangs:
```
[2023-06-14T11:01:54Z INFO  ord::options] Connecting to Bitcoin Core at 127.0.0.1:8332/wallet/ord
[2023-06-14T11:01:54Z INFO  ord::options] Using credentials from cookie file at `/root/.bitcoin/.cookie`
[2023-06-14T11:01:54Z DEBUG bitcoincore_rpc] JSON-RPC request: getblockchaininfo []
[2023-06-14T11:01:54Z TRACE bitcoincore_rpc] JSON-RPC response for getblockchaininfo: {"chain":"main","blocks":794301,"headers":794301,"bestblockhash":"00000000000000000000e2f65871279634a11d953bb3e696a9db356eb7e6fc22","difficulty":51234338863442.89,"time":1686739915,"mediantime":1686736332,"verificationprogress":0.9999974173535334,"initialblockdownload":false,"chainwork":"00000000000000000000000000000000000000004ba7629d0f4cf511bf7160ac","size_on_disk":554209415432,"pruned":false,"warnings":""}
[2023-06-14T11:01:54Z DEBUG bitcoincore_rpc] JSON-RPC request: getnetworkinfo []
[...]

Issue #1861, opened 2023-03-01
Running 'ord server' seems to leave the index in bad state, delays next ord launch


so7ow:
Running 'ord server' works as expected. Ctrl-C to exit also seems to behave as expected. However, each time after running 'ord server' and then exiting, the NEXT time I run ord I get a long delay (up to 20-30 min for a with-sats index) before it picks up with indexing and does whatever new command I've given it.

macOS 13.2.1 / M1 Max / 32 GB RAM / SSD
ord 0.5.1

gmart7t2:
I see exactly the same thing. It takes about 25 minutes before the "listening on" text shows up. If I hit control-C at that point and rerun "ord server" it takes another 25 minutes. While waiting it is constantly reading about 22 MB/s from the drive that holds the index file. That means it's reading around 33 GB of the 116 GB index file.

victorkirov:
Same here. With the logs on TRACE, there is literally nothing logged during that period either :/

nick07002:
Exact same issue I am facing..

raphjaph:
Fixed by #2275

PR #2270, opened 2023-07-13
fix: server corrupting index

victorkirov:
This PR fixes the issue where the index would be corrupted after running and shutting down the web server.

The issue occured from the update thread not being joined on shutdown and, hence, not gracefully closing its connection to the index.

2 things have been added in this PR: 
- termination feature for ctrlc crate. This allows graceful shutting down on a SIGTERM signal which docker and kubernetes use to shut down processes. Currently, running ord in a docker container completely bypasses shutting down gracefully without this.
- an UPDATE_THREAD mutex which holds the update thread which the server creates and joins onto it when shutting down gracefully. This allows the update thread to go through the graceful shutdown procedure of committing any uncommitted entries to the index and closing the connection to it. 

This has been tested with the following steps:
- Run ord in server mode
- curl the blockheight endpoint until the server is up and we get the blockheight returned
- Send SIGTERM/SIGINT signal
- Run the server again
- curl the blockheight endpoint
At this point, before the changes, the server would not accept connections and would take a while to start up while it fixes the index. After these proposed changes, it starts up straight away.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ bitcoin = { version = "0.29.1", features = ["rand"] }
 boilerplate = { version = "0.2.3", features = ["axum"] }
 chrono = "0.4.19"
 clap = { version = "3.2.18", features = ["derive", "deprecated"] }
-ctrlc = "3.2.1"
+ctrlc = { version = "3.2.1", features = ["termination"] }
 derive_more = "0.99.17"
 dirs = "5.0.0"
 env_logger = "0.10.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,6 +131,7 @@ const CYCLE_EPOCHS: u64 = 6;
 
 static SHUTTING_DOWN: AtomicBool = AtomicBool::new(false);
 static LISTENERS: Mutex<Vec<axum_server::Handle>> = Mutex::new(Vec::new());
+static UPDATE_THREAD: Mutex<Option<thread::JoinHandle<()>>> = Mutex::new(Option::None);
 
 fn integration_test() -> bool {
   env::var_os("ORD_INTEGRATION_TEST")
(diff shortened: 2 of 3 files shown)


gmart7t2:
Very nice. It fixes the issue for me. 

One problem I see is that if I run "ord server" as non-root I see

```
Listening on http://0.0.0.0:80
error: Permission denied (os error 13)
```

as expected (because non-root processes can't listen on low numbered ports), but after that the index is corrupted.

bodily11:
This fixes some persistent issues I was having using AWS + EBS. Even if I tried to kill gracefully (kill -INT %1) it would still kill forcefully and would have to repair the index, resulting in 30 min start times for me on an index-sats index after killing the server (even if killing gracefully). But this fixes it! This is AMAZING.

gmart7t2:
> I think the last commit should fix that up, along with another issue where a second ctrlc wouldn't be picked up to force kill

It doesn't appear to fix the issue. 

There's a `process::exit(1)` before the `update_thread.join()` when the main thread returns an error

gmart7t2:
The issue this is fixing is #1861 btw.

victorkirov:
> > I think the last commit should fix that up, along with another issue where a second ctrlc wouldn't be picked up to force kill
> 
> 
> 
> It doesn't appear to fix the issue. 
> 
> 
> 
> There's a `process::exit(1)` before the `update_thread.join()` when the main thread returns an error

Ah, oops. You're absolutely right. Will fix it up in a little.

victorkirov:
Ok, that should do it

victorkirov:
This will also resolve #2180 and #2142

gmart7t2:
> Ok, that should do it

So now when I run "ord server" I see the same error message because I don't have permission to listen on port 80, but then I don't get my shell prompt back. Ord keeps running, updating the index each time a block is mined. I can hit control-C once and it exits gracefully without corrupting the index file. But I shouldn't have to hit control-C when a fatal error has already happened.

gmart7t2:
How about this? This appears to fix the issue for me:

```
diff --git a/src/lib.rs b/src/lib.rs
index 020d2bc..4081d0b 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,7 +192,9 @@ pub fn main() {
       eprintln!("{}", err.backtrace());
     }
 
-    gracefully_shutdown_update_thread();
+    if !SHUTTING_DOWN.fetch_or(true, atomic::Ordering::Relaxed) {
+      gracefully_shutdown_update_thread();
+    }
 
     process::exit(1);
   }
```

victorkirov:
Yeah, there's nothing to notify that thread to break out of the loop 🙈 I think the proposed change will work in the case of the server erroring during normal operation but it wouldn't run the graceful shutdown in the case that the server errors after a shutdown is already executed.

I'll add the setting of the SHUTTING_DOWN value in the function itself to explicitly have it correct before we join.

raphjaph:
Thanks, for the PR. Looks good so far!

I just wanted to rename some variables and make the CI pass but I'm unable to push my changes. It's weird because github shows that maintainers are allowed to push changes to this branch. Opened a PR with the changes in #2275  if you could merge them in. I'm going to merge & squash anyways so don't worry about a merge commit.

veryordinally:
> Thanks, for the PR. Looks good so far!
> 
> I just wanted to rename some variables and make the CI pass but I'm unable to push my changes. It's weird because github shows that maintainers are allowed to push changes to this branch. Opened a PR with the changes in #2275 if you could merge them in. I'm going to squash-merge anyways so don't worry about a merge commit.

Reviewed and merged #2275, don't see a need to update this one. Thanks for this fix @victorkirov and the testing @gmart7t2 !

PR #2279, opened 2023-07-16
Fix ordering for reinscriptions and show all reinscriptions for sat

veryordinally:
This PR fixes a bug in the assignment of reinscription sequence numbers and allows to correctly order reinscriptions in chronological order. It also implements showing all inscriptions for a given sat (and not just a random one) in the correct sequence.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -3129,8 +3129,6 @@ mod tests {
           ..Default::default()
         });
 
-        dbg!(i, txid);
-
         let inscription_id = InscriptionId { txid, index: 0 };
         inscription_ids.push(inscription_id);
 
--- a/src/index.rs
+++ b/src/index.rs
@@ -536,17 +536,30 @@ impl Index {
     self.client.get_block(&hash).into_option()
   }
 
-  pub(crate) fn get_inscription_id_by_sat(&self, sat: Sat) -> Result<Option<InscriptionId>> {
-    Ok(
-      self
-        .database
-        .begin_read()?
-        .open_multimap_table(SAT_TO_INSCRIPTION_ID)?
-        .get(&sat.n())?
-        .next()
-        .and_then(|result| result.ok())
-        .map(|inscription_id| Entry::load(*inscription_id.value())),
-    )
+  pub(crate) fn get_inscription_ids_by_sat(&self, sat: Sat) -> Result<Vec<InscriptionId>> {
+    let rtx = &self.database.begin_read()?;
+
+    let mut ids = rtx
+      .open_multimap_table(SAT_TO_INSCRIPTION_ID)?
+      .get(&sat.n())?
+      .filter_map(|result| {
+        result
+          .ok()
+          .map(|inscription_id| InscriptionId::load(*inscription_id.value()))
+      })
+      .collect::<Vec<InscriptionId>>();
+
+    if ids.len() > 1 {
+      let re_id_to_seq_num = rtx.open_table(REINSCRIPTION_ID_TO_SEQUENCE_NUMBER)?;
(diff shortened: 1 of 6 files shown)


veryordinally:
See https://github.com/raphjaph/ord/commit/98340591c2981bad6d3215b185e1c9bc7c3e0f66 for where bug was introduced

src/subcommand/wallet/inscriptions.rs
```diff
@@ -11,6 +11,8 @@ pub(crate) fn run(options: Options) -> Result {
   let index = Index::open(&options)?;
   index.update()?;
 
+  // BACKLOG - change the logic here to only get inscriptions for outputs actually in the wallet
```
raphjaph, line 14: Make issue

templates/sat.html
```diff
@@ -11,8 +11,13 @@ <h1>Sat {{ self.sat.n() }}</h1>
   <dt>offset</dt><dd>{{ self.sat.third() }}</dd>
   <dt>rarity</dt><dd><span class={{self.sat.rarity()}}>{{ self.sat.rarity() }}</span></dd>
   <dt>timestamp</dt><dd><time>{{self.blocktime.timestamp()}}</time>{{self.blocktime.suffix()}}</dd>
-%% if let Some((inscription)) = &self.inscription {
-  <dt>inscription</dt><dd class=thumbnails>{{ Iframe::thumbnail(*inscription) }}</dd>
+%% if !self.inscriptions.is_empty() {
```
raphjaph, line 14: Did you visually confirm that this looks good?

src/index.rs
```diff
@@ -536,17 +536,30 @@ impl Index {
...
+          .ok()
+          .map(|inscription_id| InscriptionId::load(*inscription_id.value()))
+      })
+      .collect::<Vec<InscriptionId>>();
+
+    if ids.len() > 1 {
+      let re_id_to_seq_num = rtx.open_table(REINSCRIPTION_ID_TO_SEQUENCE_NUMBER)?;
+      ids.sort_by_key(
+        |inscription_id| match re_id_to_seq_num.get(&inscription_id.store()) {
+          Ok(Some(num)) => num.value() + 1,
```
raphjaph, line 556: Why the +1?

src/index.rs
```diff
@@ -989,7 +995,7 @@ impl Index {
 
     result.sort_by_key(|(_satpoint, inscription_id)| {
       match re_id_to_seq_num.get(&inscription_id.store()) {
-        Ok(Some(num)) => num.value(),
+        Ok(Some(num)) => num.value() + 1,
```
raphjaph, line 998: Why +1?

templates/sat.html
veryordinally: Yes. It's the same layout we use for showing all inscriptions in an output.

src/index.rs
veryordinally: We want the first inscription to have the smallest value in the ordering for this comparison, so we use 0 when the inscription is not in the table. That means first reinscription should get 1 to be greater than 0. Sequence numbers in the table start with 0 (that was part of the bug introduced ) and I wanted to fix this in a way that does not require an index rebuild.

src/index.rs
veryordinally: See previous comment

src/subcommand/wallet/inscriptions.rs
veryordinally: Done: #2282

raphjaph (approved):
Added some comments to remove the +1 the next time we refactor the database.

PR #2280, opened 2023-07-16
use backported miniscript branch from apoelstra (should hit mainline …

niftynei:
Temporary patch to fix rust build errors for miniscript; this should be updated when a backport to `9.0.x` is released

Tho note that `10.0.0` of miniscript is already released w/ a fix; should probs upgrade to that instead?


veryordinally (approved):

raphjaph:
I've started to update the dependencies in #2281 because I think we can't make a proper release if we have this sort of override in the `Cargo.toml`. A lot of things are breaking so that might take a while.

raphjaph:
Updated the dependency #2281

Issue #2248, opened 2023-07-04
 Couldn't receive txs expected value at line 1 column 1


h10086733:
When ord executes index run, Couldn't receive txs expected value at line 1 column 1 often occurs, which often occurs when the full amount of data is synchronized. In the mode of index-sats index run, the amount of data has almost reached 100G, and every It takes a lot of time to load index.rdb into memory for the first time, so more retries are needed to reduce reloading when exception handling occurs. And every time the database is upgraded, index.rdb needs to be reloaded. Whether to consider splitting Optimizing processing in the form of historical rdb files.

victorkirov:
Same here. Mine happened on block 790593 on a full index with SATS.

raphjaph:
That error occurs when fetching transactions but don't think it impacts performance in a large way.

>  And every time the database is upgraded, index.rdb needs to be reloaded. Whether to consider splitting Optimizing processing in the form of historical rdb files.

Every time the database schema changes we have to rebuild the database. Would be great to implement database migration but that could be quite involved. 

Generally there is a lot of optimizations that can still be done for indexing. Will try to enumerate in the coming weeks so people can take a stab at it.

victorkirov:
I traced the error to the fetcher. I didn't put in a log, so not 100% sure what bitcoin core is returning, but when transactions are fetched from the Bitcoin core, it seems to return some sort of error string instead of json. When ord tries to deserialize the response, it returns the "expected value at line 1 column 1" error.

The Bitcoin node seems to be overloaded maybe because the Electrs api running off the same node has issues around the same time. 

I wrapped the transaction fetch logic in a retry loop and the error seemed to stop 🎉

nick07002:
>I wrapped the transaction fetch logic in a retry loop and the error seemed to stop 🎉
that's really nice @victorkirov  mind sharing the location where you made the change? I probably need to do the same.
appreciate it.

victorkirov:
Sorry, I'm on my phone 🙈 It's line 79 here: https://github.com/ordinals/ord/blob/master/src/index/fetcher.rs

The response is most likely an error string instead of json. 

I stuck lines 68 to 79 in a method and called in in a retry loop that retries 3 times.

If you do it, please add a log of the response body so we can see what it is 🙏

I can post the code a bit later today if you want.

PR #2297, opened 2023-07-22
fix: add retry to fetcher

victorkirov:
Resolves: #2248 

Sometimes the transaction fetcher receives a non-json response from bitcoin core, resulting in a parsing error. It seems to happen due to load on the node. This is problematic because it will completely fail a batch of processed blocks. With the batch size set to 5000, if it fails a few thousand blocks in, it will be a lot of work to start over again.

This adds a retry which tries to fetch transactions with 5 retries which exponentially back off (Max being 3.2seconds).

### DIFF
--- a/src/index/fetcher.rs
+++ b/src/index/fetcher.rs
@@ -67,20 +67,27 @@ impl Fetcher {
     let body = Value::Array(reqs).to_string();
 
     let mut results: Vec<JsonResponse<String>>;
-
     let mut tries = 0;
 
     loop {
-      if tries >= 3 {
-        return Err(anyhow!("Failed to fetch raw transaction after 3 tries"));
-      }
-
       results = match self.try_get_transactions(body.clone()).await {
         Ok(results) => results,
         Err(error) => {
-          log::warn!("Failed to fetch raw transaction, retrying: {}", error);
+          if tries >= 5 {
+            return Err(anyhow!(
+              "Failed to fetch raw transactions after 5 retries. Error: {}",
+              error
+            ));
+          }
+
+          log::info!(
+            "Failed to fetch raw transactions, retrying. Error: {}",
+            error
+          );
           tries += 1;
-          tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+
+          let sleep_time = tokio::time::Duration::from_millis(100 * u64::pow(tries, 2));
+          tokio::time::sleep(sleep_time).await;
           continue;
         }
       };
@@ -131,7 +138,16 @@ impl Fetcher {
 
     let buf = hyper::body::to_bytes(response).await?;
 
-    let results: Vec<JsonResponse<String>> = serde_json::from_slice(&buf)?;
+    let results: Vec<JsonResponse<String>> = match serde_json::from_slice(&buf) {
+      Ok(results) => results,
+      Err(e) => {
+        return Err(anyhow!(
+          "Failed to parse JSON-RPC response: {e}. Response: {response}",
+          e = e,
+          response = String::from_utf8_lossy(&buf)
+        ))
+      }
+    };
 
     Ok(results)
   }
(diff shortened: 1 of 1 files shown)


victorkirov:
If this should go in, maybe it should be a back off retry and maybe a few more times (5 total maybe?). Let me know.

src/index/fetcher.rs
```diff
@@ -65,18 +65,27 @@ impl Fetcher {
...
+      if tries >= 3 {
+        return Err(anyhow!("Failed to fetch raw transaction after 3 tries"));
+      }
+
+      results = match self.try_get_transactions(body.clone()).await {
+        Ok(results) => results,
+        Err(error) => {
+          log::warn!("Failed to fetch raw transaction, retrying: {}", error);
+          tries += 1;
+          tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
nick07002, line 83: The wait time of 100ms seems small.
maybe could use backoff retry.

nick07002 (approved):
Great stuff! This is going to help a lot of people overcome their indexing hiccups

victorkirov:
Great. I'll fix up the logging and make it a back off retry a bit later today when I'm at my pc.

src/index/fetcher.rs
```diff
@@ -65,18 +65,34 @@ impl Fetcher {
...
+    let mut results: Vec<JsonResponse<String>>;
+    let mut tries = 0;
+
+    loop {
+      results = match self.try_get_transactions(body.clone()).await {
+        Ok(results) => results,
+        Err(error) => {
+          if tries >= 5 {
+            return Err(anyhow!(
+              "Failed to fetch raw transactions after 5 retries. Error: {}",
```
raphjaph, line 78:
```suggestion
              "failed to fetch raw transactions after 5 retries: {}",
```
We're not very consistent with not capitalizing error message ourselves but I want to try to do that. So if you could make all the error messages lowercase that would be great. Also I don't think we need the "Error:" just colon or semicolon and then the "{}". Thanks!

src/index/fetcher.rs
```diff
@@ -109,4 +125,30 @@ impl Fetcher {
...
+
+    let response = self.client.request(req).await?;
+
+    let buf = hyper::body::to_bytes(response).await?;
+
+    let results: Vec<JsonResponse<String>> = match serde_json::from_slice(&buf) {
+      Ok(results) => results,
+      Err(e) => {
+        return Err(anyhow!(
+          "Failed to parse JSON-RPC response: {e}. Response: {response}",
```
raphjaph, line 145: same as above

src/index/fetcher.rs
```diff
@@ -65,18 +65,34 @@ impl Fetcher {
...
+        Err(error) => {
+          if tries >= 5 {
+            return Err(anyhow!(
+              "Failed to fetch raw transactions after 5 retries. Error: {}",
+              error
+            ));
+          }
+
+          log::info!(
+            "Failed to fetch raw transactions, retrying. Error: {}",
```
raphjaph, line 84: same as above

raphjaph (changes requested):
Thanks, the logic looks good so far! Just some minor tweak needed.

src/index/fetcher.rs
```diff
@@ -65,18 +65,34 @@ impl Fetcher {
...
+            error
+          );
+          tries += 1;
+
+          let sleep_time = tokio::time::Duration::from_millis(100 * u64::pow(tries, 2));
```
raphjaph, line 89:
```suggestion
          let sleep_time = tokio::time::Duration::from_millis(100 * u64::pow(2, retries));
```
I think we want this for exponential backoff.

src/index/fetcher.rs
```diff
@@ -65,18 +65,34 @@ impl Fetcher {
...
+              "Failed to fetch raw transactions after 5 retries. Error: {}",
+              error
+            ));
+          }
+
+          log::info!(
+            "Failed to fetch raw transactions, retrying. Error: {}",
+            error
+          );
+          tries += 1;
```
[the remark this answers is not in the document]
victorkirov: Ah, good catch 😄

victorkirov:
Thanks for the review @raphjaph. I think it should all be sorted 🙂

raphjaph (approved):

PR #1858, opened 2023-02-28
Allow selection of multiple utxos in pad_alignment_output and add_value 

gmart7t2:
`ord wallet inscribe` only adds at most 1 padding alignment utxo and 1 value utxo.

If you try to inscribe a tiny file with 5 cardinal outputs of size 4k sats each in your wallet the inscribe will fail because ord isn't able to use enough inputs to make up the 10k postage.

This commit fixes that.

An example:

We have five 4k outputs:

    $ ord wallet outputs | jq -cM .[]
    {"output":"adea31b810f76ff7ff92332e1260c422b9326cf66f45c24451e613f01fff5009:1","amount":4000}
    {"output":"eb9f7b9427062d672118c2882df235c489208a314ed8a7f45f8b0709bc226514:1","amount":4000}
    {"output":"e1e2b82a40f33ee968c5ffbda541dc7a9d67b916da717e36065b7a92f5d0965e:0","amount":4000}
    {"output":"fb2e602809d77569623ae348f17bc593515003181bc3d81db97d0f9c0b0b2c79:0","amount":4000}
    {"output":"fc61bc0be498e348b76503bbbab06fc3e0bcc5bf799ebdec847693143f582baf:1","amount":4000}

ord can't inscribe onto any of them:

    $ ord wallet inscribe --fee-rate 1 /tmp/hello.txt
    error: wallet does not contain enough cardinal UTXOs, please add additional funds to wallet.

But with this pull request it can:

    $ ord wallet inscribe --fee-rate 1 /tmp/hello.txt
    {
      "commit": "dd49221d3ad3b02c0a9671c014974191b43d2b265805641dda908e74d1fe06da",
      "inscription": "5e1807eaca2267c6273bca0ed13b5ad280efcd474def7b8ea189b8d2d2e121dbi0",
      "reveal": "5e1807eaca2267c6273bca0ed13b5ad280efcd474def7b8ea189b8d2d2e121db",
      "fees": 408
    }

Here are the commit tx inputs:

    $ bitcoin-cli getrawtransaction dd49221d3ad3b02c0a9671c014974191b43d2b265805641dda908e74d1fe06da 1 | jq -cM '.vin[]|{txid,vout}'
    {"txid":"adea31b810f76ff7ff92332e1260c422b9326cf66f45c24451e613f01fff5009","vout":1}
    {"txid":"eb9f7b9427062d672118c2882df235c489208a314ed8a7f45f8b0709bc226514","vout":1}
[...]

PR #2303, opened 2023-07-24
Select multiple utxos

raphjaph:
Duplicate of #1858 but have to make some small changes and cannot edit PR there.


raphjaph:
@gmart7t2 I added the `--max-inputs` to the fuzz tester and let it run for a day. Didn't produce any panics. I also made the `select_cardinal_utxo logic` more verbose so it is easier to understand. If my that change doesn't fit the logic you meant let me know. All tests seem to pass though.

gmart7t2:
I've never used --max-inputs

I added it because of a comment in the PR from someone who was concerned that selecting multiple utxos might select too many of them. 

Personally I don't mind if you remove the --max- inputs flag completely

raphjaph:
Ok I will remove it then

PR #2304, opened 2023-07-25
Check for and report db corruption before attempting repair

### LINKED ISSUE #2256: Inform User when redb starts in recovery mode
Opening the index.redb is very slow when it is in recovery mode - this can take 30 minutes or more when opening an index with full sats. We should log a message informing the user that the database is opened in recovery mode.

gmart7t2:
Fix #2256 using cberner's suggested quick hack.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -18,7 +18,7 @@ use {
     TableDefinition, WriteTransaction,
   },
   std::collections::HashMap,
-  std::io::{BufWriter, Write},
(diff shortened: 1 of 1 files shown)


raphjaph (approved):
Awesome, this was on my backlog so great to have implemented! Looks good to me. Maybe @veryordinally wants to have a look at this as well

veryordinally (approved):
Thank you!

PR #2310, opened 2023-07-31
Only fetch inscriptions that are owned by the ord wallet

gmart7t2:
This makes most ord commands run a few hundred times faster:

before:

```
$ time ord wallet balance > /dev/null
real	3m39.900s, user	0m23.681s, sys	0m20.880s
```

after:

```
$ time ord wallet balance > /dev/null
real	0m0.648s, user	0m0.025s, sys	0m0.199s
$ time ord wallet balance > /dev/null
real	0m0.477s, user	0m0.010s, sys	0m0.095s
$ time ord wallet balance > /dev/null
real	0m0.415s, user	0m0.021s, sys	0m0.067s
```

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -826,24 +826,15 @@ impl Index {
 
   pub(crate) fn get_inscriptions(
     &self,
-    n: Option<usize>,
+    utxos: BTreeMap<OutPoint, Amount>,
   ) -> Result<BTreeMap<SatPoint, InscriptionId>> {
     let rtx = self.database.begin_read()?;
 
     let mut result = BTreeMap::new();
 
-    for range_result in rtx
-      .open_multimap_table(SATPOINT_TO_INSCRIPTION_ID)?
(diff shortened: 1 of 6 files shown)


src/index.rs
```diff
@@ -826,24 +826,15 @@ impl Index {
...
-      for id_result in ids {
-        let id = id_result?;
-        result.insert(Entry::load(*satpoint.value()), Entry::load(*id.value()));
-      }
-      if result.len() >= n.unwrap_or(usize::MAX) {
-        break;
-      }
+    let table = rtx.open_multimap_table(SATPOINT_TO_INSCRIPTION_ID)?;
+    for utxo in utxos.keys() {
+      result.extend(Self::inscriptions_on_output_unordered(&table, *utxo)?);
```
raphjaph, line 837: Should we get unordered or ordered?

raphjaph (approved):
Great improvement! Going through 22million inscriptions probably wasn't the most efficient way.

PR #2317, opened 2023-08-07
Reorg resistance

raphjaph:


src/index/updater.rs
```diff
@@ -96,27 +98,46 @@ impl Updater {
...
-      self.index_block(
-        index,
-        &mut outpoint_sender,
-        &mut value_receiver,
-        &mut wtx,
-        block,
-        &mut value_cache,
-      )?;
+      if self.index.client.get_block_count()? - self.index.block_height()?.unwrap_or(Height(0)).n()
+        < 50
```
victorkirov, line 106: This should probably be the maximum number of savepoints (=6). Anything higher is unnecessary unless you want to have extra reorg resistance during the initial indexing 😋

victorkirov:
@raphjaph I implemented a very basic reorg protection with savepoints, so hope you don't mind the review comments 😋

src/index/updater.rs
```diff
@@ -96,27 +98,46 @@ impl Updater {
...
+        self.initial_sync = false;
+
+        if self.savepoints.len() >= 6 {
+          dbg!("here");
+          wtx.delete_persistent_savepoint(self.savepoints.pop_front().unwrap())?;
+        }
+
+        let savepoints: Vec<u64> = wtx.list_persistent_savepoints()?.into_iter().collect();
+        dbg!(&savepoints);
+        // self.savepoints.push_back(wtx.ephemeral_savepoint()?);
```
victorkirov, line 118: I think you want to create a persistent save_point here. Also, you'll need to commit wtx and reinitialise it directly after.

src/index/updater.rs
```diff
@@ -96,27 +98,46 @@ impl Updater {
...
+          dbg!(&prev_hash);
+          if prev_hash != block.header.prev_blockhash {
+            log::info!("reorg detected at or before {prev_height}; rolling back index");
+            self.index.reorged.store(true, atomic::Ordering::Relaxed);
+
+            let savepoint = wtx.get_persistent_savepoint(self.savepoints.pop_front().unwrap())?;
+            wtx.restore_savepoint(&savepoint)?;
+            self.savepoints.clear();
+          }
+        }
```
victorkirov, line 131: This logic should probably live before the create savepoint logic, otherwise we create a savepoint and immediately rollback to an older one.

src/index/updater.rs
victorkirov: Since you have to commit the txn and reinitialise it, it might make more sense for this logic to live directly after a `self.commit(wtx)` call or maybe even inside the commit function.

victorkirov: In fact, I think wtx has to be clean (no unflushed writes) before creating the savepoint, and you have to commit straight after it as it would make the wtx dirty.

raphjaph:
This was just me playing around I'm gonna make a new branch and PR with the correct one.

src/index/updater.rs
raphjaph: your right, I put it into the Index now

PR #2326, opened 2023-08-13
Ignore invalid content type header values

### LINKED ISSUE #2322: `ord server` errors on bad content-type
Example URL:

https://ordinals.com/content/4b5168a2c93f9d2838e77fc7967f721d35133270d74ed1ee9dba95f3b1fb9eb2i0

Example server error:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidHeaderValue', src/subcommand/server.rs:920:10
[2023-08-11T13:06:24Z INFO  ord::subcommand::server] GET /content/4b5168a2c93f9d2838e77fc7967f721d35133270d74ed1ee9dba95f3b1fb9eb2i0
```

The content-type is:

```image/avif\r\n```

The failing code is:

    headers.insert(
      header::CONTENT_TYPE,
      inscription
        .content_type()
        .unwrap_or("application/octet-stream")
        .parse()
        .unwrap(),
    );

The parse() is calling this:

https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.from_str

> Attempt to convert a string to a HeaderValue.
>
> If the argument contains invalid header value characters, an error is returned. Only visible ASCII characters (32-127) are permitted".

13 and 10 aren't permitted and the `Result<>` isn't being checked.

casey:
Fixes #2322.


veryordinally (approved):
Ready to merge

Issue #2329, opened 2023-08-14
Make a proper Inscriptions in Block Html page


raphjaph:

veryordinally:
- Performance improvements to get_inscriptions_in_block
- Proper title and navigation for the page (currently just using existing `inscriptions.html` template)

PR #2333, opened 2023-08-16
Make retrieving inscriptions in block fast

veryordinally:
Implementing first part of #2329 - performance improvements to `get_inscriptions_in_block` by introducing an in-memory index.


raphjaph (approved):
Excited to build stuff with this!

PR #2337, opened 2023-08-18
Proper Block Inscriptions HTML

veryordinally:
See #2329


raphjaph:
Getting: `thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 36849', src/index/block_index.rs:177:25`

templates/block.html
```diff
@@ -28,3 +28,11 @@ <h2>{{"Transaction".tally(self.block.txdata.len())}}</h2>
   <li><a href=/tx/{{txid}}>{{txid}}</a></li>
 %% }
 </ul>
+%% if self.inscriptions.len() > 0 {
```
raphjaph, line 31: I think inscriptions should be above transactions. Maybe we should also preview them like on the inscriptions page

veryordinally:
> Getting: `thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 36849', src/index/block_index.rs:177:25`

Improved error handling for when BlockIndex has not been initialized (happens on initial server startup, when index is initially catching up on new blocks). Message is now `Failed to get inscriptions in block: Block index not yet initialized`

raphjaph (approved):
Nice!

raphjaph (approved):
now its good

Issue #1685, opened 2023-02-11
Identify and link to previous inscriptions with same content 


veryordinally:
We should add a way to identify previous inscriptions with identical content. Could easily be achieved by hashing contents of each inscription. This would also allow to link in the explorer to other inscriptions with identical content, and clearly mark the first inscription with a given content.

veryordinally:
Related to #1539

Issue #1539, opened 2023-02-07
Add content hash


casey:
Displaying a hash of content is a good way of seeing if something is a duplicate of another. It would be interesting to display the hash of the content on /inscription.

huwyane:
any updates on this? content hashes are very important

nammaki:
Yeah, any update on this yet 👀

PR #2344, opened 2023-08-21
Content Hashes

veryordinally:
Adding content hashes to address #1539 and #1685


elocremarc:
@ordinaHO what is the status on this?

elocremarc:
Hey I am indexing your fork of this. This is no scientific test but it seems like its indexing a lot slower than usual. This might be because it has to hash each inscription when building the index. I wonder if we could make this an optional flag sort of like the `--index-sats` so we don't slow down base indexing for people who might not want this functionality?

raphjaph:
> It seems like its indexing a 3x slower than usual. I wonder if we could make this an optional flag sort of like the `--index-sats` so we don't slow down base indexing for people who might not want this functionality?

Should def do this @veryordinally

elocremarc:
I am using this branch a ton haha. Love it! 

Few suggestions after playing around with it.
I have already done a few free open editions based on hash. This was super helpful finding them all. One suggestion maybe for an additional API route would be something like /contenthash/:hash/:blockheight. Because the way I was doing my open editions was it was free mint before a certain block. So then marketplaces could just easily index them as they mint using that api.

You could also maybe show the iframe of all the inscriptions in the /contenthash/<hash> html page vs just list IDs. One of my open editions had identical hashes but each inscription generated a unique rendering because it was seeded off the inscriptionId.

If this is all out of scope for this PR I can maybe just do it in a subsequent one as an improvement.

casey:
This PR is pretty old, and has some unrelated changes. I think we still want this, but closing this PR for now.

stet:
> This PR is pretty old, and has some unrelated changes. I think we still want this, but closing this PR for now.

we def still want this :) revisit soon please.

thebitmaptoshi:
@raphjaph @casey is there any way to get the content hash stuff out of this PR and implement?

Adding text search capabilities would unlock a lot of potential

PR #2347, opened 2023-08-21
Reduce durability to speed up tests

casey:


raphjaph (approved):
Bit awkward, but does speed up tests 2x

PR #2128, opened 2023-05-29
GITBOOK: Jinglee's May 30 changes Handbook_CHN Version 1.0

DrJingLee:
Updated to 5-30 with the new donation address.

PR #2350, opened 2023-08-22
Handbook Chinese

DrJingLee:
mdbook type file;
new "chn" fold under docs with the same structure of english version.
except the chapter of “sat hunting” since its（out of date）

#2128


DrJingLee:
Same lint error

raphjaph:
This is a great start but I need #2385 implemented first.

DrJingLee:
> This is a great start but I need #2385 implemented first.

Cool try to work it out. #2385

PR #2352, opened 2023-08-22
Fix reorg typo

gmart7t2:
Height and depth are in the wrong order, leading to wrong log messages


raphjaph (approved):
lgtm!

casey:
@gmart7t2 Good catch! I changed `Reorg` to a record struct, so this mistake isn't possible.

PR #2353, opened 2023-08-22
Implement provenance in the index

casey:
This is an initial PR implementing [draft spec](https://github.com/ordinals/ord/pull/2278/) for parent/child inscriptions. It only implements indexing parent inscriptions and displaying them on the front end and returning them from the JSON API.

Other features are not implemented and are deferred to a future PR:
- Creating inscriptions with parents in the wallet
- Displaying child inscriptions on the parent inscription page

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ tower-http = { version = "0.4.0", features = ["compression-br", "compression-gzi
 [dev-dependencies]
 executable-path = "1.0.0"
 pretty_assertions = "1.2.1"
-reqwest = { version = "0.11.10", features = ["blocking"] }
+reqwest = { version = "0.11.10", features = ["blocking", "json"] }
 test-bitcoincore-rpc = { path = "test-bitcoincore-rpc" }
 unindent = "0.2.1"
 
--- a/src/index.rs
+++ b/src/index.rs
@@ -30,7 +30,7 @@ mod reorg;
 mod rtx;
 mod updater;
 
-const SCHEMA_VERSION: u64 = 5;
+const SCHEMA_VERSION: u64 = 6;
 
 macro_rules! define_table {
   ($name:ident, $key:ty, $value:ty) => {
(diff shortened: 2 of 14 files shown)


sondotpin:
What's about end point /parent/:inscriptionId to get metadata of parent inside child @casey

casey:
> What's about end point /parent to get metadata of parent inside child @casey

I'm not sure this is necessary. The parent is known to the inscriber at inscribe time, so they have and can use the parent's inscription ID and can use it in the child, no need to look it up from an API endpoint.

casey:
@cypherpork I just finished this initial implementation of provenance, in case you want to test/review.

sondotpin:
So as far as I understand, we can only solve the 1-dimensional problem at the moment. The child will point to the parent, but the parent won't know who their child is
@casey

cypherpork:
> @cypherpork I just finished this initial implementation of provenance, in case you want to test/review.

@casey, thank you - I have spun it up and started testing/reviewing. I will come back with notes in a day or two.

casey:
> > @cypherpork I just finished this initial implementation of provenance, in case you want to test/review.
> 
> @casey, thank you - I have spun it up and started testing/reviewing. I will come back with notes in a day or two.

And, note that there's a change to the implementation: https://github.com/ordinals/ord/pull/2278#issuecomment-1696376457

casey:
Okay, changed the implementation back to use the variable length encoding for the index. It's not a big savings on its own, but if we have a few savings like these they add up.

sondotpin:
@casey can we do multiple parents with one child?
```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_PUSH 1
  OP_PUSH 3
  OP_PUSH parent1_insription_id
  OP_PUSH 3
  OP_PUSH parent2_insription_id
  OP_PUSH 0
  OP_PUSH "Hello, world!"
OP_ENDIF
```

casey:
> @casey can we do multiple parents with one child?

I'd like to enable this eventually, but it's not in the first version. What's the use case you're thinking of?

sondotpin:
1. breeding nft
2. whitelist voting, Dao, Group

raphjaph (approved):
v exciting!

PR #2357, opened 2023-08-23
Testing guide

elocremarc:
Part of the intent of [PR#2215](https://github.com/ordinals/ord/pull/2215/commits/e6bf6a2ad2a8c23cfeca10ab16c75d6935fe48c6)
was to document how to use ord to test out regtest for use in recursion. As well as test out more complicted HTML inscriptions like libraies or multiple linked pages. I don't think `ord preview` is sufficent for this usecase. Therefore this Pr is to add a `testing` page for how to use the network flags in ord as well as an example regtest workflow. I use this religiously when doing my own HTML inscriptions and this will help people avoid testing recursion on mainet.


raphjaph (approved):
Thanks for the guide! Made some small clarifications

PR #2359, opened 2023-08-23
Unrecognized even tag unbound and negative

raphjaph:


casey (approved):

devords:
> Digital artifacts can have owners. A number is not a digital artifact, because nobody can own it.

> The definition of a digital artifact is intended to reflect what NFTs should be, sometimes are, and **what inscriptions always are, by their very nature**.

Would these properties still be true for these inscriptions, if these are not bound to specific sats?

devords:
> Digital artifacts can have owners. A number is not a digital artifact, because nobody can own it.

> The definition of a digital artifact is intended to reflect what NFTs should be, sometimes are, and **what inscriptions always are, by their very nature**.

@casey @raphjaph Would these properties still be true for these inscriptions, if these are not bound to specific sats?

PR #2362, opened 2023-08-24
add ordinal handbook arabic version

ordinaHO:


veryordinally:
Clippy tests seem to fail.
Need to work out structure for adding new languages. See for example https://github.com/ordinals/ord/pull/2128 for another example (Chinese).

DrJingLee:
Seems  CI / lint (pull_request) is failed.
The error says on the line 29 of "tests/parse.rs", the # is not nessessary ：

_.stderr_regex(r#"error: .*: unrecognized object\n.*"#)_

ChatGPT Advice to revise to but i have no idea:

**.stderr_regex(r"error: .*: unrecognized object\n.*")**

DrJingLee:
> Clippy tests seem to fail. Need to work out structure for adding new languages. See for example #2128 for another example (Chinese).

Exactly. I try to merge the handbook by creating a new sub-folder called "chn" under "docs" ,the url will be docs.ordinals.com/chn.and "chn" have the same structure with the "docs" files. 

We might need to revise the index of urls to link different version of handbook translations. 

Does this workable ? @veryordinally @raphjaph

ordinaHO:
> Seems CI / lint (pull_request) is failed. The error says on the line 29 of "tests/parse.rs", the # is not nessessary ：
> 
> _.stderr_regex(r#"error: ._: unrecognized object\n._"#)_
> 
> ChatGPT Advice to revise to but i have no idea:
> 
> **.stderr_regex(r"error: ._: unrecognized object\n._")**

Yep i think i worked my way out of this , installed nodejs markdown lint  aswell and i was able to scan my errors and now should be ok 👌

ordinaHO:
> 
i think that what should be updated into the new bool.toml to run the new version on ordinals.com ! for now im using my local server to check if all the changes im making is in the correct form i need .

veryordinally:
OK, confirm that the remaining clippy issue is due to new rust release.

raphjaph:
If someone could implement this first: https://github.com/ordinals/ord/issues/2385 that would be awesome!

ordinaHO:
> 

Hi raph , i took a look about the plugins i thinnk it for translation of documentation written in mdbook into multiple languages. Is that what are you referring ? im not sure if there is Arabic as i went through the book almost each word and the most hard part is that arabic writing is from right to left and a lot of words cant just be automatically translated.
Both ways will follow up on this so we can figure the best way so all incoming requests will have one commen steps to do .
best regards

raphjaph:
I just merged https://github.com/ordinals/ord/pull/2403, so have a look at the updated [README](https://github.com/ordinals/ord#translations) for instructions.

PR #2368, opened 2023-08-25
Fix `wallet inscriptions` so it can list multiple inscriptions on the…

### LINKED ISSUE #2367: `wallet inscriptions` only lists one inscription per sat
Ord recognizes reinscriptions as cursed, but "ord wallet inscriptions" only lists one inscription per sat. I've seen it list the cursed inscription, or the regular one, but never both together.

This happens in ord 0.8.2.

The json /output/ endpoint correctly lists them all.

gmart7t2:
…  #same sat

Fix #2367.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -864,9 +864,20 @@ impl Index {
     &self,
     utxos: BTreeMap<OutPoint, Amount>,
   ) -> Result<BTreeMap<SatPoint, InscriptionId>> {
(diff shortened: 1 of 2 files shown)


raphjaph:
Good catch! Would be great to have a test for this edge case

raphjaph:
Expanded this and added tests in #3101

PR #2374, opened 2023-08-27
Make homepage more interesting

raphjaph:


veryordinally (approved):
So much more interesting!

PR #2383, opened 2023-08-28
Add pointer spec

casey:
This specifies an offset field, which allows inscriptions to be made on output sats other than the first. This allows making multiple inscriptions per transaction.


rot13maxi:
I believe this will also prevent child inscriptions from being cursed

docs/src/inscriptions/offset.md
```diff
@@ -0,0 +1,48 @@
...
+offset can be provided with tag `2`, causing the inscription to be made on the
+sat at the given offset. If the offset is equal to or greater than the number
+of total sats in the outputs of the inscribe transaction, it is ignored, and
+the inscription is made as usual. The value of the offset tag is the little
+endian offset, with trailing zeros omitted.
+
+An even tag is used, so that old versions of `ord` consider the inscription to
+be unbound, instead of assigning it, incorrectly, to the first sat.
+
+This can be used to create multiple inscriptions in a single transaction on
```
rot13maxi, line 14: unclear what the mechanism is to create multiple inscriptions. To make `n` inscriptions do you need `n` inputs with an inscription envelope in the witness of each one? Can you put multiple inscriptions in the same input?

raphjaph: you can do either. At the moment inscriptions in different inputs assigned to first sat of their respective input and inscriptions in same input are made unbound (they will be made reinscriptions on first sat of input soon)

docs/src/inscriptions/pointer.md
```diff
@@ -0,0 +1,48 @@
+Pointer
+=======
+
+In order to make an inscription on a sat other than the first of its input, a
+pointer can be provided with tag `2`, causing the inscription to be made on the
+sat at the given position in the outputs. If the pointer is equal to or greater than the number
```
raphjaph, line 6: starting at 0

raphjaph (approved):

PR #2384, opened 2023-08-28
Remove unused content_response match statement

casey:
The `None` case of this match statement will never have an effect, since we do `body?` later one, which will case the whole function to return `None` if the body is empty.


raphjaph (approved):
Correct, let's try to put it in the right spot

PR #2397, opened 2023-08-30
Make any zero-valued input inscription unbound

raphjaph:
Fixes https://github.com/ordinals/ord/issues/2386


casey (changes requested):
For the record, this makes me want to kms 😂 But this is inline with the idea that inscriptions are made on input sats.

This looks good, but we need to test that we aren't unbinding any existing bound inscriptions first, i.e., envelopes on zero-value inputs where the total input value is nonzero.

casey (approved):

PR #2394, opened 2023-08-30
zh.po documents by the mdbook-i18n-helpers

DrJingLee:
Following #2385, 

Construct the new documents by mdbook-i18n-helpers;
Complete the Chaptor 3.
updating the remaining.

following ISO-639-1 code standard of languages:
zh = Chinese Version of Language

PR #2406, opened 2023-08-31
Handbook Chinese using mdbook-i18n-helpers

DrJingLee:
revised the index.hbs to add zh language;
upload zh.po into docs fold;
local test good,request for merge.
#2394


raphjaph (approved):
Awesome, great work! Just removed some unnecessary file and added to our build system

PR #2409, opened 2023-08-31
Allocate blocks vector ahead of time

casey:
Implements #1893. This is a pretty straightforward optimization, so I think we should just implement it.


veryordinally (approved):
Ready to merge.

Issue #2053, opened 2023-04-27
Add named properties directly to inscription data


cognitive-canvas:
Incorporating named properties can significantly enhance inscriptions, especially with regards to NFT compatibility and other applications involving metadata. Although there have been various attempts to integrate metadata, they often involve a certain degree of complexity and indirection, such as using JSON inscriptions that reference other ordinals. This approach is not only unwieldy but also challenging to maintain.

To streamline this process, I propose adding named properties directly to the inscription data. One way to do it is the following:

```
OP_5 <KEY_0>
OP_7 <VALUE_0>
OP_5 <KEY_1>
OP_7 <VALUE_1>
...
```

Adopting this method for inscriptions would transform them into key-value stores, effectively resolving a multitude of issues and ensuring entirely on-chain operations. Additionally, this approach appears to be feasible for implementation in a way that is both forward and backward compatible.

casey:
Since metadata isn't relevant to the protocol, if we use this approach, we should stuff all metadata in a single tag. See #943 for some discussion.

casey:
After thinking about metadata a bit, I'm actually warming up to this approach. I modified the tags to be odd, and use the next available odd tags after the upcoming parent tag.

casey:
@veryordinally Actually thinking now that this might be the best approach. What do you think? Also @raphjaph, take a look at this.

Psifour:
We are actively using this in our tech demo of quadkey.

While you will 100% not be interested in the product (for obvious reasons), the usage is a really cool example of how these can be used.

Specifically, we use the following envelope:

`OP_FALSE
OP_IF
OP_PUSH "ord"
OP_PUSH 1
OP_PUSH image/svg
OP_PUSH "qey"
OP_PUSH {4 byte binary representation of the quadkey}
OP_PUSH 0
OP_PUSH {content blob}
OP_ENDIF`

I spend a fair bit of time looking at on-chain transactions and the envelope (as well as indexing non-standard inscriptions) so was very comfortable with the current rule that user-defined/application-specific operations needed to be prefixed with a byte that is odd (to avoid conflicts with protocol reserved even bytes), this led to us using 'qey' (q for quadkey instead of k for key) which begins with an odd byte.

I would love to engage on this topic as I have already been experimenting with this probably more than most other developers in the space. Interesting envelope mechanics I have explored include (but are not limited to):

[...]

PR #2421, opened 2023-09-04
Inscription metadata

casey:
Add metadata, based on #2053.


docs/src/inscriptions/metadata.md
```diff
@@ -0,0 +1,39 @@
...
+  ...
+  <dt>metadata</dt>
+  <dd>
+    <dl>
+      <dt>foo</dt>
+      <dd>bar</dd>
+      <dt>this</dt>
+      <dd>that</dd>
+    </dl>
+  <dd>
```
onepixelinscriptions, line 42: </dd>

elocremarc:
We should map keys to the [Open Graph Protocol](https://ogp.me/) standard so metatags could also display on socials and SEO.
Example for `/inscription` page:
  ```html
    <head>
      <meta property="og:foo">bar</meta>
      <meta property="og:this">that</meta>
      <meta property="og:title">dope title</meta>
      <meta property="og:image" content="/content/<inscriptionId>" />
    </head>
```

File: templates/inscription.html [hunk 2]
```diff
@@ -19,6 +19,17 @@ <h1>Inscription {{ self.number }} (unstable)</h1>
...
+  <dd>
+    <dl>
+%% for (key, value) in self.inscription.metadata() {
+      <dt>{{ key }}</dt>
+      <dd>{{ value }}</dd>
+%% }
+    </dl>
+  </dd>
+%% }
 %% if let Some(parent) = &self.parent {
```
elocremarc, line 33: Do you think we should  also add these to the metatags on the head of the `page.html` if they key to open graph keys? That way someone could have the title of the inscription page be about what the inscription is instead of the inscription number if they desire as well as a description. Then links to this page could show more about the inscription on say twitter. Or would that not be ideal because then people can shit post there haha.
See comment: 
https://github.com/ordinals/ord/pull/2421#issuecomment-1705677285

casey: I think we wouldn't want to add arbitrary metadata, although if there is a `title` and `description` field present, those could be added. This is probably out-of-scope for the initial spec and PR though, just because it's ideal to keep PRs as small as possible so they stay managable.

elocremarc: Yeah for sure agreed.

src/inscription.rs
elocremarc, line 1: This might already be the case. Just wondering if this returns an error if the `METADATA_TAG` is over 520 bytes? @Psifour was telling me there are issues if its over this amount. 
"The KV storage breaks in almost all indexers of your K or V are larger than that" -Psifour

casey: Yup, that's correct. Taproot data pushes are limited to 520 bytes, so keys and values must be 520 bytes or less.

elocremarc: Should this be added to the metadata.md?
```diff
+```Note: Taproot data pushes are limited to 520 bytes, so keys and values must be 520 bytes or less.```
```

casey: Yup, good call.

elocremarc:
Needs a page ref on `SUMMARY.md`
```markdown
 - [Inscriptions](inscriptions.md)
     - [Recursion](inscriptions/recursion.md)
     - [Metadata](inscriptions/metadata.md)
```

0xEclair:
Maybe we could support parse metadata in a file?

uzyn:
Would be great if we can align with [OIP-01](https://www.oips.io/oip-01-inscription-metadata), a non-breaking metadata implementation. 

```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_PUSH 1
  OP_PUSH 0
  OP_PUSH "Hello, world!"
OP_ENDIF

OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_PUSH 1
  OP_PUSH "application/json;charset=utf-8"
  OP_PUSH 0
  OP_PUSH "{\"p\":\"oip-01\"}"
OP_ENDIF
```

It's been implemented by [Sado SDK](https://sado.space), [Ordzaar](https://ordzaar.com), [Ordlify](https://ordlify.com/). 

Here's an example of a defined metadata using OIP-01: https://explorer.ordzaar.com/inscription/105758bb912665f5f803ec0f5268d2218b51978b16de05622c64c9faafd2d22ei0

While remaining fully backward-compatible, so it shows up fine on Ordinals.com https://ordinals.com/inscription/105758bb912665f5f803ec0f5268d2218b51978b16de05622c64c9faafd2d22ei0


It has a looser format, JSON vs named key-value pairs, for metadata than #2053, allowing for more flexible use cases such as: [verifiable ordinal collection](https://www.oips.io/oip-02-verifiable-ordinal-collection) and [file chunking](https://www.oips.io/oip-04-chunking-of-inscriptions-for-larger-files).

Happy to discuss further on this.

casey:
> Needs a page ref on `SUMMARY.md`
> 
> ```
>  - [Inscriptions](inscriptions.md)
>      - [Recursion](inscriptions/recursion.md)
>      - [Metadata](inscriptions/metadata.md)
> ```

Good catch!

casey:
> Would be great if we can align with [OIP-01](https://www.oips.io/oip-01-inscription-metadata), a non-breaking metadata implementation.

We considered something like this, but I see a few issues with using a different inscription on the same sat to contain metadata, but it seems inelegant to make the metadata its own inscription, when it isn't interesting on its own, and shouldn't be displayed as an inscription in its own right.

kodemon:
> metadata its own inscription, when it isn't interesting on its own, and shouldn't be displayed as an inscription in its own right

This makes sense to me, as having secondary inscriptions on sats being interpreted as unbound leaves inconsistent representation in the consumer layers _(eg. one place may support OIP where as another may not)_.

I think a single layer approach with OP codes may be a little limiting. Having meta data as a flexible object with ability to add nested separation of concern and lists through JSON is more beneficial than a key value pair approach.

So something like:
```
OP_5 meta # marker, when value "meta" next value OP_7 should be JSON
OP_7 json
```

So I definitely like an approach where we don't create two inscriptions on the same sat, but would miss having the ability to add a somewhat more complex meta object _(unless there are some elegant solutions to this using the proposed approach)_. 🤔

casey:
> but would miss having the ability to add a somewhat more complex meta object

What specific metadata did you have in mind? From what I've observed, a simple key-value store covers the vast majority of use-cases, and has the advantage that it's attractive and simple to display.

kodemon:
I was thinking in line of the meta presented in this example: https://explorer.ordzaar.com/inscription/105758bb912665f5f803ec0f5268d2218b51978b16de05622c64c9faafd2d22ei0

So for example a `trait` key with a list of traits. or the creator with its own key value pair object. The one thing I could think of was doing:
```
OP_5 creator.name
OP_7 TheArtist
```
But for the traits list it may be a little more complicated 🤔

0xEclair:
> a simple key-value store covers the vast majority of use-cases, and has the advantage that it's attractive and simple to display.

Many traits need nested key value pairs. In the scene, simple key-value pair could make no sense. Maybe json is a good way to store traits.

uzyn:
@casey 
 
> We considered something like this, but I see a few issues with using a different inscription on the same sat to contain metadata, but it seems inelegant to make the metadata its own inscription, when it isn't interesting on its own, and shouldn't be displayed as an inscription in its own right.

Agree on all the above. OIP-1 was designed, in order to not break existing support, based on a specific negative test case at the time of proposal (where trailing inscription is expected to be ignored). It seems to now no longer true, due to https://github.com/ordinals/ord/pull/2145

Trying to be as accommodating as possible to existing adopted social standards, e.g. https://www.oips.io/oip-02-verifiable-ordinal-collection utilizes [OIP-1](https://www.oips.io/oip-01-inscription-metadata) but is proposed based on BRC-20. 

I would adopt the model, but would highly recommend doing a more structured data in a similar style of BRC-20 for more extensibility.

orkunkilic:
What do you think of supporting inner metadata with smth like this @casey: 

```
 OP_5 just_key 
 OP_7 just_value
 OP_5 outer_key
 OP_7 
   OP_5 inner_key_and_outer_value
   OP_7 inner_value
```

We can limit inner depth with some const like 3? I am just brainstorming to have more flexible metadata structure while avoiding json.

casey:
@kodemon

> I was thinking in line of the meta presented in this example: https://ordinals.com/content/105758bb912665f5f803ec0f5268d2218b51978b16de05622c64c9faafd2d22ei1

There are some nested fields, but I feel like they could avoided and expressed as key-value pairs:

| key | value |
| --- | --- |
| Title | Elemental |
| Description | Azuki Elementals are a collection of 20,000 characters within the four domains of the Garden. |
| Hair | Electrified Long - Black |
| Offhand | Elemental Blade - Lightning |
| Eyes | Enticing |
| Type | Blue |
| Creator | TheArtist <artist@example.org> |
| Tips | 1FFBzYT55auox4boSaQUYcsQ9LCaJJWhRw |

@0xEclair

> Many traits need nested key value pairs. In the scene, simple key-value pair could make no sense. 

Examples of traits that need nested key value pairs would be great.

@orkunkilic 

> What do you think of supporting inner metadata with smth like this @casey:

In that case it would probably be better to use JSON, instead of rolling a new serialization format.

@uzyn

> I would adopt the model, but would highly recommend doing a more structured data in a similar style of BRC-20 for more extensibility.

The purpose of the inscription metadata proposed here is to always for it to be human readable. My worry with structured data is that it won't be possible to display it to the user in an appealing format.

0xEclair:
> There are some nested fields, but I feel like they could avoided and expressed as key-value pairs

Yeah. Normal key value pairs could be like this. But for developers, maybe they need distinct types of keys(for display or for using in app). If only for display, it's no problem. I think metadata also needs more information to help developer to use them.

kodemon:
> The purpose of the inscription metadata proposed here is to always for it to be human readable. My worry with structured data is that it won't be possible to display it to the user in an appealing format.

I think this is probably where the mindset deviates as I am third party consumer of ord and shape the data it provides into product specific requirements. So for my use cases structured data is more important than ease of display as that is something I deal with during formatting.

So lets imagine the following object:

```
{
  "p": "super",
  "traits": [
    {
      "power": "anything",
      "strength": 10
    },
    ...
  ],
  ...other
}
```

The tricky part with this requirement is that its hard to flatten because the traits array may be dynamic, so if I flatten this it becomes hard to identify what is part of a trait and what is not. Where as when I am processing traits as part of my application I may have some internal logic specific for handling dynamic trait information.

So while I understand the appeal of human readable out of the box, as a third party developer relying on the data ord provides and the way that I personally prefer meta data to be presented I still think a flat key value pair representation is too limiting and I can see myself needing to hack around the limitation rather than just being able to use the data as is.

I would probably need to do something like

key | value
----|-------
p | super
data | { "traits": [ ... ], ... }

I think from a developers perspective consuming ord as a service, a structured meta object is more future proof.

For this specific issue I decided to be a little more vocal in my thoughts as it has a massive impact on the work I do with ord 😅

casey:
> For this specific issue I decided to be a little more vocal in my thoughts as it has a massive impact on the work I do with ord and we heavily rely on the ability add meta data to inscriptions 😅

Your feedback is very welcome!

I think that, as much as I would like to constrain metadata to key-value pairs, it's a losing battle, so I changed the spec to use JSON instead. I've also included, as part of the spec, examples of how `ord` will render JSON metadata as HTML on `/inscription` pages. Hopefully this will encourage inscribers to make their metadata legible, and to omit anything from metadata that isn't intended for human consumption.

bruffstar:
It's great this was moved to a JSON setup. I think it would be really useful to have an endpoint where we get the metadata returned as JSON. 

e.g. `https://ordinals.com/inscription/<inscription-id>/metadata`

This way, we can consume the JSON and display it any way we want.

Psifour:
To confirm, the plan for ord going forward is uncompressed UTF-8 JSON as a metadata format?
Despite all the excellent ways to encode data we have chosen one of the least data efficient ones to implement?

casey:
> To confirm, the plan for ord going forward is uncompressed UTF-8 JSON as a metadata format? Despite all the excellent ways to encode data we have chosen one of the least data efficient ones to implement?

I'm not set on this, but I don't think it's an entirely unreasonable choice. Yes, JSON is inefficient, however it is ubiquitous, and being human readable and writable is a huge advantage.

stet:
> human readable and writable is a huge advantage

I encourage use of efficient CBOR etc encoding. Not sure why it matters if metadata is immediately readable if it is rendered for consumer in appropriate format (which could be transformed to JSON if needed). Saving on bytes and bandwidth and energy efficiency are important considerations esp in context of Bitcoin onchain data store. 

This is a great opportunity to be smart about the broad impact as opposed to a narrow-minded approach focused on _display_ of data. 

I don't think people and software should be writing and inscribing raw JSON moving forward.

kodemon:
> I encourage use of efficient CBOR etc encoding. Not sure why it matters if metadata is immediately readable if it is rendered for consumer in appropriate format (which could be transformed to JSON if needed).

I agree with the take that readability should be a consumer level concern and not dictate its storage format in a way that is less efficient than it can be.

It would be interesting to hear some good ideas on an efficient and flexible solution for storing meta data on inscriptions. I don't have in depth knowledge on the topic of optimized encode/decode topics, and I am mostly concerned with the ability to consume the data stored to transform it into output that is used in product and/or protocol level decisions.

raphjaph (approved):
LGTM! 

Let's add special title and description display in another PR

PR #2423, opened 2023-09-04
Implement `/block/:query` JSON endpoint

terror:


tests/json_api.rs
```diff
@@ -397,3 +397,29 @@ fn json_request_fails_when_not_enabled() {
...
+
+  assert_eq!(response.status(), StatusCode::OK);
+
+  let block_json: BlockJson = serde_json::from_str(&response.text().unwrap()).unwrap();
+
+  assert_eq!(
+    block_json,
+    BlockJson {
+      hash: String::from("000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"),
+      target: String::from("0000000000000000000000000000000000000000000000000000ffff00000000"),
```
casey, line 418: This target looks backwards to me. It should have one bits on the left side, since it's a number that the block hash must be less than.

casey: We should fix it in both places, both here and on the non-json block page.

casey (changes requested):
Mostly looks good, only comment is that we should fix the target.

casey (approved):

PR #2431, opened 2023-09-07
Add children recursive endpoint

elocremarc:
We have the recursive endpoint `/content/` which opened up quite a lot of creative ability. With the addition of provenance I think it would be really cool to have a way for a parent inscription to be aware of any of its decedents. I propose we add the recursive endpoint `/children/:inscriptionId`.

This could be used for a parent inscription to be a gallery of all of its children. It also gives the ability for creators to make inscriptions where both the creator and the owner can take part in recursively creating the content of what is on a parent inscription. Other benefits allow more complex HTML inscriptions where say a CSS file could be inscribed as a child or even updated potentially. 

Edit:
For this to work you get the `inscriptionId` of the parent with javascript and then feed it into the endpoint so that it can see what inscriptions it has as children. So it would be `/children/<Parent_InscriptionId>`

Example:
```html
<!DOCTYPE html>
<html>
<body>
  <script>
    fetch(`${window.location.origin}/children/${window.location.pathname.split('/').pop()}`)
      .then(response => response.json())
      .then(data => console.log(data))
      // Console logs all the children inscribed later.
  </script>
</body>
</html>

```


rot13maxi:
You wont be able to do static child inscription ids in the parent because there would be a circular dependency when it comes to pre-determining the inscription IDs. 

Id suggest you do an endpoint to list the children of a parent. The parent inscription can then have some js that iterates through the list and fetches each one via the content endpoint.

elocremarc:
> Id suggest you do an endpoint to list the children of a parent. The parent inscription can then have some js that iterates through the list and fetches each one via the content endpoint.

Yeah that is what this does. You get the `inscriptionId` of the parent with javascript and then feed it into the endpoint so that it can see what inscriptions it has as children. So it would be `/children/<Parent_InscriptionId>`

rot13maxi:
Oh, I misunderstood. I thought the inscription-id in the path was meant to be the child

rot13maxi (approved):

elocremarc:
I think I should change this to not return all the inscriptions and only return 1 at a time. Also this should probably have a content route.
`/children/:inscriptionId_parent/:index`
`/content/children/:inscriptionId_parent/:index`
See https://github.com/ordinals/ord/pull/2452 for new approach.

BitTogether:
I love this! FINALLY I can make dynamic art where the collector and artist can make art together like never before possible!

raphjaph:
Since we are adding more and more recursive endpoints and these will start to clash with existing html endpoints, what do you think about putting all new ones under `/-/`?

zmeyer44:
> Since we are adding more and more recursive endpoints and these will start to clash with existing html endpoints, what do you think about putting all new ones under `/-/`?
> 
> #2476

Strictly from an aesthetic point of view, I'd be hesitant to add the `/-/` prefix. I don't see why we couldn't continue to just use the base paths. Given that these endpoints will be utilized in sandboxed iframes, it should be easy enough for clients to ensure they are managing conflicts on their end. It is also a bit odd to have some recursive endpoints with the `/-/` and some without.

src/subcommand/server.rs
```diff
@@ -1346,6 +1351,35 @@ impl Server {
...
+    Self::children_recursive_paginated(Extension(index), Path((inscription_id, 0))).await
+  }
+
+  async fn children_recursive_paginated(
+    Extension(index): Extension<Arc<Index>>,
+    Path((parent, page)): Path<(InscriptionId, usize)>,
+  ) -> ServerResult<Response> {
+    let parent_sequence_number = index
+      .get_inscription_entry(parent)?
+      .ok_or_not_found(|| format!("inscription {parent}"))?
```
raphjaph, line 1367: should we return 404 or just an empty id array?

src/subcommand/server.rs
```diff
@@ -238,6 +238,11 @@ impl Server {
         )
         .route("/r/blockheight", get(Self::block_height))
         .route("/r/blocktime", get(Self::block_time))
+        .route("/r/children/:inscription_id", get(Self::children_recursive))
+        .route(
+          "/r/children/:inscription_id/:page",
+          get(Self::children_recursive_paginated),
+        )
         .route("/r/metadata/:inscription_id", get(Self::metadata))
```
raphjaph, line 246: should we add a `/r/children/:inscription_id/at/:index` as well?

src/subcommand/server.rs
```diff
@@ -2496,6 +2530,69 @@ mod tests {
...
+    assert_eq!(children_json.page, 0);
+
+    let children_json =
+      server.get_json::<ChildrenJson>(format!("/r/children/{parent_inscription_id}/1"));
+
+    assert_eq!(children_json.ids.len(), 11);
+    assert_eq!(children_json.ids[0], hundred_first_child_inscription_id);
+    assert_eq!(children_json.ids[10], hundred_eleventh_child_inscription_id);
+    assert!(!children_json.more);
+    assert_eq!(children_json.page, 1);
```
raphjaph, line 2593: Test the error response

src/subcommand/server.rs
raphjaph: test that parent not found returns 404 and empty array if parent found but no children

raphjaph (approved):
lgtm!

PR #2435, opened 2023-09-07
handbook arabic/العربية and german/deutsch version 

ordinaHO:
Tested on local server , few commands added to index.hbs to help with  errors that I was having regarding the RTL languge situation also rtl.css 
added german translation


raphjaph:
Could you create two seperate PRs? This would make review easier. Thanks!

PR #2442, opened 2023-09-09
arabic language for ordinal handbook

ordinaHO:
while serving on local , i had to add some script as you see so the language will display correctly and avoiding any errors! this one was kind of hard to do but will see how it goes
thanks in advance


raphjaph:
This looks good but can't really verify that everything is correct. Did you have someone proof-read it? 

Also everything is right-aligned, which needs to be for Arabic, but if this looks good for people I'll just merge it.

Screenshot 2023-10-03 at 20 21 20

ordinaHO:
hey @raphjaph I went through multiple times and all should be correct and ofcours will need to update some details here and there ! currently im on vacation , however I will have acsses to my office next week before amesterdam conference to go through again .

raphjaph (approved):
Looks good to me!

PR #2448, opened 2023-09-13
Add Spanish Translation

Psifour:
Translation by Zerone
Implementation into mdbook by Psifour

- [x] Added to language selection
- [x] Verified localization
- [x] Updated to 0.9 documentation
- [x] Added to just


Psifour:
Added all localization .po files within docs/po to the bin/forbid whitelist for the purposes of the forbidden word check. This check is intended to avoid inclusion (filtering for "todo" and dbg"), but conflicts with the use of similar words in other languages, like the word "todo" in Spanish.

raphjaph:
> Added all localization .po files within docs/po to the bin/forbid whitelist for the purposes of the forbidden word check. This check is intended to avoid inclusion (filtering for "todo" and dbg"), but conflicts with the use of similar words in other languages, like the word "todo" in Spanish.

Yes, good add.

> Failed test appears to be flakey as no relevant code to it was changed in [49314d7](https://github.com/ordinals/ord/commit/49314d726fec977a3a29e51bd093f57b2ab33e81)

Yeah on Windows there's one flaky test.

raphjaph (approved):
Gracias por la magnífica traducción!

PR #2449, opened 2023-09-13
Add metaprotocol field

casey:
Allows inscriptions to declare their protocol in a dedicated field.


elocremarc:
This is exactly what I needed!

PR #2460, opened 2023-09-20
Add sequence number

raphjaph:
Adds an unstable sequence number to make some logic cleaner

- [x] maybe remove sequence number from front-end

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -1007,9 +1007,9 @@ impl Index {
     let rtx = self.database.begin_read()?;
 
     let height_to_last_sequence_number = rtx.open_table(HEIGHT_TO_LAST_SEQUENCE_NUMBER)?;
-    let inscription_id_by_sequence_number = rtx.open_table(SEQUENCE_NUMBER_TO_INSCRIPTION_ID)?;
+    let sequence_number_to_inscription_id = rtx.open_table(SEQUENCE_NUMBER_TO_INSCRIPTION_ID)?;
 
-    let Some(next_sequence_number) = height_to_last_sequence_number
+    let Some(newest_sequence_number) = height_to_last_sequence_number
       .get(&block_height)?
       .map(|ag| ag.value())
     else {
@@ -1021,15 +1021,23 @@ impl Index {
       .map(|ag| ag.value())
       .unwrap_or(0);
 
+    dbg!(&sequence_number_to_inscription_id
+      .iter()?
+      .map(|result| result.ok().unwrap())
+      .map(|(a, b)| (a.value(), InscriptionId::load(*b.value())))
+      .collect::<Vec<(u64, InscriptionId)>>());
+
     Ok(
-      (oldest_sequence_number..next_sequence_number)
-        .map(|num| match inscription_id_by_sequence_number.get(&num) {
-          Ok(Some(inscription_id)) => Ok(InscriptionId::load(*inscription_id.value())),
-          Ok(None) => Err(anyhow!(
-            "could not find inscription for inscription number {num}"
-          )),
-          Err(err) => Err(anyhow!(err)),
-        })
+      dbg!(oldest_sequence_number..newest_sequence_number)
+        .map(
+          |num| match sequence_number_to_inscription_id.get(dbg!(&num)) {
+            Ok(Some(inscription_id)) => Ok(InscriptionId::load(*inscription_id.value())),
+            Ok(None) => Err(anyhow!(
+              "could not find inscription for inscription number {num}"
+            )),
+            Err(err) => Err(anyhow!(err)),
+          },
+        )
         .collect::<Result<Vec<InscriptionId>>>()?,
     )
   }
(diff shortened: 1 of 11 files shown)


templates/inscription.html
```diff
@@ -23,6 +23,8 @@ <h1>Inscription {{ self.number }} (unstable)</h1>
   <dt>parent</dt>
   <dd><a class=monospace href=/inscription/{{ parent }}>{{ parent }}</a></dd>
 %% }
+  <dt>sequence number</dt>
+  <dd>{{ self.sequence_number }}</dd>
```
elocremarc, line 27: I think that people don't want to add a sequence number below a certain activation number to keep the original numbers prominent socially. What if we didn't show the sequence number below a threshold and instead just showed that if you run the API? We would still index both but we would only show inscription number for earlier inscriptions.
```html
%% if self.inscription_number >= 50,000,000 {
<dt>sequence number</dt>
  <dd>{{ self.sequence_number }}</dd>
  }
```

elocremarc: Another idea maybe we think of this like an engineering tolerance.
```html
%% if self.inscription_number >= 50,000,000 {
<dt>sequence number</dt>
  <dd>+ {{ self.sequence_number - self.inscription_number }}</dd>
  }
  elseif self.inscription_number < 50,000,000 &&  self.inscription_number > 0{
  <dt>sequence offset</dt>
  <dd>+ {{ self.sequence_number - self.inscription_number }}</dd>
  }
  elseif self.inscription_number < 0{ 
    <dt>sequence number</dt>
  <dd>+ {{ self.sequence_number }}</dd>
  }
  
  ```

Comment (casey) [reply templates/inscription.html:27]: We specifically want to display sequence numbers so that people understand exactly how inscription numbers would change *if* they were made unstable. Right now people are pretty much in the dark.

Comment (elocremarc) [reply templates/inscription.html:27]: The sentiment I found on all the twitter spaces is people don't want to change the prior inscription numbers at all behind a certain activation number. Their view is the trust in the entire protocol will be broken if we decide to change these numbers and make them unstable and we changed to a sequence number. 

I think the solution that a lot of people are in agreement with on both sides is @huuep comment:
https://github.com/ordinals/ord/discussions/2458#discussioncomment-7051777
He proposes that we immutably store the prior inscription numbers using an inscription. Then at a certain inscription number activation we will then move forward with immutable sequence numbers. 

I tend to agree with you. However I want to make sure the other voices are heard here.

Comment (raphjaph) [reply templates/inscription.html]: yeah gonna remove it from the front-end

File: src/index.rs [hunk 2]
```diff
@@ -663,27 +665,32 @@ impl Index {
...
     if ids.len() > 1 {
-      let re_id_to_seq_num = rtx.open_table(REINSCRIPTION_ID_TO_SEQUENCE_NUMBER)?;
-      ids.sort_by_key(
-        |inscription_id| match re_id_to_seq_num.get(&inscription_id.store()) {
-          Ok(Some(num)) => num.value() + 1, // remove at next index refactor
+      let inscription_id_to_entry = rtx.open_table(INSCRIPTION_ID_TO_INSCRIPTION_ENTRY)?;
+      ids.sort_by_key(|inscription_id| {
+        match inscription_id_to_entry
+          .get(&inscription_id.store())
+          .unwrap()
```
casey, line 672: Remove unwrap

casey: You could build a vec of sequence numbers with a for loop, which would be fallible and return errors, then zip together with the list of ids and sort.

casey: You can also unwrap if there's no entry, since that's an invariant.

src/index/updater.rs
```diff
@@ -378,24 +378,32 @@ impl<'index> Updater<'_> {
...
+    let mut sequence_number_to_inscription_id =
+      wtx.open_table(SEQUENCE_NUMBER_TO_INSCRIPTION_ID)?;
     let mut statistic_to_count = wtx.open_table(STATISTIC_TO_COUNT)?;
 
     let mut lost_sats = statistic_to_count
       .get(&Statistic::LostSats.key())?
       .map(|lost_sats| lost_sats.value())
       .unwrap_or(0);
 
+    let num_cursed_inscription = statistic_to_count
```
casey, line 397:
```suggestion
    let cursed_inscription_count = statistic_to_count
```

src/index/updater.rs
```diff
@@ -378,24 +378,32 @@ impl<'index> Updater<'_> {
...
+      .get(&Statistic::CursedInscriptions.key())?
+      .map(|count| count.value())
+      .unwrap_or(0);
+
+    let num_blessed_inscriptions = statistic_to_count
```
casey, line 402:
```suggestion
    let blessed_inscription_count = statistic_to_count
```

src/subcommand/server.rs
```diff
@@ -994,9 +994,13 @@ impl Server {
       )
     };
 
-    let previous = index.get_inscription_id_by_inscription_number(entry.number - 1)?;
+    let previous = if entry.sequence_number == 0 {
```
casey, line 997: I would do a checked sub here:

```
if let Some(n) = entry.sequence_number.checked_sub(1) {
...
```

casey (changes requested):
LGTM except for comments

casey (approved):

BitTogether:
Please add this, we need to have one numbering system that includes both sets of ordinals without the positive negative gimmick thing.

PR #2464, opened 2023-09-21
Add font/woff2 media type with extension .woff2

devords:
This would make it easier for people to inscribe custom fonts.


elocremarc:
Should we add more types?

devords:
@elocremarc Are you thinking about any specific file types? AFAIK, WOFF 2.0 is the most widely accepted standard for fonts today. So, maybe makes sense to start with that and add other font types if there's any demand.

elocremarc:
Maybe .ttf ?

devords:
Added .otf, .ttf and .woff as well. Might as well just add them all while at it.

raphjaph (approved):

PR #2466, opened 2023-09-22
fix  issue with `--satpoint` when offset > 0. (Fix #2465)

### LINKED ISSUE #2465: `--satpoint` flag not working on certain condition
using `--satpoint=<SatPoint>` when `SatPoint.offset > 0` is not working. Example:
```
$ ord wallet inscribe --satpoint=7b70c5d9e7ea5e42d93a3bcf118b2adfd5eeb7592414006c8285cea8984c9d8b:2:1000 --fee-rate=1.5 file.txt

error: Failed to send reveal transaction
because: JSON-RPC error: RPC error response: RpcError { code: -26, message: "non-mandatory-script-verify-flag (Invalid Schnorr signature)", data: None }
```

felipelincoln:
the prevouts being used to construct the reveal signature is:
```rust
let mut prevouts = vec![unsigned_commit_tx.output[0].clone()];

if let Some(parent_info) = parent_info {
  prevouts.insert(0, parent_info.tx_out);
}
```
but using `unsigned_commit_tx.output[0]` as one of its elements is a wrong assumption since a inscribe run can be called using `--satpoint` with offset > 0, which would make the first output of the commit transaction to be a change (part not used of the satpoint), and not the output encoding the taproot spending condition.

Example of a inscribe using `--satpoint=cad80aa153050fa300fc539e35fbb2046f4c99d3e09b27437db2232bfec43f71:2:1000`
image
(example produced building ord with the code from this pr. https://mempool.space/pt/testnet/tx/aab6b689ec23b60e0943adf6357eb3a2cdb8c50ebfaf10cb972f96155a239f4b#flow=&vout=1)
In this case, `unsigned_commit_tx.output[0]` is the first output with `1000sat` value

---

this pr fix #2465

### DIFF
--- a/src/subcommand/wallet/inscribe.rs
+++ b/src/subcommand/wallet/inscribe.rs
@@ -187,9 +187,11 @@ impl Inscribe {
 
     let commit = client.send_raw_transaction(&signed_commit_tx)?;
 
-    let reveal = client
-      .send_raw_transaction(&signed_reveal_tx)
-      .context("Failed to send reveal transaction")?;
+    let reveal = match client
+      .send_raw_transaction(&signed_reveal_tx) {
+        Ok(txid) => txid,
+        Err(err) => return Err(anyhow!("Failed to send reveal transaction: {err}\nCommit tx {commit} will be recovered once mined"))
+      };
 
     Ok(Box::new(Output {
       commit,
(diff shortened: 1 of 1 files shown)


raphjaph (approved):
Nice, good catch! 

Pro tip: Create a feature branch on your repo with the changes and open a PR with that. Makes development for you easier

PR #2473, opened 2023-09-25
Fixes release tarbomb

uzyn:
Packages it similar to how bitcoin release does it, avoids making a tarbomb.


casey:
Nice, looks good to me!

PR #2477, opened 2023-09-26
Inscribe directly to an Opendime

raphjaph:
- [x] test should write dummy `address.txt` to tmp dir
- [ ] add ability to send to opendime (not only inscribe)
- [ ] test on mainnet
- [ ] more checks and asserts
- [ ] check that Opendime is sealed
- [ ] ability to import unsealed Opendime and send with wallet
  - [ ] should run through all checks that inscription not sent in fee
  - [ ] should be able to add inputs if necessary
  - [ ] for now refuse to send if more than one output/inscription on Opendime
 - [ ] would also be cool to save the actual inscription on the Opendime since it is a USB drive
 - [ ] ability to show inscriptions on opendime (import to ord as watch-only)


rot13maxi:
There’s already a destination flag for the inscribe command (added by some handsome genius), what are you going for that isnt already covered? You can already inscribe straight to an opendime/satscard. 

Is the idea to add deeper integration or something else?

raphjaph:
I liked the UX of just plugging in a fresh Opendime and it generating the address directly instead of manually dragging files into it. Also want to add support to plug in an unsealed Opendime, importing the privkey and sending it off with the wallet.

rot13maxi:
Dope

PR #2479, opened 2023-09-27
Fix unbound outpoint server error

raphjaph:
When trying to look at `/output/0000000000000000000000000000000000000000000000000000000000000000:0` it gets internal server error. 

Error has to do with Bitcoin Core but can't seem to figure out where we are calling Bitcoin Core.


casey (approved):
LGTM!

PR #2483, opened 2023-09-30
Filipino Version of the Handbook

jcatama:
Filipino Version of the Handbook
Locally Tested
We originally published this translation on the [gamma.io website](https://twitter.com/trygamma/status/1697809366513336686)


raphjaph (approved):
Looks good to me! Did some spot checks with google translate. Also since gamma.io has the same translation going to trust that as well.

PR #2491, opened 2023-10-02
Add rune index

casey:
*NB: This implementation is not final and you are guaranteed to get rekt if you use it on mainnet."

This PR adds a rune index to `ord` behind the `--index-runes` flag, and only when not on mainnet.

All discussion of runes should be limited to this PR or the [runestone](https://github.com/casey/runestone/) repo, since early development and design will likely generate a lot of discussion, and it would be nice to keep it out of the `ord` repo for now.


src/index/updater/rune_updater.rs
```diff
@@ -0,0 +1,189 @@
...
+        Some(etching) => {
+          // If the issuance symbol is already taken, the issuance is ignored
+          if self.rune_to_id.get(etching.rune.0)?.is_some() {
+            None
+          } else {
+            // Construct an allocation, representing the new runes that may be
+            // allocated.
+            match u16::try_from(index) {
+              Ok(index) => Some(Allocation {
+                id: (height as u128) << 16 | index as u128,
```
darioush, line 81: ❤️

rot13maxi:
The ordinals ecosystem has a tendency of taking things with big “this is a quick hack, do NOT use this in production” signs on them and building products and services on top (cursed inscriptions on demand to try to get early negative numbers was a fun recent episode). As much as I *LOVE* the way this ecosystem says “whatever, nerd!” and does it anyway, i think we should be VERY deliberate about creating any dependency between ord and runes.

rot13maxi:
If this is a matter of “ord already has indexing, i want to reuse that”, maybe fork ord?

src/runes/runestone.rs
```diff
@@ -0,0 +1,501 @@
+use super::*;
```
rot13maxi, line 1: Would be awesome to have runestone as its own crate so you can depend on it without all of ord

src/index/updater/rune_updater.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    // A vector of allocated transaction output rune balances
+    let mut allocated: Vec<HashMap<u128, u128>> = vec![HashMap::new(); tx.output.len()];
+
+    if let Some(runestone) = runestone {
+      // Determine if this runestone conains a valid issuance
+      let mut allocation = match runestone.etching {
+        Some(etching) => {
+          // If the issuance symbol is already taken, the issuance is ignored
```
rot13maxi, line 69: I think unique symbols is a bad idea. A couple scenarios where this can go awry: 

- bug (or new feature) in the indexer means that previously-unrecognized issuances are now valid for new indexes
- new token issuances can be front-run and nullified either by bots watching the mempool and outbidding them, or by miners directly

src/runes/edict.rs
```diff
@@ -0,0 +1,8 @@
+use super::*;
+
+#[derive(Default, Serialize, Debug, PartialEq)]
+pub struct Edict {
```
rot13maxi, line 4: nice

src/runes/runestone.rs
```diff
@@ -0,0 +1,501 @@
...
+      let (integer, length) = varint::decode(&payload[i..])?;
+      integers.push(integer);
+      i += length;
+    }
+
+    let mut edicts = Vec::new();
+    let mut etching = None;
+
+    for chunk in integers.chunks(3) {
+      match *chunk {
```
rot13maxi, line 32: maybe think about reifying the `[rune]`, `[rune, decimals]`, `[id, amount, output]` variants here into an enum?

src/runes/varint.rs
```diff
@@ -0,0 +1,130 @@
+use super::*;
+
+#[cfg(test)]
+pub fn encode(n: u128) -> Vec<u8> {
```
rot13maxi, line 4: should `encode` only be for tests?

src/runes/runestone.rs
casey: I'm definitely open to pulling it out later.

src/index/updater/rune_updater.rs
casey: I'm really of two minds of this. You're right about the two issues you mention, along with the fact that if a project has some token name that they want, they have to come up with a new one if it's already taken.

However, unique symbols is pretty cool. Having short, human readable names you can look things up by is nice, and having a symbol, if there are duplicates, seems pretty worthless to m.

Also, duplicate symbols make fake token scams easier. There have been a bunch of scams where someone has been sent, for example, a fake USDT token, and treated it like real tether, and sent eth in return.

src/runes/varint.rs
casey: Currently it's only for tests, since the non-test code uses `encode_to`, which is more efficient.

src/runes/runestone.rs
casey: I actually think this is probably okay, just because it makes it clear how it works, i.e., extra values are treated as rune, or rune and decimals, and adding an enum might make it less clear.

casey:
> The ordinals ecosystem has a tendency of taking things with big “this is a quick hack, do NOT use this in production” signs on them and building products and services on top (cursed inscriptions on demand to try to get early negative numbers was a fun recent episode). As much as I LOVE the way this ecosystem says “whatever, nerd!” and does it anyway, i think we should be VERY deliberate about creating any dependency between ord and runes.

> If this is a matter of “ord already has indexing, i want to reuse that”, maybe fork ord?

Definitely guilty as charged. A few things:

- Ord already has indexing. Forking ord and ripping out ordinals and
  inscriptions would be a lot of work, and maintaining a separate binary would
  be a lot of work. Ord also runs on `ordinals.com`, so getting runes up there
  would also be nice. The rune indexing is pretty simple (famous last words),
  it's behind an `--index-runes` flag, and there isn't any coupling between
  ordinals and runes.

- Adding metadata via an inscription envelope would require some code from
  `ord`, so that's convenient.

I definitely see the argument. Honestly, I don't know if it's personally worth
it for me to write runes if it has to live in a separate binary, with all that
entails. Now of course, maybe that's an argument not to do Runes at all.

src/index/updater/rune_updater.rs
summraznboi: Can we have unique symbols, but also make symbols not a requirement? Meaning that if a rune declares a symbol already in use or just doesn't declare a symbol by using empty string, if can only be referenced by its ID (which should be a number so as to have no possibility of colliding with symbol space)?

rot13maxi: > Also, duplicate symbols make fake token scams easier. There have been a bunch of scams where someone has been sent, for example, a fake USDT token, and treated it like real tether, and sent eth in return.

maybe the better solution here is to make it easier for an issuer to share an output/pubkey/etc in a format that’s consumed by wallets. If Tether can vend a json file that says “here are my issuances”, and wallets can consume that, then wallets can say “this is tether”. That blob of json could be served up on a well known url path, or via a dns record or something else. Just spitballing. 

We have the same issue with inscriptions: if someone issues a collection, anyone can make an inscription with the same artwork. So we need some offchain declaration from the artist (heres a link to the inscription) or a parent-child collection (but you need some way to get the authenticity of the parent)

t4t5: Unique symbols will make people rush in to squat anything remotely good, just like with BRC-20. It will diminish the chance of real USDT/USDC ever being issued on Runes imo

I think frauds around duplicate symbols is something that can be resolved on a higher layer. There could be public lists of notable token IDs shared by companies/DEXes/indexers, and they can show a little verified checkmark or something in the UI

casey:
Still need to review:
- runestone::from_transaction
- varint::decode

Tests we reviewed:
- [x] etching without edict creates no rune
- [x] etching with one edict creates rune with correct supply
- [x] check that first rune in block is uncommon
- [x] cannot allocate a rune below minimum for block
- [x] runes have correct symbol
- [x] runes have correct divisibility
- [x] edicts can't allocate more runes than new etching maximum
- [x] etching with one edict creates rune with correct supply which is less than max
- [x] etching with two edicts creates rune with correct supply
- [x] edict with invalid outputs are skipped
- [x] edicts correctly transfer new runes to ouputs
- [x] edicts correctly transfer input runes to ouputs
- [x] unallocated runes are assigned to first non-op-return output when op return is not first output
- [x] can't allocate a duplicate rune
- [x] runes get correct id
- [x] unallocated runes are assigned to first non-op-return output when op return is first output
- [x] check that second rune in block is common

Tests we need to review:
- [ ] edicts which refer to id 0 but are not in a tx with an etching are skipped
- [ ] edicts which refer to a rune with no balance are skipped
- [ ] edicts can't transfer more runes than are in inputs
- [ ] edicts which transfer 0 runes don't create balance entries
- [ ] balance entries for outputs with no runes of transaction with runestone are not present
- [ ] edicts may explicitly transfer runes to op_return outputs

raphjaph:
Went through the following files:

- rune_updater.rs
- edict.rs
- rune.rs
- runes.rs

Functionality checked:

- [x] etching without edict creates no rune
- [x] etching with one edict creates rune with correct supply
- [x] check that first rune in block is uncommon
- [x] cannot allocate a rune below minimum for block
- [x] runes have correct symbol
- [x] runes have correct divisibility
- [x] edicts can't allocate more runes than new etching maximum
- [x] etching with one edict creates rune with correct supply which is less than max
- [x] etching with two edicts creates rune with correct supply
- [x] edict with invalid outputs are skipped
- [x] edicts correctly transfer new runes to ouputs
- [x] edicts correctly transfer input runes to ouputs
- [x] unallocated runes are assigned to first non-op-return output when op return is not first output
- [x] can't allocate a duplicate rune
- [x] runes get correct id
- [x] unallocated runes are assigned to first non-op-return output when op return is first output
- [x] check that second rune in block is common

Functionality still to check (not complete list):

- [ ] runestone::from_transaction
- [ ] varint::decode

Write tests for these:

- [ ] edicts which refer to id 0 but are not in a tx with a etching are skipped
- [ ] edicts which refer to a rune with no balance are skipped
- [ ] edicts can't transfer more runes than are in inputs
- [ ] edicts which transfer 0 runes don't create balance entries
- [ ] balance entries for outputs with no runes are not present
- [ ] edicts may explicitly transfer runes to op_return outputs

Further questions:

- should rune id be renamed? since it doesn't refer to a transaction id?

src/runes/rune.rs
```diff
@@ -0,0 +1,127 @@
...
+
+  fn from_str(s: &str) -> crate::Result<Self> {
+    let mut x = 0u128;
+    for (i, c) in s.chars().enumerate() {
+      if i > 0 {
+        x += 1;
+      }
+      x *= 26;
+      match c {
+        'A'..='Z' => {
```
summraznboi, line 52: We need a mode to display the rune in a runic alphabet for the memes, we can use this one easily since it is also 26 characters: https://en.wikipedia.org/wiki/Anglo-Saxon_runes

raphjaph (approved):
🚀💥

sondotpin:
- [ ] etch is incorrect syntax `OP_RETURN aaaaaa bbbbbb ccccc` is valid?
- [ ] edict with invalid rune from inputs are skipped or not
- [ ] ord wallet etch + ord wallet transfer
- [ ] number of BTC to register characters

@casey

casey:
@sondotpin This is an initial, incomplete implementation. Feel free to open issues on the https://github.com/casey/runestone/ repo, but do a search for dupes first, there are already a bunch of issues there.

PR #2499, opened 2023-10-05
Implement pointer spec

raphjaph:
TODO:
~~- [ ] move logic into `update_inscription_location()`~~
- [x] refactor spaghetti if else
- [x] many more tests
- [x] handle reinscriptions 
- [x] handle cursedness
- [x] trailing zeros

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -10,6 +10,7 @@ pub(crate) const PROTOCOL_ID: [u8; 3] = *b"ord";
 
 pub(crate) const BODY_TAG: [u8; 0] = [];
 pub(crate) const CONTENT_TYPE_TAG: [u8; 1] = [1];
+pub(crate) const POINTER_TAG: [u8; 1] = [2];
 pub(crate) const PARENT_TAG: [u8; 1] = [3];
 pub(crate) const METADATA_TAG: [u8; 1] = [5];
 pub(crate) const METAPROTOCOL_TAG: [u8; 1] = [7];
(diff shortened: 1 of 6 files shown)


raphjaph:
Went through the following files:

- rune_updater.rs
- edict.rs
- rune.rs
- runes.rs

Functionality checked:

- [x] etching without edict creates no rune
- [x] etching with one edict creates rune with correct supply
- [x] check that first rune in block is uncommon
- [x] cannot allocate a rune below minimum for block
- [x] runes have correct symbol
- [x] runes have correct divisibility
- [x] edicts can't allocate more runes than new etching maximum
- [x] etching with one edict creates rune with correct supply which is less than max
- [x] etching with two edicts creates rune with correct supply
- [x] edict with invalid outputs are skipped
- [x] edicts correctly transfer new runes to ouputs
- [x] edicts correctly transfer input runes to ouputs
- [x] unallocated runes are assigned to first non-op-return output when op return is not first output
- [x] can't allocate a duplicate rune
- [x] runes get correct id
- [x] unallocated runes are assigned to first non-op-return output when op return is first output
- [x] check that second rune in block is common

Functionality still to check (not complete list):

- [ ] runestone::from_transaction
- [ ] varint::decode

Write tests for these:

- [ ] edicts which refer to id 0 but are not in a tx with a etching are skipped
- [ ] edicts which refer to a rune with no balance are skipped
- [ ] edicts can't transfer more runes than are in inputs
- [ ] edicts which transfer 0 runes don't create balance entries
- [ ] balance entries for outputs with no runes are not present
- [ ] edicts may explicitly transfer runes to op_return outputs

Further questions:

- should rune id be renamed? since it doesn't refer to a transaction id?

casey (approved):
LGTM 😭😭😭😭😭😭😭😭😭😭😭😭😭😭😭

PR #2504, opened 2023-10-08
Batch inscriptions

raphjaph:
`ord wallet batch-inscribe inscriptions.yaml`

`inscriptions.yaml` should include:
- [x] path to files
- [x] metadata to include for each file
- [x] should set parent
- [x] fee-rate

Additional cool stuff:
- [ ] destination
- [ ] chain commit/reveal if a single transaction over size limit
- [ ] save state of batch in `state.yaml`


goodwinmark:
👁

raphjaph:
@casey, @terror, @veryordinally this is ready for an initial review.

I wrote some high-level integration tests, but still need a lot of unit tests.

Additionally, I haven't yet implemented the different modes.

dannydeezy:
could this eventually support choosing which satpoints to inscribe on for each of the inscriptions in the batch?

raphjaph:
yeah for sure! Want to keep it as simple as possible for now but we can do really cool stuff with this

khademreza688 (approved):

raphjaph:
TODO:
- [ ] parent info should contain inscription id
- [ ] conflicting flags tests
- [ ]

casey (approved):
WHY NOT

PR #2523, opened 2023-10-11
Make inscriptions with pointer cursed

raphjaph:


casey:
Nice, looks good to me.

casey (approved):

casey:
I see a change to ord.service, is that intentional?

veryordinally:
> I see a change to ord.service, is that intentional?

No, that should be temporary for the dev server - needs to be removed from this PR

PR #2531, opened 2023-10-13
Allocating zero amount means allocate all remaining

casey:


raphjaph (approved):

Ifxscript:
whats your say on pipe protocol rodarmor i think the protocol is pretty cool too

PR #2532, opened 2023-10-13
Delta encode edict ids

casey:
This also includes #2529, and you can see the third commit shows the size reductions, which are really nice. Will remove #2529 prior to merging.


raphjaph (approved):

raphjaph:
much savings, much wow!

PR #2537, opened 2023-10-14
Create single-directory release archives

casey:
This puts the files in release archives inside a directory, so that when you extract them, they stay within a directory. This should be merged right before doing a release, so that the next release archive is compatible.


casey:
Tested this locally by building a release tarball with the new bin/package, and installing it with install.sh.

PR #2538, opened 2023-10-14
add python media type

elocremarc:
Add file type python closes  https://github.com/ordinals/ord/issues/1481

### DIFF
--- a/static/preview-code.js
+++ b/static/preview-code.js
@@ -1,21 +1,21 @@
 import hljs from 'https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11/build/es/highlight.min.js';
-import javascript from 'https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11/build/es/languages/javascript.min.js';
-import yaml from 'https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11/build/es/languages/yaml.min.js';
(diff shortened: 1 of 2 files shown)


elocremarc:
@raphjaph I was realizing after I submitted this that the mime-type returned for the highlight code JS is gonna be `text/x-python` This won't be recognized as a type for highlight Js. Do you think it would be worth making a more generic highlight JS template with auto highlight to support adding executable code like rust or say typescript? We have the optimized one already where we specify the languge for code types that are used in inscriptions. However any other code that is just reference code using ord as a CDN we would fallback to auto highlight these ones won't be rendered nearly as often.

elocremarc:
@raphjaph  got highlight working with latest commit.
image

static/preview-code.js
```diff
@@ -3,16 +3,20 @@ import javascript from 'https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11/b
...
 hljs.registerLanguage('css', css);
 hljs.registerLanguage('json', json);
+hljs.registerLanguage('python', python);
 
 const inscription = document.documentElement.dataset.inscription;
 const response = await fetch(`/content/${inscription}`);
 const contentType = response.headers.get("content-type");
-const language = contentType.split("/")[1];
+let language = contentType.split("/")[1];
+if (language.includes("python")) {
```
casey, line 18:
```suggestion
if (language === "x-python") {
```

casey (approved):
LFG

PR #2546, opened 2023-10-15
Improve minimum at height

casey:


raphjaph (approved):

casey:
It's used in tests, and it's a rune which is always above the minimum, no matter the height.

PR #2552, opened 2023-10-17
Refactor rune assertions

casey:
This is a little more compact, and because the balances and runes are both sorted before comparison, it won't be insensitive to txids reordering, which happens when the runestone changes.


casey:
Doing this as part of #2548.

PR #2555, opened 2023-10-17
dependency purification

boppleton:
this pr attempts to replace some of the CDN imports used by the explorer with links to inscribed libraries.

the inscribed code is an exact copy from the original CDN links for easy SHA256 checksum verification, except some just have extra comments added to the top for copyright/licensing/etc

### DIFF
--- a/docs/theme/index.hbs
+++ b/docs/theme/index.hbs
@@ -63,7 +63,7 @@
 
         {{#if mathjax_support}}
         <!-- MathJax -->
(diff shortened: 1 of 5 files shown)


casey:
This is a super cool idea! I think there are a number of downsides though:

- We would have to review the inscribed files to make sure they're the same as the existing dependencies.
- If an index hasn't yet indexed the inscribed dependencies, the previews that refer to those inscriptions won't work.
- If we want to update a dependency, we'd have to reinscribe it.

Honestly though, it's a rad idea.

Issue #2524, opened 2023-10-12
Add a collections page


casey:
I'd like to be able to see in one place all the inscriptions which are using genetic provenance. "collections" might not be the right word, but something like `/collections`, where all collections are shown, in order of how recently a child was added to the collection.

devords:
Probably a good idea to paginate the children on inscription page too, otherwise it'll try to load all the children on the same page. Maybe, the children list could be capped at 10 along with a link to this new collections page.

a-edakin:
> Probably a good idea to paginate the children on inscription page too, otherwise it'll try to load all the children on the same page. Maybe, the children list could be capped at 10 along with a link to this new collections page.

Yep, and that is exactly a problem I encountered 

`curl --location --request GET 'http://<testnet>/inscription/4728a9610731cc4ddcc7285d3c52aa535ba73ca4400b34d002bce703612e3c68i0' --header 'Accept: application/json'`

`http://<testnet>/inscription/4728a9610731cc4ddcc7285d3c52aa535ba73ca4400b34d002bce703612e3c68i0`

This collection has over 2000+ children and page takes forever to load just like JSON API

ordinalOS:
Good work team. Please consider a "Network Graph" style display here: https://ordinalos.github.io/bussy-tree/

raphjaph:
Closed by #2561

PR #2561, opened 2023-10-19
Implement /collections Page

veryordinally:
Implements #2524

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -10,12 +10,12 @@ use {
     page_config::PageConfig,
     runes::Rune,
     templates::{
-      BlockHtml, BlockJson, BlocksHtml, ChildrenHtml, ClockSvg, HomeHtml, InputHtml,
-      InscriptionHtml, InscriptionJson, InscriptionsBlockHtml, InscriptionsHtml, InscriptionsJson,
-      OutputHtml, OutputJson, PageContent, PageHtml, PreviewAudioHtml, PreviewCodeHtml,
-      PreviewImageHtml, PreviewMarkdownHtml, PreviewModelHtml, PreviewPdfHtml, PreviewTextHtml,
-      PreviewUnknownHtml, PreviewVideoHtml, RangeHtml, RareTxt, RuneHtml, RunesHtml, SatHtml,
-      SatJson, TransactionHtml,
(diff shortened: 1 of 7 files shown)


veryordinally:
So far just a quick proof of concept that displays all parent inscriptions in random order. Probably need an index to efficiently show the parents in the order suggested in #2524 so will require upgrade to database schema version. Also obviously required pagination, and it would make sense to add a /collection page as well.

raphjaph:
TODO:
- [ ] add pagination
- [ ] test that the page only shows 100 inscriptions/collections
- [ ] test prev and next

src/subcommand/server.rs
```diff
@@ -1236,6 +1238,36 @@ impl Server {
...
+    Self::collections_paginated(Extension(page_config), Extension(index), Path(0)).await
+  }
+
+  async fn collections_paginated(
+    Extension(page_config): Extension<Arc<PageConfig>>,
+    Extension(index): Extension<Arc<Index>>,
+    Path(page_index): Path<u64>,
+  ) -> ServerResult<Response> {
+    let (collections, more_collections) =
+      index.get_collections_paginated(100, page_index.try_into().unwrap())?;
```
casey, line 1254: This needs to return an error if it fails, since otherwise on a 32 bit platform, the try into could fail and it could crash.

src/subcommand/server.rs
```diff
@@ -4153,4 +4263,53 @@ mod tests {
       ),
     );
   }
+
+  #[test]
+  fn collections_page_shows_children() {
```
casey, line 4268: I don't think I understand the purpose of this test vs the other test. Could this one be removed?

casey (changes requested):
Left a few comments, otherwise looks great.

src/subcommand/server.rs
raphjaph: Oh yeah this is extraneous now, doing this in the other one

casey (approved):
I just realized that we can avoid integer conversions if we just accept a usize argument, so I pushed that change. LGTM!

raphjaph (approved):

casey:
I added the collections page to the homepage too, so it's in the tabs.

PR #2570, opened 2023-10-21
Guilt by association

### LINKED ISSUE #2569: Ord 0.9.x inscription numbers different than 0.7.x and 0.8.x
Inscriptions [35321413](https://ordinals.com/inscription/adc8d37dbce946cf90c1b6c106b8b8404646a7189bb05cede7e40984bef3af5di0) and [35329860](https://ordinals.com/inscription/a1b05b551557ef7365626e561e788ee965fb53f932cebd9b02ff553a8533b096i0) are indexed by ord 0.9.0 but not by ord 0.7.0 or 0.8.0. As a result, 0.9.0 is assigning numbers that are off by 2 from those assigned by older versions.

Both affected inscriptions are the first inscription in a "pointer" test inscription, where no pointer is specified for the first envelope in the witness (since the default offset of 0 is correct and doesn't need a pointer).

casey:
Attempt to fix #2569. In the old, pre 0.9.0 parser, if an inscription was followed by an inscription that caused an error, earlier inscriptions would not be processed. The things that would trigger this were unrecognized even fields, duplicate fields, and incomplete fields. This adds a new curse, GuiltByAssociation, which which causes an inscription which is valid, but is followed by an inscription which triggers one of the errors to be cursed. I'd be mad but honestly this is kind of hilarious at this point.

### DIFF
--- a/deploy/ord.service
+++ b/deploy/ord.service
@@ -13,7 +13,7 @@ ExecStart=/usr/local/bin/ord \
   --data-dir /var/lib/ord \
   --config-dir /var/lib/ord \
   --chain ${CHAIN} \
-  --index-sats \
+#  --index-sats \
   server \
   --acme-contact mailto:casey@rodarmor.com \
   --http \
--- a/justfile
+++ b/justfile
@@ -70,15 +70,16 @@ profile-tests:
 fuzz:
   #!/usr/bin/env bash
   set -euxo pipefail
-
   cd fuzz
-
   while true; do
     cargo +nightly fuzz run transaction-builder -- -max_total_time=60
     cargo +nightly fuzz run runestone-decipher -- -max_total_time=60
     cargo +nightly fuzz run varint-decode -- -max_total_time=60
     cargo +nightly fuzz run varint-encode -- -max_total_time=60
   done
+
+decode txid:
+  bitcoin-cli getrawtransaction {{txid}} | xxd -r -p - | cargo run decode
 
 open:
   open http://localhost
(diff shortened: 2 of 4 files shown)


casey:
I haven't tested this, and won't be able to get to it until some time this weekend, or even monday, so if anyone wants to build an index with this patch and confirm that it reports the same inscription numbers as 0.8.0, that would be much appreciated.

sanjfomojis:
You’re quick!! 🔥

Building the index now so will keep you posted

crypticmeta:
this update will need a new index?

devords:
#2504 seems to be skipping pointer for the 1st inscription when `batch.yml` does not specify any parent. Example, [35360664](https://ordinals.com/inscription/962e8b7c237a64a344bb5c6d82854a19a8546c2532e820833b656561ebcaf6e6i0)

Pointer is included when there's a [parent](https://ordinals.com/inscription/11d3f4b39e8ab97995bab1eacf7dcbf1345ec59c07261c0197e18bf29b88d8dai0).

devords:
Will this change bring inscription numbering in line with 0.8.x? Will [35321413](https://ordinals.com/inscription/adc8d37dbce946cf90c1b6c106b8b8404646a7189bb05cede7e40984bef3af5di0), and others alike, become cursed?

casey:
Yes, that's correct. 35321413, and other inscriptions which share a witness with a cursed inscription, will become cursed.

leonidasord:
After talking with many people over the past 24 hours here are my two sats on this: https://github.com/ordinals/ord/issues/2569#issuecomment-1774228502

casey:
The best thing to do seems to be to not make this change, and consider 0.9.0 as the source of truth going forward. Check out [this tweet](https://twitter.com/rodarmor/status/1716284984800297380) and #2569 for the discussion.

PR #2581, opened 2023-10-24
Requre --batch or --file for `ord wallet inscribe`

### LINKED ISSUE #2579: Error inscribing on testnet
When trying to inscribe on testnet using

ord -t wallet inscribe --fee-rate "1"

i get:
thread 'main' panicked at src/subcommand/wallet/inscribe.rs:150:27:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

when i add a filename at the end  to use to inscribve i get

ord -t wallet inscribe --fee-rate "1" file.txt
error: unexpected argument 'file.txt' found

Usage: ord wallet inscribe [OPTIONS] --fee-rate <FEE_RATE>

Can anyone tell me why neither of these work?

casey:
Fixes #2579.


elocremarc:
Can you default it to fallback to file if no flag is specified? This would not cause scripts people made to break that used the old syntax.

raphjaph (approved):

PR #2584, opened 2023-10-25
Update redb recovery detection to use public API

cberner:
This requires the new redb API in https://github.com/cberner/redb/pull/700

Lemme know if this looks good to you and I'll add that API to redb.


casey:
Nice! Changed it to use `Once` to check if the message had already been printed. Also, you can use `Path::display` to print paths.

PR #2604, opened 2023-10-27
Add metadata recursive endpoint

### LINKED ISSUE #2603: Feature Request: recursive endpoint to read metadata
It would be really handy if an inscription could hit an endpoint to read its own metadata.

rot13maxi:
closes #2603

quick little endpoint. open to change the path, or whatever other feedback folks have :)

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -231,6 +231,7 @@ impl Server {
         .route("/static/*path", get(Self::static_asset))
         .route("/status", get(Self::status))
         .route("/tx/:txid", get(Self::transaction))
+        .route("/r/metadata/:inscription_id", get(Self::metadata))
         .layer(Extension(index))
         .layer(Extension(page_config))
         .layer(Extension(Arc::new(config)))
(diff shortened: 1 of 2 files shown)


elocremarc:
Oh nice!
@raphjaph lets just yeet into /-/ and get the endpoints moving.

@rot13maxi can you test out if this works on regtest via recursion by adding the CSP-Headers like Raphs pr? Wonder if we should all rebase off his?

src/subcommand/server.rs
```diff
@@ -694,6 +695,51 @@ impl Server {
     )
   }
 
+  async fn metadata(
+    Extension(index): Extension<Arc<Index>>,
+    Extension(config): Extension<Arc<Config>>,
+    Path(inscription_id): Path<InscriptionId>,
+  ) -> ServerResult<Response> {
```
elocremarc, line 702: Should return json
```diff
-  ) -> ServerResult<Response> {
+ ) -> ServerResult<Json<String>> {
```

elocremarc, line 702: Response should be type Json?

casey:
I think for metadata, we should return the CBOR. Converting the CBOR to JSON makes the details of that conversion part of the contract of the endpoint. The mapping from CBOR to JSON isn't well defined, and I'm not sure how serde would handle types in the CBOR which have no JSON equivalent. It sucks to make inscriptions deserialize CBOR, but someone can always inscribe a CBOR parser, and everyone can use that, and when something goes wrong, we can blame the library author instead of us 😂

For the response, I would just return a string with the hex-encoded CBOR, since all recursive endpoint responses must be JSON.

rot13maxi:
Hilariously I originally just returned cbor, but then i saw raph was trying to jsonify all the things so changed it. Im happy to change it back and make deserialzation someone else’s problem.

devords:
Instead of metadata specific endpoint, how about a recursive endpoint for inscription that includes metadata among other things like current owner, sat location etc? Basically, whatever we can include without needing a full sat index.

raphjaph (approved):
LGTM!

PR #2608, opened 2023-10-29
Speed up indexing of re-inscriptions

SmarakNayak:
When logging re-inscription info, use already calculated `next_sequence_number` rather than calling `id_to_entry.len()`

On my PC (5600x, 32GB) speeds up indexing by ~40s per re-inscription.


casey (approved):

raphjaph (approved):
Nice catch!

PR #2617, opened 2023-10-31
Paginate children inscriptions

### LINKED ISSUE #2578: Inscription page should paginate children
Currently, the inscription page tries to render all the children without any pagination. With the new batch inscription, this will result in attempting to load 100s of children on the same page.

I know this is being handled in #2561 as well. But, probably needs to be fixed soon now that 0.10.0 is released.

devords:
Fixes #2578


src/subcommand/server.rs
```diff
@@ -1172,6 +1184,37 @@ impl Server {
...
+    )
+    .await
+  }
+
+  async fn inscription_children_from_pages(
+    Extension(page_config): Extension<Arc<PageConfig>>,
+    Extension(index): Extension<Arc<Index>>,
+    Path((inscription_id, page_index)): Path<(InscriptionId, usize)>,
+  ) -> ServerResult<Response> {
+    let _inscription = index
```
devords, line 1205: @raphjaph Wondering if there's better way to check for the validity of an inscription.

src/subcommand/server.rs
```diff
@@ -1127,6 +1136,8 @@ impl Server {
     let next = index.get_inscription_id_by_sequence_number(entry.sequence_number + 1)?;
 
     let children = index.get_children_by_inscription_id(inscription_id)?;
+    let end_index = usize::min(children.len(), 20);
```
devords, line 1139: Which is the best place to store the per-page count as a constant?

raphjaph, line 1139:
```suggestion
    let end_index = usize::min(children.len(), 100);
```

raphjaph: I think just putting it here is fine, we do the same of the normal inscriptions pagination.

src/subcommand/server.rs
raphjaph: Do you mean if it exists or if the inscription id is malformed? This will return a 404 if it's not found and a 400 if the id is malformed.

raphjaph: Ah I see what you mean. You could delete the `let _inscription....` part and check wheter the `let children...` returns and empty array and do a 404 but then it would be unclear of the parent inscription doesn't exist or just doesn't have any children. So I think this should be fine for now

src/subcommand/server.rs
```diff
@@ -1172,6 +1184,37 @@ impl Server {
...
+  async fn inscription_children_from_pages(
+    Extension(page_config): Extension<Arc<PageConfig>>,
+    Extension(index): Extension<Arc<Index>>,
+    Path((inscription_id, page_index)): Path<(InscriptionId, usize)>,
+  ) -> ServerResult<Response> {
+    let _inscription = index
+      .get_inscription_by_id(inscription_id)?
+      .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
+
+    let children = index.get_children_by_inscription_id(inscription_id)?;
```
raphjaph, line 1209: Instead of getting all children, let's try to only get only the number we need (100). Have a look how we do the inscriptions pagination, specifically `fn inscriptions_inner()` in `server.rs` and `fn get_latest_inscriptions_with_prev_and_next` in `index.rs`. You will have to use the `SEQUENCE_NUMBER_TO_INSCRIPTION_ID` table to get the children in the right order since the redb multimap is unordered (at least in the sense we care about). If this is too daunting we can also show this/do it together on the coding stream tomorrow.

src/templates/inscription_children.rs
```diff
@@ -0,0 +1,101 @@
...
+  ) -> Result<Self> {
+    let num_children = children.len();
+
+    let start = page_index * 20;
+    let end = usize::min(start + 20, num_children);
+
+    if start > num_children || start > end {
+      return Err(anyhow!("page index {page_index} exceeds inscription count"));
+    }
+    let children = children[start..end].to_vec();
```
raphjaph, line 25: Move this logic into the function I proposed in the above comment.

src/subcommand/server.rs
devords: Makes sense!

src/subcommand/server.rs
devords: Would be great to do this on the stream!

casey:
Did some refactoring during coding club, this is just missing some testss:

server.rs tests:
- test for fetching /children without page
- test for fetching /children with page
- only first four children are displayed on /inscription
- if more than four children are displayed, display more button

raphjaph:
@devords I just finished up the last tests. Have a look if this looks good.

src/index.rs
```diff
@@ -795,6 +796,45 @@ impl Index {
       .collect()
   }
 
+  pub(crate) fn get_children_by_inscription_id_with_limit(
```
casey, line 799: Let's merge these two functions. We can add a page size argument to `get_children_by_inscription_id_paginated`, and request the first page with page size 4.

src/templates/children.rs
```diff
@@ -0,0 +1,71 @@
...
+  pub(crate) parent: InscriptionId,
+  pub(crate) parent_number: i64,
+  pub(crate) children: Vec<InscriptionId>,
+  pub(crate) prev_page: Option<usize>,
+  pub(crate) next_page: Option<usize>,
+}
+
+impl PageContent for ChildrenHtml {
+  fn title(&self) -> String {
+    format!("Children of Inscription {}", self.parent_number)
```
casey, line 14:
```suggestion
    format!("Inscription {} Children", self.parent_number)
```

casey: Also in header

templates/children.html
```diff
@@ -0,0 +1,22 @@
+<h1 class=light-fg>Children of <a href=/inscription/{{ self.parent }}>Inscription {{ self.parent_number }}</a></h1>
```
casey, line 1: Can we remove `class=light-fg`? And also in inscriptions_block.rs.

src/subcommand/server.rs
```diff
@@ -3238,6 +3296,248 @@ mod tests {
...
+      format!("/inscription/{parent_inscription_id}"),
+      StatusCode::OK,
+      format!(
+        ".*<title>Inscription 0</title>.*
+.*<a href=/inscription/.*><iframe .* src=/preview/.*></iframe></a>.*
+.*<a href=/inscription/.*><iframe .* src=/preview/.*></iframe></a>.*
+.*<a href=/inscription/.*><iframe .* src=/preview/.*></iframe></a>.*
+.*<a href=/inscription/.*><iframe .* src=/preview/.*></iframe></a>.*
+    <div class=center>
+      <a href=/children/{parent_inscription_id}>more</a>
```
casey, line 3448: Call more button `all`, and always link to children page.

casey (changes requested):
Left some comments.

casey (approved):
LGTM

devords:
@raphjaph Looks great to me. Thanks for adding those tests!

PR #2563, opened 2023-10-19
Respect locked utxos

dannydeezy:
Locked utxos should not be available to spend. (https://github.com/ordinals/ord/issues/2496)

This PR attempts to ensure that if a utxo has been locked with the lockunspent bitcoin RPC call, then it will not be selected as a cardinal utxo. This enables people to safely lock utxos that they don't want ord to spend when doing inscriptions or sends (unless they are deliberately inscribing/sending that particular utxo)

PR #2618, opened 2023-11-01
have ord respect locked coins

rot13maxi:
this is just a refactor of  #2563 on top of `master` now that everything is done through `Batch`.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -352,7 +352,7 @@ impl Index {
     self.durability = durability;
   }
 
-  pub(crate) fn get_unspent_outputs(&self, _wallet: Wallet) -> Result<BTreeMap<OutPoint, Amount>> {
+  pub(crate) fn get_unspent_outputs(&self, wallet: Wallet) -> Result<BTreeMap<OutPoint, Amount>> {
     let mut utxos = BTreeMap::new();
     utxos.extend(
       self
@@ -367,7 +367,7 @@ impl Index {
         }),
     );
 
-    let locked_utxos: BTreeSet<OutPoint> = self.get_locked_outputs(_wallet)?;
+    let locked_utxos: BTreeSet<OutPoint> = self.get_locked_outputs(wallet)?;
 
     for outpoint in locked_utxos {
       utxos.insert(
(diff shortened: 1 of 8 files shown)


src/index.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
```
raphjaph, line 1: move this use statement into the other ones

raphjaph (changes requested):
This looks good! Would be great if you could add an integration test for both send and inscribe.

src/index.rs
rot13maxi: good catch

raphjaph (approved):

raphjaph:
@rot13maxi, @dannydeezy I added some integration tests which unearthed another spot where we forgot to check locked outputs. Have a look if this looks good.

@casey You can also have look at this.

src/index.rs
```diff
@@ -351,19 +367,18 @@ impl Index {
         }),
     );
 
-    #[derive(Deserialize)]
-    pub(crate) struct JsonOutPoint {
-      txid: bitcoin::Txid,
-      vout: u32,
-    }
+    let locked_utxos: BTreeSet<OutPoint> = self.get_locked_outputs(_wallet)?;
```
casey, line 370: Wallet is being used now, so the underscore should be removed.

src/index.rs
```diff
@@ -351,19 +367,18 @@ impl Index {
...
+    for outpoint in locked_utxos {
       utxos.insert(
-        OutPoint { txid, vout },
-        Amount::from_sat(self.client.get_raw_transaction(&txid, None)?.output[vout as usize].value),
+        outpoint,
+        Amount::from_sat(
+          self
+            .client
+            .get_raw_transaction(&outpoint.txid, None)?
+            .output[outpoint.vout as usize]
```
casey, line 379: This should always work, but try_into/unwrap is always good, to avoid having to think about whether or not it could truncate:

```suggestion
            .output[outpoint.vout.try_into().unwrap()]
```

src/subcommand/wallet/inscribe/batch.rs
```diff
@@ -40,6 +40,7 @@ impl Batch {
     chain: Chain,
     index: &Index,
     client: &Client,
+    locked_utxos: &BTreeSet<OutPoint>,
```
casey, line 43: I like calling this `locked` instead of `locked_utxos`, since it's a collection of `OutPoint`s, but TIOLI.

casey (approved):
Looks good to me, very minor nits.

dannydeezy:
awesome thanks for getting this across the line folks!

PR #2628, opened 2023-11-03
Recursive endpoint for retrieving details about an inscription

devords:
This should be useful for a variety of use cases.


src/subcommand/server.rs
```diff
@@ -712,6 +717,56 @@ impl Server {
...
+    Ok(
+      Json(InscriptionDetailsJson {
+        address,
+        inscription_number: entry.inscription_number,
+        content_type: inscription.content_type().map(|s| s.to_string()),
+        content_length: inscription.content_length(),
+        genesis_fee: entry.fee,
+        genesis_height: entry.height,
+        timestamp: timestamp(entry.timestamp).timestamp(),
+      })
```
elocremarc, line 765: We should add output as well I think so we can eventually do a output recursive endpoint. Output was the first endpoint I tried working on like 4 months ago. This would have tons of uses. You could do dynamic content like a website that you can update based on coin control. Having an inscription aware of what output its on is how you would do this.

src/subcommand/server.rs
```diff
@@ -712,6 +717,56 @@ impl Server {
...
+    let address = output
+      .as_ref()
+      .and_then(|o| page_config.chain.address_from_script(&o.script_pubkey).ok())
+      .map(|address| address.to_string());
+
+    Ok(
+      Json(InscriptionDetailsJson {
+        address,
+        inscription_number: entry.inscription_number,
+        content_type: inscription.content_type().map(|s| s.to_string()),
```
elocremarc, line 760: Can't we just get this from the response of /content?

elocremarc:
I was thinking about making this endpoint and I would have done the ones you didn't include haha.
What I have wanted for months are
Address
Sat
Output

We should add output so we can eventually do a output recursive endpoint.  This would have tons of uses. You could do dynamic content like a website that you can update based on coin control. Having an inscription aware of what output its on is how you would do this.

Address would also play into this and all those combined gets us very close to making a PSBT marketplace all on inscriptions.

src/subcommand/server.rs
devords: That'd require loading the content though. If you want to fetch content types for a couple of inscriptions, you probably wouldn't want to load their content too.

devords:
I wanted to keep the response as minimal as possible, as we can always add stuff. But probably never remove. Happy to add whatever we want, if there's consensus. Output sounds good to me!

Re: sat, I think that'd require sat index, no? That's why I didn't add it.

elocremarc:
> Re: sat, I think that'd require sat index, no? That's why I didn't add it.
Yeah but were already crossing that bridge with sat endpoint. So it would return null probably if no sat index.

elocremarc:
> if there's consensus. Output sounds good to me!

@rot13maxi and I were talking on the sat endpoint spaces the other day about how we really want to get to the point where we can start making crazy cool inscriptions that get us closer to calling data on the bitcoin node. Been wanting to see if we can get openordex onto an inscription since it came out. Knowing an inscriptions output is invaluable for this kinda stuff.

src/subcommand/server.rs
elocremarc: True. This comment can probably be resolved.

devords:
I'll work on adding `output` to JSON.

Still not convinced about `sat`. My gut feeling is that we should refrain from adding recursive endpoints that rely on a full sat index. The risk is creating two classes of inscriptions, especially as the index grows larger and slower with every block.

We can achieve most of the things with `/children` recursive endpoint that would be made possible with `sat` endpoint. Agree that `sat` endpoint would make it much easier as there are no sit ups involved in figuring out the inscription id. But it also comes at a cost and maybe something we can punt on in favor of `/children`.

owenstrevor:
Would be awesome to add the Sat Endpoint. IMO the Sat Endpoint is the most important endpoint for Ordinals right now. With the Sat Endpoint, you can futureproof your collection so that you can access all of the endpoints that come out in the future, and then you can lock down your work by burning the sat. This would probably also give the dev some breathing room from people wanting stuff right away because they can just add the Sat Endpoint and then get it when it comes out.

devords:
I'm curious to know what thing is possible with sat endpoint that's not possible with a children endpoint?

devords:
@elocremarc Added `satpoint` to response JSON.

elocremarc:
> I'm curious to know what thing is possible with sat endpoint that's not possible with a children endpoint?

Sat is more controlled by creator and child can be collector basically. However you could make the sat endpoint work for a collector if you reference the sat that an inscription is being inscribed to. With sat you can know what the sat is prior to inscribing. However if you want to know the sat in some future child its good to have it here in this endpoint because it might not be known ahead of time.

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+    Extension(index): Extension<Arc<Index>>,
+    Path(inscription_id): Path<InscriptionId>,
+  ) -> ServerResult<Response> {
+    let inscription = index
+      .get_inscription_by_id(inscription_id)?
+      .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
+
+    let entry = index
+      .get_inscription_entry(inscription_id)?
+      .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
```
casey, line 731: We can `.unwrap` here, because we should never have an inscription in the inscription to ID table without a corresponding entry.

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+
+    let satpoint = index
+      .get_inscription_satpoint_by_id(inscription_id)?
+      .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
```
casey, line 735: We can unwrap here too.

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+          .output
+          .into_iter()
+          .nth(satpoint.outpoint.vout.try_into().unwrap())
+          .ok_or_not_found(|| format!("inscription {inscription_id} current transaction output"))?,
+      )
+    };
+
+    let address = output
+      .as_ref()
+      .and_then(|o| page_config.chain.address_from_script(&o.script_pubkey).ok())
```
casey, line 753:
```suggestion
      .and_then(|output| page_config.chain.address_from_script(&output.script_pubkey).ok())
```

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+      .map(|address| address.to_string());
+
+    Ok(
+      Json(InscriptionDetailsJson {
+        address,
+        inscription_number: entry.inscription_number,
```
casey, line 759:
```suggestion
        number: entry.inscription_number,
```

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+        content_type: inscription.content_type().map(|s| s.to_string()),
```
casey, line 760:
```suggestion
        type: inscription.content_type().map(|s| s.to_string()),
```

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+        content_length: inscription.content_length(),
```
casey, line 761:
```suggestion
        content: inscription.content_length(),
```

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+        genesis_fee: entry.fee,
```
casey, line 762:
```suggestion
        fee: entry.fee,
```

src/subcommand/server.rs
```diff
@@ -712,6 +717,57 @@ impl Server {
...
+        genesis_height: entry.height,
```
casey, line 763:
```suggestion
        height: entry.height,
```

src/templates/inscription.rs
```diff
@@ -19,6 +19,18 @@ pub(crate) struct InscriptionHtml {
   pub(crate) timestamp: DateTime<Utc>,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct InscriptionDetailsJson {
+  pub address: Option<String>,
```
casey, line 24:
```suggestion
  pub address: Option<Address>,
```

casey, line 24: These fields should be sorted.

casey (changes requested):
Almost there! See comments.

devords:
@elocremarc @casey I am planning to work on this again this week. Let me know and I can add whatever other keys to JSON.

elocremarc:
Been making some onchain inscription viewers I think having parent would also be super helpful here. Then you can traverse up and down the family tree. 
example: if you are on some inscription then you can see the siblings with the /children endpoint. Or find the ancestor.

> Let me know and I can add whatever other keys to JSON.

My vote adding the `sat_number` and `parent` to the JSON

owenstrevor:
> @casey don't we wanna add the sat number in this endpoint? This is very useful combined with the child endpoint or doing the sat endpoint on children you might not know the sat number for prior to inscribing like with batch inscriptions.

YES PLEASE. Right now we have to inscribe the sat number itself, either in the individual inscription metadata or in a lookup table afterwards. Both are hacky solutions where truth is what the dev wrote and not the blockchain. Biggest gap for us right now.

devords:
@owenstrevor Yeah, planning to work on this today!

raphjaph:
Is this ready for review?

devords:
@raphjaph Going to working on the requested changes today!

src/subcommand/server.rs
devords: @casey @raphjaph "type" is a reserved keyword, so changign this resulted in a bunch of errors. So, I kept `content_type` and `content_length` as I wasn't sure how to change `type`.

src/templates/inscription.rs
devords: @casey @raphjaph I couldn't figure this out. Followed the lead of what we already do with `InscriptionJson`. Can you please help?

devords:
@raphjaph Updated the branch with the suggested changes. Few open questions:

- I think we should make it clear that this recursive endpoint may add new keys in the future. For example, we may want to add `delegate` etc.
- I added `sat` to the JSON. The request will fail with `404` if sat index isn't found. Can you please review that bit?

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
     })
   }
 
+  async fn inscription_details(
```
raphjaph, line 840: Maybe call this `inscription_recursive` so it's clear that this function has stricter guarantees than the normal server functions.

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
...
+    Extension(server_config): Extension<Arc<ServerConfig>>,
+    Extension(index): Extension<Arc<Index>>,
+    Path(inscription_id): Path<InscriptionId>,
+  ) -> ServerResult<Response> {
+    if !index.has_sat_index() {
+      return Err(ServerError::NotFound(
+        "this server has no sat index".to_string(),
+      ));
+    }
```
raphjaph, line 849: I think the JSON should just have the sat field as an option instead of requiring a sat index for this endpoint.

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
...
+    let inscription = index
+      .get_inscription_by_id(inscription_id)?
+      .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
+
+    let entry = index
+      .get_inscription_entry(inscription_id)
+      .unwrap()
+      .unwrap();
+
+    let sat = entry.sat.unwrap();
```
raphjaph, line 860: Make Option

src/subcommand/server.rs
raphjaph: content_type should be fine

src/subcommand/server.rs
raphjaph: content_length is fine

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
...
+        server_config
+          .chain
+          .address_from_script(&output.script_pubkey)
+          .ok()
+      })
+      .map(|address| address.to_string());
+
+    Ok(
+      Json(InscriptionDetailsJson {
+        address,
```
raphjaph, line 894: We do not want to return an address, so as to discourage address reuse.

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
...
+        content_type: inscription.content_type().map(|s| s.to_string()),
+        content_length: inscription.content_length(),
+        fee: entry.fee,
+        height: entry.height,
+        number: entry.inscription_number,
+        postage: output.as_ref().map(|o| o.value),
```
raphjaph, line 900:
```suggestion
        value: output.as_ref().map(|o| o.value),
```

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
...
+        sat,
+        satpoint,
```
raphjaph, line 902: maybe add the output field as well

src/templates/inscription.rs
```diff
@@ -20,6 +20,20 @@ pub(crate) struct InscriptionHtml {
   pub(crate) charms: u16,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct InscriptionDetailsJson {
```
raphjaph, line 24:
```suggestion
pub struct InscriptionRecursiveJson {
```

tests/server.rs
```diff
@@ -281,6 +281,58 @@ fn inscription_metadata() {
...
+  let bitcoin_rpc_server = test_bitcoincore_rpc::spawn();
+  let ord_rpc_server =
+    TestServer::spawn_with_server_args(&bitcoin_rpc_server, &["--index-sats"], &[]);
+
+  create_wallet(&bitcoin_rpc_server, &ord_rpc_server);
+
+  bitcoin_rpc_server.mine_blocks(1);
+
+  let output = CommandBuilder::new("wallet inscribe --fee-rate 1 --file wizards.txt")
+    .write("wizards.txt", "THEWIZARDSOFORD")
```
raphjaph, line 295:
```suggestion
    .write("foo.txt", "FOO")
```

tests/server.rs
```diff
@@ -281,6 +281,58 @@ fn inscription_metadata() {
   );
 }
 
+#[test]
+fn recursive_inscription_metadata() {
```
raphjaph, line 285:
```suggestion
fn recursive_inscription_endpoint() {
```

raphjaph (changes requested):
Looking good, have a look at my comments

raphjaph:
I think we would also want to add things like charms

src/subcommand/server.rs
devords: Fixed in debb433585e50e3e3f2615204c6984d5e702db9c

src/subcommand/server.rs
```diff
@@ -832,6 +837,75 @@ impl Server {
     })
   }
 
+  async fn inscription_details(
+    Extension(server_config): Extension<Arc<ServerConfig>>,
+    Extension(index): Extension<Arc<Index>>,
+    Path(inscription_id): Path<InscriptionId>,
+  ) -> ServerResult<Response> {
```
raphjaph, line 844: Ah ja and wrap this whole thing in a task::block_in_place

src/subcommand/server.rs
devords: Added!

devords:
@raphjaph I think I addressed all of the comments, except `address` one.

- Added charms and output
- Made sat optional

Still holding out some hope for including `address`. It'll unlock so many use cases! I am curious if there's some way without having to compromise on address reuse. Maybe the address could be hashed?

raphjaph:
What use cases does address unlock?

devords:
@raphjaph The primary one I'm interested in is using address for randomization seed for generative art, where rarity would be tied to an address. Moving inscription to a new address would generate a new output. But moving it back to the previous address would then generate the output you had before.

Additionally, you can also do soulbound'esq inscriptions where the inscription behaves different when it's held by a specific owner.

devords:
@raphjaph Removed the address

ZedZeroth:
@devords Thank you for raising this. I've been thinking about such endpoints since mid-2023 and how much this kind of self-data would benefit the inscription ecosystem. I didn't realise that it was already in progress until @gmart7t2 pointed me here. I would like inscriptions be able to reference their own:

1. Inscription ID
2. Sat number
3. UTXO location
4. Address location
5. Metadata

It looks as though some of these are being pushed through soon? Could you clarify which ones are currently looking likely to be included and then I may respond with some arguments for including the others? :)

ZedZeroth:
> @raphjaph The primary one I'm interested in is using address for randomization seed for generative art, where rarity would be tied to an address. Moving inscription to a new address would generate a new output. But moving it back to the previous address would then generate the output you had before.
> 
> Additionally, you can also do soulbound'esq inscriptions where the inscription behaves different when it's held by a specific owner.

Yes 100% this!

ZedZeroth:
> @raphjaph I think I addressed all of the comments, except `address` one.
> 
>     * Added charms and output
> 
>     * Made sat optional
> 
> 
> Still holding out some hope for including `address`. It'll unlock so many use cases! I am curious if there's some way without having to compromise on address reuse. Maybe the address could be hashed?

Is this just a general "reusing bitcoin addresses is bad" argument, or is there something more specific to ordinals/inscriptions? Thanks

ep150de:
I also have Gen art that depends on the art changing based on the address that holds the inscription.Since a majority use wallets like xverse, unisat, etc. Which uses a constant static receiving address, we need to open the possibility to take advantage of this for creative features such as dynamic art based on the address of the holder that owns it.Sent via the Samsung Galaxy Note20 Ultra 5G, an AT&T 5G smartphone
-------- Original message --------From: ZedZeroth ***@***.***> Date: 2/18/24  04:46  (GMT-08:00) To: ordinals/ord ***@***.***> Cc: Subscribed ***@***.***> Subject: Re: [ordinals/ord] Recursive endpoint for retrieving details about an
  inscription (PR #2628) 

@raphjaph The primary one I'm interested in is using address for randomization seed for generative art, where rarity would be tied to an address. Moving inscription to a new address would generate a new output. But moving it back to the previous address would then generate the output you had before.
Additionally, you can also do soulbound'esq inscriptions where the inscription behaves different when it's held by a specific owner.

Yes 100% this!

casey (approved):
Looks good to me!

casey (approved):
Looks good to me!

devords:
This will be a fun one! Thank you @casey and @raphjaph 🙏

ZedZeroth:
Thanks for this. I think I am going to map out what inscriptions can currently access, and what they can't, and then push for "full inter-ordinal operability" where possible. It looks like a sat/inscription's current address is the only major one missing though? I have some further ideas but they may be more technically challenging/prohibitive to include. Huge progress with this has been made in the last six months though, it's fantastic for building :)

PR #2629, opened 2023-11-03
Add a link to Ordicord

devords:
Ordicord is a good place to seek help. This should make it easier to discover.


raphjaph (approved):
We have it in the docs but there's no harm in adding it to the website as well 👍

PR #2631, opened 2023-11-05
Make children link a little more visible

devords:
I found the current link very easy to miss because the text "all" being just 3 letters. Making the link a bit longer should make it a bit more visible.


raphjaph:
Thanks for the PR, but I think we like it better with one word.

PR #2632, opened 2023-11-05
Allow passing satpoint when batch inscribing

devords:
I am most likely missing something here as I imagine there's a reason why this is being prevented in the first place. But this change would be useful when you want to batch inscribe on a specific UTXO.

### DIFF
--- a/src/subcommand/wallet/inscribe.rs
+++ b/src/subcommand/wallet/inscribe.rs
@@ -55,7 +55,7 @@ pub(crate) struct Inscribe {
     long,
     help = "Inscribe a multiple inscriptions defines in a yaml <BATCH_FILE>.",
     conflicts_with_all = &[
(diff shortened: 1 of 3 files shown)


elocremarc:
Closes https://github.com/ordinals/ord/issues/2593

raphjaph:
The satpoint would have to be part of the batch file and be set for each inscription separately. In your implementation how do you know which inscription it will be applied to? The difficult part is building robust guardrails and defaults. 

Some considerations:
- If you set the same satpoint for multiple inscriptions, should it be a reinscription? What order?
- How does the satpoint interact with the two modes (`shared-output` and `separate-outputs`)?
- What to do with satpoints that are less than the dust limit apart?
- What if satpoints are crossing over each other? How do you manage the utxos to be in the correct order?

casey:
In the case of batch inscribing, the statpoint should be specified in the batch file. See #3100.

PR #2637, opened 2023-11-07
Avoid caching error responses

devords:
Error responses are currently cached by cloudflare, which is an attack vector. Someone could just watch mempool and make /content requests before the tx confirmation.

@raphjaph I am not able to get the tests to pass. Can you please point me in the right direction re: how to do this?


raphjaph (approved):

raphjaph:
@devord you set the header in the wrong place, this was a quick fix so we just did it for you :)

devords:
Thank you!

PR #2645, opened 2023-11-08
Simplify varint encoding

eagr:
Nit. Simplify the varint code a bit. :))


casey:
Nice, looks good to me! I switched it to use binary literals, since that makes more sense to me, and replaced `& 0x7f | 0x80` with just `| 0x80`, since it's the same.

PR #2655, opened 2023-11-09
Codespace bootstrap

### LINKED ISSUE #1760: Get GitHub codespaces up and running
- [ ] install cargo
- [ ] install just

eagr:
closes #1760 

Not sure if I got the intention right. Let me know if I'm mistaken. @raphjaph 

Did some clean up for the install script:
* Use `$()` instead of nonstd backticks
* `err()` would exit immediately so the latter lines are unreachable. Guess it's meant to use `say_err()`?
* Iterate over globs instead of ls output, which could be quite fragile

We should probably quote some of the variables to avoid accidental splitting or globbing if you want to be extra careful. Let me know if you up for it. :))

### DIFF
--- a/codespace.sh
+++ b/codespace.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+check_cmd() {
+  command -v "$1" > /dev/null 2>&1
+}
+
+if ! check_cmd cargo; then
+  curl https://sh.rustup.rs -sSf | sh -s -- -y --component clippy
+fi
+
+if ! check_cmd just; then
+  cargo install just
+fi
--- a/install.sh
+++ b/install.sh
@@ -23,17 +23,16 @@ OPTIONS:
 EOF
 }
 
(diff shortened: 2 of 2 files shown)


casey:
Can you put the changes to `installl.sh` into a different PR? I like backticks, so the change to`$(…)` should be reverted.

Correct me if I'm wrong, but I don't think this is the correct way to configure a codespace. I looked at the codespace docs, and I didn't find any reference to `setup.sh`. I think it should be done via a dockerfile in `.devcontainer/Dockerfile`. Please reopen if I'm incorrect!

eagr:
You're probably right. I missed this piece of doc. I read a different one that says you could auto-configure your codespaces via files with certain names that live in the dotfiles repo of your choice, like a `setup.sh` in `casey/dotfiles`. I thought the best bet was to tell people they can put that file in their dotfiles repo, which is a bad idea now that I think of it. Sorry about that. :stuck_out_tongue_closed_eyes:

PR #2656, opened 2023-11-10
The Jubilee: uncurse new inscriptions after jubilee height

### LINKED ISSUE #2495: The Jubilee
Currently, inscriptions that would not have been recognized by the first version of `ord` are cursed, meaning that they are assigned negative inscription numbers.

In the future, we would like to make new inscriptions that would have been cursed blessed, and assign them positive inscription numbers.

Since this requires coordination, we should pick a future block height at which this will take place. This can be called "The Jubilee", after years in which debts and sins are forgiven.

Are we ready to do this? Are there any sources of cursed inscriptions that we may want to allow and bless?

What block height should we choose?

I think we are ready to do this. There are a few types unrecognized inscriptions, but I don't think they should be recognized:

- Inscriptions in P2WSH and P2SH scripts. Don't provide much utility, make implementation more complex.
- Inscriptions using minimal opcodes instead of data pushes. Make inscriptions slightly smaller, but make the implementation more complex.

As for block height, I think we should give a significant amount of time, perhaps two full difficulty adjustments, and do it on the first block of a difficulty adjustment, so at least 28 days of notice.

This python script calculates an activation height when `current_height` is set to the current block height:

```
import math

current_height = 810528
next_adjustment = math.ceil(current_height / 2016)
activation_adjustment = next_adjustment + 2
activation_height = activation_adjustment * 2016
print(activation_height)
```

casey:
Fixes #2495.


gmart7t2:
This looks like new inscriptions with unrecognized even fields won't be cursed going forward. Is that the case? And if so, is that intended?

I would expect that we want unrecognized even fields to be cursed indefinitely.

casey:
That's correct. Inscriptions with unrecognized even fields will still be unbound, so they don't have a location, but not cursed.

adanamanya:
release the beast

rot13maxi:
my body is ready

rot13maxi:
Assuming 144 blocks per day, we’re now within two weeks of the activation height in this pr. I notice it’s still marked as a draft. @casey how are you feeling about it?

casey:
I think we should definitely give it minimum two weeks, and I'd like to do it on a difficulty adjustment, for style points, so it'll probably wind up being later.

raphjaph (approved):
All is forgiven 💸

dami025:
New inscription?

PR #2664, opened 2023-11-13
Use sequence numbers as database keys

casey:
Use sequence numbers instead of inscription IDs as database keys and values. Sequence numbers are 4 bytes, whereas Inscription IDs are 36 bytes, so this should save 32 bytes per key/value.


src/index.rs
```diff
@@ -803,16 +804,23 @@ impl Index {
...
+    let rtx = self.database.begin_read()?;
+
+    let sequence_number_to_inscription_entry =
+      rtx.open_table(SEQUENCE_NUMBER_TO_INSCRIPTION_ENTRY)?;
+
+    let mut collections = rtx
+      .open_multimap_table(SEQUENCE_NUMBER_TO_CHILDREN)?
       .iter()?
       .skip(page_index * page_size)
       .take(page_size + 1)
```
raphjaph, line 816:
```suggestion
      .take(page_size.saturating_add(1))
```

src/index.rs
```diff
@@ -984,7 +1047,7 @@ impl Index {
     if self
       .database
       .begin_read()?
-      .open_table(INSCRIPTION_ID_TO_SATPOINT)?
+      .open_table(INSCRIPTION_ID_TO_SEQUENCE_NUMBER)?
       .get(&inscription_id.store())?
       .is_none()
     {
```
raphjaph, line 1053: you could use `fn inscription_exists` here

src/index.rs
```diff
@@ -1455,65 +1554,22 @@ impl Index {
...
-
-    let inscription_id_to_satpoint = rtx.open_table(INSCRIPTION_ID_TO_SATPOINT).unwrap();
-    assert!(inscription_id_to_satpoint
-      .get(&inscription_id.store())
+    assert!(self
+      .database
+      .begin_read()
+      .unwrap()
+      .open_table(INSCRIPTION_ID_TO_SEQUENCE_NUMBER)
       .unwrap()
```
raphjaph, line 1562: could reuse `fn inscription_exists` here as well

raphjaph (changes requested):
some nits

raphjaph (approved):
LGTM! 

30% reduction in size + 20% speed up

PR #2669, opened 2023-11-14
Clean up install.sh

eagr:
Backticks are deprecated. Curious, why prefer them here?


casey:
As far as I know, the only downside of backticks is that they can't be nested. I think mainly I like backticks for aesthetic reasons, I think they look nicer than `$(…)`.

PR #2681, opened 2023-11-16
Add inscription charms

casey:
Charms are cute emoji that appear on inscription pages. Charms include rarity charms, for inscriptions on rare sats, cursed inscriptions, inscriptions on block 9 sats, lost sats, unbound inscriptions, and reinscriptions.

### DIFF
--- a/src/charm.rs
+++ b/src/charm.rs
@@ -42,7 +42,7 @@ impl Charm {
       Charm::Epic => "🪻",
       Charm::Legendary => "🌝",
       Charm::Lost => "🤔",
-      Charm::Nineball => "🌅",
+      Charm::Nineball => "9️⃣",
       Charm::Rare => "🧿",
       Charm::Reinscription => "♻️",
       Charm::Unbound => "🔓",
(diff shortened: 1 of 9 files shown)


devords:
If we're able to show the sequence number for the specific kind of curse, I think that'll be really really valuable. Sort of like an unstable inscription # for specific kind of curses.

Pretty please 🙏

src/charm.rs
```diff
@@ -0,0 +1,66 @@
...
+    charms & self.flag() != 0
+  }
+
+  pub(crate) fn icon(self) -> &'static str {
+    match self {
+      Charm::Cursed => "👹",
+      Charm::Epic => "🪻",
+      Charm::Legendary => "🌝",
+      Charm::Lost => "🤔",
+      Charm::Nineball => "🌅",
```
raphjaph, line 45:
```suggestion
      Charm::Nineball => "❾",
```

raphjaph (approved):

raphjaph:
🪄

casey:
> If we're able to show the sequence number for the specific kind of curse, I think that'll be really really valuable. Sort of like an unstable inscription # for specific kind of curses.
> 
> Pretty please 🙏

Showing sequence is a little bit controversial. I'm in favor of doing it, but it should probably be discussed beforehand. Feel free to open an issue!

PR #2683, opened 2023-11-16
Init codespace

### LINKED ISSUE #1760: Get GitHub codespaces up and running

eagr:
closes #1760 

Using 1.70 as it seems to be the msrv of many tools being used like the fuzzer.


raphjaph:
Thanks for the PR, but I don't think we want this anymore, it's probably too much of a pain to maintain.

PR #2691, opened 2023-11-17
Add Dockerfile and docker-compose.yml

bensig:
This is a simple way to add docker support using a small docker image (ubuntu/latest) and installing the latest `ord` release using `install.sh` instead of building from source which uses a much bigger image.

Paths in docker-compose.yml can be updated to run on OS X, or in a custom path instead of /ordinals - for example if you prefer `/var/local/ord` or the more bitcoiny `~/.ord`.

Can be easily run with:

`docker-compose up -d`


bensig:
Updated docker-compose.yml to include default paths for ord index data in each OS

raphjaph:
`install.sh` doesn't seem to work properly inside a docker image. Maybe we should build from source? 

I'm on a Mac and getting this:
```
 [7/7] RUN ./install.sh --to /ordinals:
#10 0.643 install.sh: Could not determine target from output of `uname -m`-`uname -s`, please use `--target`: aarch64-Linux
```

bensig:
That minimum Dockerfile has a huge image for building rust. This Dockerfile uses a smaller image which was chosen based on the comments there. @raphjaph I fixed the Dockerfile so that `install.sh` works correctly now.

PR #2692, opened 2023-11-17
Font Inscription Preview

elocremarc:
We need a way to view font inscriptions. This PR adds an editable textbox in the iframe that loads the font in the preview from the inscription so you can test it out and see it. 
image
image


templates/preview-font.html
```diff
@@ -0,0 +1,10 @@
+<!doctype html>
+<html lang=en data-inscription={{self.inscription_id}}>
+  <head>
+    <meta charset=utf-8>
+    <link rel=stylesheet href=/static/preview-font.css></link>
+    <script src=/static/preview-font.js defer type=module></script>
```
casey, line 6: Can we do this without any JS, like this?

```suggestion
    <style>
      @font-face {
        font-family: 'Font';
        src: url(/content/{{self.inscription}}) format('{self.format}');
      }

      textarea {
        font-family: 'CustomFont';
        background-color: #292c2f;
        color: #98a3ad;
        border: 1px solid #292c2f;
        width: 90vw;
        height: 90vh;
        resize: none;
        padding: 10px;
        font-size: 24px;
    }

```

templates/preview-font.html
```diff
@@ -0,0 +1,10 @@
...
+  </head>
+  <body>
```
casey, line 8:
```suggestion
  <body>
    <textarea>Sphinx of black quartz, judge my vow.</textarea>
```

casey (changes requested):
This is sick. I think we can avoid JS by templatizing the CSS.

templates/preview-font.html
elocremarc: Hmm yeah that seems cleaner. So would we do this logic serverside? 
```js
const contentType = response.headers.get("content-type");
let fontType = contentType.split("/")[1];
switch (fontType) {
    case "ttf":
        fontType = "truetype";
        break;
    case "otf":
        fontType = "opentype";
        break;
}
```
You had mentioned doing somthing serverside in the stream for the codepreview. Could you elaborate on how you would approach doing this? woff and woff2 the fontType can be the mime type however ttf and otf need this mapping.

casey: Yeah, I think it's better to avoid calculations in JS where possible. Our JS is untested, since frontend tests are slow and flakey, so having as little logic there as possible is best.

For the code preview, it would also be nice to pass the code type string instead of getting it from the content type, for the same reasons.

rot13maxi:
this is dope

casey:
@elocremarc I just [read](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src#format) that the `format()` declaration is optional, so I think we can just omit it, and the browser will auto-detect the font type.

elocremarc:
@casey way more elegent. It works! Just pushed changes. 
image

elocremarc:
Made some style tweaks broke it out into a css file again. See if your satisfied with how the CSS renders on the iframe and preview page.

PR #2701, opened 2023-11-19
Add destination field to batch

raphjaph:
Tests:
- [x] with some destinations set and others not
- [x] if `shared-output` mode then destination field not allowed -> should fail
- [x] invalid address set in destination


batch.yaml
```diff
@@ -25,10 +25,13 @@ inscriptions:
         metus est et odio. Nullam venenatis, urna et molestie vestibulum, orci
         mi efficitur risus, eu malesuada diam lorem sed velit. Nam fermentum
         dolor et luctus euismod.
+    destination: bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4
 
   - file: token.json
     metaprotocol: brc-20
+    destination: bc1pxwww0ct9ue7e8tdnlmug5m2tamfn7q06sahstg39ys4c9f3340qqxrdu9k
```
casey, line 32: Let's have one default.

```suggestion
```

src/subcommand/wallet/inscribe/batch.rs
```diff
@@ -532,6 +532,7 @@ pub(crate) struct BatchEntry {
   pub(crate) file: PathBuf,
```
casey, line 532: sort

src/subcommand/wallet/inscribe/batch.rs
```diff
@@ -568,13 +569,22 @@ impl Batchfile {
...
+    client: &Client,
     chain: Chain,
     parent_value: Option<u64>,
     metadata: Option<Vec<u8>>,
     postage: Amount,
     compress: bool,
-  ) -> Result<Vec<Inscription>> {
+  ) -> Result<(Vec<Inscription>, Vec<Address>)> {
     assert!(!self.inscriptions.is_empty());
+    assert!(
```
casey, line 580: We should return an anyhow error here, since this is an error that the user could trigger.

src/subcommand/wallet/inscribe/batch.rs
```diff
@@ -600,9 +611,20 @@ impl Batchfile {
         compress,
       )?);
 
+      if !(self.mode == Mode::SharedOutput && i >= 1) {
```
casey, line 614: Maybe try refactoring this into something like this, so that we get some type safety:

```rust
enum Destinations {
  Shared { inscriptions: Vec<Inscription>, destination: Address },
  Separate { bar: Vec<(Inscription, Address)> }
}

impl InscriptionDestinations {
  fn bar(&self) -> Vec<(Inscription, Address)> {
    match self {
      Self::Shared { inscription, destination} => {
        self.inscriptions.map(|inscription| (inscription, self.address)).collect(),
      }
      Self::Separate { bar} => self.bar.iter().collect(),
    }
  }
}
```

tests/wallet/inscribe.rs
```diff
@@ -1343,3 +1343,86 @@ fn inscriptions_are_not_compressed_if_no_space_is_saved_by_compression() {
   assert_eq!(response.status(), StatusCode::OK);
   assert_eq!(response.text().unwrap(), "foo");
 }
+
+#[test]
+fn batch_inscribe_fails_if_invalid_destination_address() {
```
casey, line 1348: I think we can remove this. But we should have a test that if the network is wrong, we catch it.

tests/wallet/inscribe.rs
```diff
@@ -1343,3 +1343,86 @@ fn inscriptions_are_not_compressed_if_no_space_is_saved_by_compression() {
...
+      "mode: separate-outputs\ninscriptions:\n- file: inscription.txt\n  destination: bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4\n- file: tulip.png\n- file: meow.wav\n  destination: bc1pxwww0ct9ue7e8tdnlmug5m2tamfn7q06sahstg39ys4c9f3340qqxrdu9k\n"
+    )
+    .rpc_server(&rpc_server)
+    .run_and_deserialize_output::<Inscribe>();
+
+  rpc_server.mine_blocks(1);
+
+  assert_eq!(rpc_server.descriptors().len(), 3);
+
+  TestServer::spawn_with_args(&rpc_server, &[]).assert_response_regex(
```
casey, line 1408: We should spawn the server once and reuse it for these three asserts

tests/wallet/inscribe.rs
```diff
@@ -1343,3 +1343,86 @@ fn inscriptions_are_not_compressed_if_no_space_is_saved_by_compression() {
...
+    format!("/inscription/{}", output.inscriptions[0].id),
+    ".*
+  <dt>address</dt>
+  <dd class=monospace>bc1qw508d6qejxtdg4y5r3zarvary0c5xw7kv8f3t4</dd>.*",
+  );
+
+  TestServer::spawn_with_args(&rpc_server, &[]).assert_response_regex(
```
casey, line 1415: Let's add new change addresses to a `HashMap<Address>` in the test server, and check that the address that it's sent to is one of the change addresses.

casey (changes requested):

src/subcommand/wallet/inscribe/batch.rs
```diff
@@ -570,14 +571,26 @@ impl Batchfile {
...
     assert!(!self.inscriptions.is_empty());
 
+    if self
+      .inscriptions
+      .iter()
+      .any(|entry| entry.destination.is_some())
+      && self.mode == Mode::SharedOutput
+    {
+      return Err(anyhow!(
+        "destination field cannot be used in shared-output mode"
```
casey, line 590:
```suggestion
        "individual inscription destinations cannot be set in shared-output mode"
```

src/subcommand/wallet/inscribe/batch.rs
```diff
@@ -602,9 +616,20 @@ impl Batchfile {
         compress,
       )?);
 
+      if !(self.mode == Mode::SharedOutput && i >= 1) {
```
casey, line 619: Let's move this out of the loop, and replace it with something like:

```
let destinations = match self.mode {
  Mode::SharedOutput => vec![get_change_address(client, chain)],
  Mode::SeparateOutputs => {
  }
}
```

casey (approved):
LGTM!

PR #2708, opened 2023-11-21
add origin to CSP so Safari can do recursion

### LINKED ISSUE #2229: Recursive Ordinals are blocked and not functional for display on Safari and Safari & Chrome & Imbeded browsers on iOS
Due to incompatibility with Safari and apples CSP policy recursive ordinals do not function at all on roughly 40% of the mobile market. This is a hard block to which we've been unable to find a successful work around. 

This issue was mentioned prior to recursion release with regards to Safari (OSX) however this is the case in all browsers on iOS which makes up the a good portion of internet traffic globally, albiet currently there are few ordinal wallets or websites geared towards mobile. 

This means all the recursive ordinals sold thus far, or in the future utilizing the current protocol standards may/will fail to render as expected for the majority of owners or platforms. 

Example: 

https://ordinals.com/content/821006d94ecbe675dcb6bcb911a6e55c0fc3015f57b64ef366c84899ff8a7a20i0 Refused to load because it appears in neither the img-src directive nor the default-src directive of the Content Security Policy.

We have tested this on multiple recursive ordinals from both ordinals.com and the magiceden content mirror during testing of recursion for project and artist usage. 

Examples of testing include basic recursion such as recursive frogs:  (Simply layering) 
https://ord-mirror.magiceden.dev/content/3e0838c4e814f7ed88ef932e09ae56a5b6627dbfed89ee8ba160edcdb527b303i0

And more complex which pulls in libraries to display images: (rotating 3d gallery) 
https://ordinals.com/content/0829ab0e3a096e37604a52a08acbf343146fe4daa4469f54408a19e10e5a5917i0

Images below.. note this is the case on ordinals.com and magiceden and other sites which implement protocol correct security policies. 

Given the issue with CSPs was presented prior to release, this is likely not news with regards to safari incompatibility, but iOS broadly renders recursion non operable for the majority of users, and projects selling or users buying recursive ordinals need to understand these current limitations. 

Apples CSP policy is not likely to change, and is secure, if anything it is likely to propagate to other browsers over time. 

<details>
IMG_7AC4C94C7D78-1
IMG_0044
IMG_0045
IMG_0046
IMG_0048
</details>

rot13maxi:
adds a `content_security_policy_origin` option to `server`. Tested on safari on macos 14.1.1 by doing `ord server --content-security-policy-origin http://0.0.0.0:80` and was able to view recursive inscriptions correctly. 

fixes #2229


casey:
Does this have to be a full url with `http/https`, or can it just be a host?

Otherwise this looks good. Maybe a shorter flag name, like `--csp-origin`?

rot13maxi:
it looks like you need the full url. 

flag shortened

src/subcommand/server.rs
```diff
@@ -3151,6 +3172,38 @@ mod tests {
...
+      AcceptEncoding::default(),
+      &PageConfig {
+        content_security_policy_origin: Some("https://ordinals.com".into()),
+        ..Default::default()
+      },
+    )
+    .unwrap()
+    .unwrap();
+
+    assert_eq!(headers["content-security-policy"], HeaderValue::from_static("default-src https://ordinals.com/content/ https://ordinals.com/blockheight https://ordinals.com/blockhash https://ordinals.com/blockhash/ https://ordinals.com/blocktime https://ordinasl.com/r/ 'unsafe-eval' 'unsafe-inline' data: blob:"));
```
rot13maxi, line 3204:
```suggestion
    assert_eq!(headers["content-security-policy"], HeaderValue::from_static("default-src https://ordinals.com/content/ https://ordinals.com/blockheight https://ordinals.com/blockhash https://ordinals.com/blockhash/ https://ordinals.com/blocktime https://ordinals.com/r/ 'unsafe-eval' 'unsafe-inline' data: blob:"));
```

raphjaph (approved):
lgtm!

PR #2709, opened 2023-11-21
Combine inscriptions_on_output functions

casey:
This does a few things:

- removes a comment which I think no longer applies
- folds `inscriptions_on_output_ordered` and `inscriptions_on_output_unordered`, since `inscriptions_on_output_unordered` is never called anywhere where the increased performance would make a difference


casey:
I already did this in the sequence number key PR.

PR #2724, opened 2023-11-22
Fix coin selection.

### LINKED ISSUE #2723: Coin selection sometimes fails
https://github.com/ordinals/ord/pull/2303#issuecomment-1656941131 misses a case from my original pull request. 

If the best coin so far doesn't satisfy the condition and the new coin does, we don't care whether the new coin is "closer" than the old one, it is the new best coin whether it is closer or not. 

I've seen this cause failures to inscribe and to send inscriptions, where the algorithm finds a very close, but too small, utxo, it won't switch to a big enough but not as close utxo, and so fails completely to fund the tx.

gmart7t2:
Fixes #2723.

### DIFF
--- a/src/subcommand/wallet/transaction_builder.rs
+++ b/src/subcommand/wallet/transaction_builder.rs
@@ -1985,4 +1985,58 @@ mod tests {
       outpoint(2),
     );
   }
+
+  #[test]
+  fn prefer_further_away_utxos_if_they_are_newly_under_target() {
+    let utxos = vec![
+      (outpoint(1), Amount::from_sat(510)),
+      (outpoint(2), Amount::from_sat(400)),
+    ];
+
+    let mut tx_builder = TransactionBuilder::new(
+      satpoint(0, 0),
+      BTreeMap::new(),
+      utxos.into_iter().collect(),
+      BTreeSet::new(),
+      recipient(),
+      [change(0), change(1)],
+      FeeRate::try_from(1.0).unwrap(),
+      Target::Value(Amount::from_sat(10_000)),
+    );
+
+    assert_eq!(
+      tx_builder
+        .select_cardinal_utxo(Amount::from_sat(500), true)
+        .unwrap()
+        .0,
+      outpoint(2),
+    );
+  }
+
+  #[test]
+  fn prefer_further_away_utxos_if_they_are_newly_over_target() {
+    let utxos = vec![
+      (outpoint(1), Amount::from_sat(490)),
+      (outpoint(2), Amount::from_sat(600)),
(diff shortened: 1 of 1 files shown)


casey:
Thanks for the PR! I added a couple of tests. Do they look right to you?

casey:
I pushed some tests, but it looks like you force-pushed the branch. This needs tests before it can be merged, so that we can be sure that it fixes the issue, and so that we can be sure we don't break it in the future. Here are the tests I added:

```rust
  #[test]
  fn prefer_further_away_utxos_if_they_are_under_target() {
    let utxos = vec![
      (outpoint(1), Amount::from_sat(500)),
      (outpoint(2), Amount::from_sat(300)),
    ];

    let mut tx_builder = TransactionBuilder::new(
      satpoint(0, 0),
      BTreeMap::new(),
      utxos.into_iter().collect(),
      BTreeSet::new(),
      recipient(),
      [change(0), change(1)],
      FeeRate::try_from(1.0).unwrap(),
      Target::Value(Amount::from_sat(10_000)),
    );

    assert_eq!(
      tx_builder
        .select_cardinal_utxo(Amount::from_sat(400), true)
        .unwrap()
        .0,
      outpoint(2),
    );
  }

  #[test]
  fn prefer_further_away_utxos_if_they_are_over_target() {
    let utxos = vec![
      (outpoint(1), Amount::from_sat(300)),
      (outpoint(2), Amount::from_sat(500)),
    ];

    let mut tx_builder = TransactionBuilder::new(
      satpoint(0, 0),
      BTreeMap::new(),
      utxos.into_iter().collect(),
      BTreeSet::new(),
      recipient(),
      [change(0), change(1)],
      FeeRate::try_from(1.0).unwrap(),
      Target::Value(Amount::from_sat(10_000)),
    );

    assert_eq!(
      tx_builder
        .select_cardinal_utxo(Amount::from_sat(400), false)
        .unwrap()
        .0,
      outpoint(2),
    );
  }
```

Please confirm that the test the issue that's being fixed here, and if they do, add them to the PR.

raphjaph:
@gmart7t2

gmart7t2:
I will check the test and add it to the PR.

Basically the issue is that if you are looking for an output bigger than or equal to 10, and you see outputs of size 9 and 20, in that order, the code will take the 9 output as the best, and refuse to switch to the 20 because even though it is the first one that is big enough, it isn't closer to the target than the current best. So the coin selection will select the 9, which isn't big enough.

raphjaph (approved):

casey:
Nice, thank you!

PR #2680, opened 2023-11-15
Add sat recursive endpoints with index and pagination

raphjaph:
This PR takes the discussion from #2452 and implements the following recursive endpoints:

`/r/sat/:sat` ->  get the first 100 inscription ids on a sat (ascending reinscription order)
`/r/sat/:sat/:page` -> get 100 inscription ids on sat paginated (ascending reinscription)

These two endpoints return:

```json
{
  "ids": [ 
    "c3a05e88abd51708954ce562389478028567184e32fb96ab8d07b672989155b8i0", 
    "c3a05e88abd51708954ce562389478028567184e32fb96ab8d07b672989155b8i1"
  ],
  "page": 2,
  "more": false,
}
```

`/r/sat/:sat/at/:index` -> get a single inscriptions id on a sat by index (-1 is latest, 0 is first inscription)

This returns: 

```json
{  
  "id": "c3a05e88abd51708954ce562389478028567184e32fb96ab8d07b672989155b8i0", 
}
```

`:sat` can be any represenation of a sat including number, degree or letter notation.

PR #2732, opened 2023-11-23
Only accept sat number in recursive endpoint

### LINKED ISSUE #2730: Move sat recursive endpoint test from integration to server test

raphjaph:
I think it would be best for now to only accept sat numbers in the recursive endpoints because I think marketplaces and other explorers do not have the sat deserialization implemented. We can always expand this later.

Fixes #2730 

Changes #2680


casey:
Is that because they have reimplementations of all the endpoints?

raphjaph:
I'm just not sure if they have implemented notations like degree, decimal or percentile. If we accept all of those they will have to implement it. Don't want to drown them in work.

casey (approved):

PR #2745, opened 2023-11-24
Add stuttering curse

casey:
The envelope parser advances instruction by instruction, which causes the parser to advance too far when parsing inscriptions with the following structure:

```
OP_FALSE
OP_FALSE # checks for OP_IF, but skips instruction, so following inscription isn't recognized
OP_IF
"ord"
OP_ENDIF
```

```
OP_FALSE
OP_IF
OP_FALSE  # checks for "ord", but skips instruction, so following inscription isn't recognized
OP_IF
"ord"
OP_ENDIF
```

This adds yet another curse, the stuttering curse, for inscriptions with this structure.

@ipoh123 Have a look! Also, if there are any other issue you found, let us know! And, by the way, how did you find this?

### DIFF
--- a/src/envelope.rs
+++ b/src/envelope.rs
@@ -2,8 +2,13 @@ use {
   super::*,
   bitcoin::blockdata::{
     opcodes,
-    script::{self, Instruction, Instructions},
+    script::{
+      self,
+      Instruction::{self, Op, PushBytes},
+      Instructions,
+    },
   },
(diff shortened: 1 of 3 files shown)


ipoh123:
:) 
From the code, searching for the envelopes should be akin to the substring matching problem. It is easy to make mistakes here.

raphjaph (approved):
Did I ststststutter?

PR #2748, opened 2023-11-25
 uncommon charm parity with other explorers

elocremarc:
Most people already associate a diamond with an uncommon its become the standard across different explorers and marketplaces. Doesn't look like the rare and others `gemstones` are emojis so we could start with this one. 
 
 🌱 -> 💎


casey:
We already don't use gemstones for Rare and Epic, so those won't match. Plus, the current charms are color-coded, so they match the color of the rarity. Also, I think the diamond looks to rare for uncommon. So I'm inclined to keep it as it is.

PR #2753, opened 2023-11-26
Hide all text

raphjaph:


casey:
Maybe we should just hide all text inscriptions from the homepage. Text inscriptions are cool, but the number that are non-spam, and that look good on the homepage is small.

raphjaph:
Works for me

casey (approved):
RIP text inscriptions 🫡

Issue #2597, opened 2023-10-26
Provide a way to burn an inscription


devords:
Ord client should provide an official way to burn an inscription. This way, a parent inscription can be burned after creating a collection, and we can show a "burned" label on the new collection page #2561.

cc @veryordinally

DrJingLee:
Something like When inscription has an ownwr address of 1BitcoinEaterAddressDontSendf59kuE

then give it a tag of “burned” ?

elocremarc:
I think burning to an unspendable output would be ideal. Then you can still sign for the inscription. It would allow for POAPs and Soul Bound Tokens.

raphjaph:
Ideally we would want to burn to an `OP_RETURN` because this way Bitcoin can prune the output.

casey:
Dupe of https://github.com/ordinals/ord/issues/1333.

PR #2766, opened 2023-11-28
implement burn

onchainguy-eth:
PR for #2597

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /target
 /test-times.txt
 /tmp
+/.fleet/
--- a/src/charm.rs
+++ b/src/charm.rs
@@ -1,5 +1,6 @@
 #[derive(Copy, Clone)]
 pub(crate) enum Charm {
+  Burned,
   Coin,
   Cursed,
   Epic,
@@ -13,7 +14,7 @@ pub(crate) enum Charm {
 }
 
 impl Charm {
-  pub(crate) const ALL: [Charm; 10] = [
+  pub(crate) const ALL: [Charm; 11] = [
     Self::Coin,
     Self::Uncommon,
     Self::Rare,
@@ -24,6 +25,7 @@ impl Charm {
     Self::Cursed,
     Self::Unbound,
     Self::Lost,
+    Self::Burned,
   ];
 
   fn flag(self) -> u16 {
@@ -40,6 +42,7 @@ impl Charm {
 
   pub(crate) fn icon(self) -> &'static str {
     match self {
+      Self::Burned => "🔥",
       Self::Coin => "🪙",
       Self::Cursed => "👹",
       Self::Epic => "🪻",
@@ -55,6 +58,7 @@ impl Charm {
 
   pub(crate) fn title(self) -> &'static str {
     match self {
+      Self::Burned => "burned",
       Self::Coin => "coin",
       Self::Cursed => "cursed",
       Self::Epic => "epic",
--- a/src/index/updater/inscription_updater.rs
+++ b/src/index/updater/inscription_updater.rs
@@ -18,6 +18,7 @@ pub(super) struct Flotsam {
   inscription_id: InscriptionId,
   offset: u64,
   origin: Origin,
+  burned: bool,
 }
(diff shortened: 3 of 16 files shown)


casey (changes requested):
Thanks for the PR! Just doing a quick review for now.

- There are a lot of whitespace changes, can you reformat to remove them?
- I'd prefer not to introduce a new struct to represent an OutputScript. Can we just use a ScriptBuf? There is a helper function to convert a script buf to an address, if possible.
- We should use the helper functions `ScriptBuf::new_op_return` and `Script::is_op_return`, to create a new op return and check if an output is an op return.
- Op returns can contain data pushes, `OP_N` opcodes, and `OP_RESERVED`. I think we should have a plan that allows us to display both correctly, and eventually allow users to perform both types of burns. I'm thinking that we should burn with a marker that indicates that this is a human-readable text. So if we're burning text, what winds up on chain is `OP_RETURN MARKER <PUSHES>`. If the OP_RETURN matches that pattern, we collect the pushes into a string and display them. Otherwise we display the script. For the marker, I'm thinking we should use `OP_1NEGATE`. `OP_1NEGATE` is pretty useless in an OP_RETURN, it pushes 0x83 onto the stack, and is pretty much only useful when evaluating a Bitcoin Script, so isn't very useful in an OP_RETURN.
- Let's add a burn charm, it should be 🔥, natch.
- Burn OP_RETURNS should always have value 0. We should add a test for that.

onchainguy-eth:
@casey thanks for the feedback. All makes sense. Except the problem with the Script helper is that it only allows to create p2pkh, p2sh or witness programs from script. It will fail with an `UnrecognizedScript` error if you try it with an OP_RETURN. We could create our own helper but actually then going for the extra struct might be cleaner.

casey:
> @casey thanks for the feedback. All makes sense. Except the problem with the Script helper is that it only allows to create p2pkh, p2sh or witness programs from script. It will fail with an `UnrecognizedScript` error if you try it with an OP_RETURN. We could create our own helper but actually then going for the extra struct might be cleaner.

I'm not sure I understand where specifically this is an issue. I know we need to convert the script to an address in some places, but if conversion fails, can we substitute the correct value for an op return with `unwrap_or_else`?

onchainguy-eth:
> > @casey thanks for the feedback. All makes sense. Except the problem with the Script helper is that it only allows to create p2pkh, p2sh or witness programs from script. It will fail with an `UnrecognizedScript` error if you try it with an OP_RETURN. We could create our own helper but actually then going for the extra struct might be cleaner.
> 
> I'm not sure I understand where specifically this is an issue. I know we need to convert the script to an address in some places, but if conversion fails, can we substitute the correct value for an op return with `unwrap_or_else`?

lol, you're right. I'm a rust noob sorry 💀 Will get done with the changes!

casey:
> lol, you're right. I'm a rust noob sorry 💀 Will get done with the changes!

No worries, Rust is hard!

onchainguy-eth:
@casey was just about to complete the changes, one thing I wanted to discuss is setting OP_RETURN value to 0 (maybe I don't get what you mean). If we set the value to 0, we need to store some info on the inscription (in ord) that it has been burned because the sats still exist. So the sats would be freed which is nice. However, marketplaces would also need to fetch the info to avoid people buying burned inscriptions in a collection. Now somebody might list an inscription on a marketplace but then submit a burn shortly before the buyer gets it. I mean yes, the buyer gets it but it would have the burned property. Instead if we just burn the sats, the inscription can't be traded at all anymore. The sats might be "wasted" but a little deflation for BTC might be ok, people won't really care to much in most scenarios. We could also have different burn modes if we would like to support both.

casey:
@onchainguy-eth Whoops, I'm an idiot, of course the OP_RETURN has to have a nonzero value in order to burn the sat 😅

onchainguy-eth:
@casey just pushed an update with all the requested changes. Some stuff wasn't sure if best way to do it like e.g. assigning the charm.

onchainguy-eth:
Just rebased the rune changes

PR #2786, opened 2023-11-30
Minimal Dockerfile

raphjaph:
Still need to find a docker linux image that included `GLIBC_2.32`.

```
docker run ordinals/ord ord --help
ord: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ord)
ord: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ord)
ord: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ord)
```

How do we fix this?


casey:
On a plan, so I can't actually fix this, but I think we can use the rust image both as the build image, and as the runtime image: https://hub.docker.com/_/rust/

lesterli:
> On a plan, so I can't actually fix this, but I think we can use the rust image both as the build image, and as the runtime image: https://hub.docker.com/_/rust/

I tested this idea, and it works.

raphjaph:
Update the Dockerfile and works now. Would just merge this @casey

casey (approved):

PR #2787, opened 2023-11-30
Add binary to audit Cloudflare caching

casey:
Added a crate, `audit-caching`, which hits endpoints and prints whether or not they are correctly cached by cloudflare.

Can be run with `cargo run -p audit-caching`, and prints out results with cute colors:

Screenshot 2023-11-30 at 3 43 50 PM

We can add new test cases if our caching strategy gets more complex, but for now it seems good enough.


raphjaph (approved):
Cache Rules Everything Around Me

PR #2798, opened 2023-12-02
HTML endpoint /inscriptions/block/<height>/<page> returns nothing after page 0

gmart7t2:
* go to https://ordinals.com/block/800000
* it correctly shows "2914 Inscriptions"
* click "more" -> https://ordinals.com/inscriptions/block/800000
* it correctly shows the first 100 inscriptions
* click "next" -> https://ordinals.com/inscriptions/block/800000/1
* it incorrectly shows no inscriptions

Both `Server::inscriptions_in_block_paginated()` and `InscriptionsBlockHtml::new()` are removing the first `page_index * page_size` inscriptions. Only one of them should do it. 

This commit fixes the issue.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -1453,13 +1453,13 @@ impl Server {
       .take(page_size.saturating_add(1))
       .collect::<Vec<InscriptionId>>();
 
-    let more = inscriptions.len() > page_size;
-
-    if more {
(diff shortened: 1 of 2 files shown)


raphjaph (approved):
Thanks! 

Just simplified the logic a bit and wrote a test.

PR #2804, opened 2023-12-03
Update audit-cache binary

casey:
- Cache non-HTML previews for four hours
- Test expected cache-control header


raphjaph (approved):
lgtm

PR #2819, opened 2023-12-04
Status

casey:
@raphjaph This isn't ready to merge yet, but I wanted to ask: What do you think we should call this page? I was thinking /status, /info, or /stats. We already have a `/status` page, which reports reorgs, so maybe `/info`?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2108,6 +2108,7 @@ dependencies = [
  "hex",
  "html-escaper",
  "http",
+ "humantime",
  "hyper",
  "indicatif",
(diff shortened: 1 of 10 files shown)


raphjaph:
Why not put it under `/status` and making `reorged` a field in the HTML? Would need to update a couple tests and deploy scripts but should be fine

casey:
@raphjaph Ready for review! I didn't write tests for the actual status fields. I'm not sure it's actually worth it.

raphjaph (approved):

PR #2831, opened 2023-12-08
Reserve runes AAAAAAAAAAAAAAAAAAAAAAAAAAA and above

casey:


casey:
@raphjaph This is ready for review.

src/index/updater/rune_updater.rs
```diff
@@ -68,10 +68,34 @@ impl<'a, 'db, 'tx> RuneUpdater<'a, 'db, 'tx> {
...
+                .map(|entry| entry.value())
+                .unwrap_or_default();
+
+              self
+                .statistic_to_count
+                .insert(&Statistic::ReservedRunes.into(), reserved_runes + 1)?;
+
+              Rune::reserved(reserved_runes.into())
+            };
+
```
raphjaph, line 98: Since you want to start writing some documentation comments, I think this would be a great place to write some things. Always more difficult to find these spots later.

src/runes.rs
```diff
@@ -261,6 +262,218 @@ mod tests {
     }
   }
 
+  #[test]
+  fn runes_must_be_unreserved() {
```
raphjaph, line 266: Maybe this?
```suggestion
  fn runes_cannot_specify_reserved_rune() {
```

src/runes.rs
```diff
@@ -736,6 +949,52 @@ mod tests {
...
+      inputs: &[(1, 0, 0, Witness::new())],
+      op_return: Some(
+        Runestone {
+          edicts: vec![Edict {
+            id: 0,
+            amount: u128::max_value(),
+            output: 0,
+          }],
+          etching: Some(Etching::default()),
+          burn: true,
```
raphjaph, line 970: Instead of setting burn, maybe explicitly add an unrecognized even tag. I know this burn thing does that but maybe being explicit here helps test the space of unrecognized even tags.

src/runes/runestone.rs
```diff
@@ -6,10 +6,16 @@ const TAG_RUNE: u128 = 2;
...
 const TAG_TERM: u128 = 6;
+const TAG_FLAGS: u128 = 8;
+
+const FLAG_ETCH: u128 = 1;
 
 #[allow(unused)]
 const TAG_BURN: u128 = 256;
 
+#[allow(unused)]
+const TAG_NOP: u128 = 257;
```
raphjaph, line 17: We should document what exactly these flags mean.

src/runes/runestone.rs
```diff
@@ -68,38 +74,51 @@ impl Runestone {
...
-      rune: Rune(rune),
-      symbol: fields
-        .remove(&TAG_SYMBOL)
-        .and_then(|symbol| u32::try_from(symbol).ok())
-        .and_then(char::from_u32),
-      term: fields
-        .remove(&TAG_TERM)
-        .and_then(|term| u32::try_from(term).ok()),
-    });
+    let flags = fields.remove(&TAG_FLAGS).unwrap_or_default();
```
raphjaph, line 77: What is the reason for a tag flag?

raphjaph (changes requested):
There's still some details I don't understand.

ordinalOS:
`struct Focus {
    runes: u32,
    charmed_sats: u32,
}

impl Focus {
    // As the alchemist turns lead to gold.
    fn transmute(&mut self) {
        self.charmed_sats += self.runes;
        self.runes = 0;
        println!("Transmutation complete: {} charmed sats", self.charmed_sats);
    }
}

fn main() {
    let mut my_focus = Focus {
        runes: 10,
        charmed_sats: 5,
    };
    my_focus.transmute();
}`

src/runes/runestone.rs
casey: The thinking here is that we need to indicate that the runestone is an etching, which we could do by having a `TAG_ETCHING` with an empty value. That's a bit inefficient though, since we only need one bit. Instead we introduce `TAG_FLAGS` which is a bitfield of tags, and use the least significant bit to indicate that it's an etching. Later, we can add new flags if we need them.

raphjaph (approved):

t4t5:
What's the reason for reserving these runes?

raphjaph:
@casey should really start writing dome docs ;) 

If I remember correctly an etching may not specify a name and will then get a rune name assigned from this range. This is for  when you don't care about the name and just want to issue a rune.

t4t5:
> @casey should really start writing dome docs ;)
> 
> If I remember correctly an etching may not specify a name and will then get a rune name assigned from this range. This is for when you don't care about the name and just want to issue a rune.

Ahh that makes sense! Cool feature. Thank you @raphjaph! 🙏

PR #2834, opened 2023-12-08
PSBT functions added to the check.rs and receive.rs modified

DavidYashar:
the check.rs is created for the functions of buy and sell to follow PSBT flag

the receive.rs is modified


DavidYashar:
PSBT functions added to the check.rs and receive.rs modified

PR #2848, opened 2023-12-10
Added Inscription Eras

BennyTheDev:
### Abstract

Eras allow to filter inscriptions by ord versions that are compatible with a specific era or ranges of eras.

This PR makes Ordinal hard-forks (aka "freeze"-forks) that are currently floating around obsolete. It gives builders, creators and collectors the information they need to determine in what "era" an inscription has been made or is compatible with.

This is important because different projects started in different eras and committed their projects to those. This helps to spot invalid inscriptions as of the creator's era-committment and prevents confusion and conflicts moving forward.

It also allows 3rd-party indexers to filter by eras of interest instead of relying on forks, while benefitting from all current & future inscription types as well as maintenance upgrades.

### Rationale

Currently, a part of the Ordinals space is very reluctant to move forward with ord wallet updates and sticks with ord wallet version 0.9. Many marketplaces, projects and Ordinal collectors are not aware of this and this poses several risks:

- Confusion about different indexing results across marketplaces that either go by such forks or the ord wallet main branch (e.g. post-jubilee, inscription numbers and many more)
- Security & performance risks because of improperly maintained forks
- Hard-Forking but not changing the "ord" tag in the witnessdata envelope vs. "source of truth" (what is a valid inscription)

By the introduction of eras in the ord wallet, projects, creators and collectors can easily verify in what era an inscription has been made and benefit from current & future updates from the original ord wallet branch.

Having this information allows for cross-compatibility between different projects that are made in different era should that be of importance.

This PR does not alter the behavior of the ord wallet or changes what is an inscription in any way. It's purely for filtering and verifying the eras of interest.

### Eras covered in this PR

Eras are reflected in the inscription html view as well as in the inscription JSON endpoint:

era 0: genesis (all inscriptions valid as of 0.9)
era 1: awakening (...as of 0.9)
era 2: stutter (...as of 0.12.1)
era x: any new inscription type that is being discovered/recognized moving forward will get its era ID upon each ord wallet release

### Conclusion

Eras are a reliable way to disincentivze freeze-forks as they allow the use of the original ord wallet branch moving forward. Hard-Forks should under all circumstances change the "ord" tag in the envelope to avoid confusion about what a valid inscription actually is.

Should this PR not being included in the main-branch, I am willing to maintain this branch with all future ord wallet updates such that projects that care about this can benefit from it.

### DIFF
--- a/deploy/setup
+++ b/deploy/setup
@@ -128,7 +128,7 @@ systemctl daemon-reload
 systemctl enable ord
 systemctl restart ord
 
-while ! curl --fail https://$DOMAIN/status; do
+while ! curl --fail https://$DOMAIN/status > /dev/null; do
   echo "Waiting for ord at https://$DOMAIN/status…"
   sleep 1
 done
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -51,7 +51,7 @@ impl Chain {
     match self {
       Self::Mainnet => 824544,
       Self::Regtest => 110,
-      Self::Signet => 170000,
+      Self::Signet => 175392,
       Self::Testnet => 2544192,
     }
   }
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -51,7 +51,7 @@ impl Chain {
     match self {
       Self::Mainnet => 824544,
       Self::Regtest => 110,
-      Self::Signet => 175392,
+      Self::Signet => 170000,
       Self::Testnet => 2544192,
     }
   }
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -148,6 +148,7 @@ pub(crate) struct InscriptionEntry {
   pub(crate) sat: Option<Sat>,
   pub(crate) sequence_number: u32,
   pub(crate) timestamp: u32,
+  pub(crate) era: u16,
 }
 
 pub(crate) type InscriptionEntryValue = (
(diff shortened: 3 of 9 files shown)


tgkanrewrus:
ебашь бени!

ryguy077:
This is Genius. I support 🫡

NRCHST:
Possibility to add eras to avoid forks is a unique chance for a stronger protocol

I support

casey:
What are the major ordinal forks out there which are frozen at a previous release?

I worry a bit that these eras won't be perfectly accurate, which would make me nervous if people rely on them.

Also, I believe that it is wrong as-is in the case of multiple sources of curses. For example, let's say something has both a duplicate field, as well as a stutter. Currently, it gets era 1, when in fact, it should get era 2, since the stutter would prevent it from being recognized until the stuttering curse was added. Also, I don't know if the current state of the codebase can distinguish correctly between all historical curses/edge cases.

I think the subtleties here make me not want to merge this, so that people don't get the mistaken impression that they can rely on it.

Another issue is that testing this will be very hard. If we make change, you would basically need to test against every version of ord which was known to recognize new inscriptions, to make sure that the eras stayed accurate.

casey:
I think it would be worth first opening an issue to discuss the issue of frozen forks out there. Like what's BRC-20 doing these days? Knowing what the forks were, why they had forked, where they had forked, and what their needs were, would make me more confident that something like this was meeting those needs.

BennyTheDev:
done

https://github.com/ordinals/ord/issues/2851

> I think it would be worth first opening an issue to discuss the issue of frozen forks out there. Like what's BRC-20 doing these days? Knowing what the forks were, why they had forked, where they had forked, and what their needs were, would make me more confident that something like this was meeting those needs.

PR #2853, opened 2023-12-11
Make varint decoding infallible

casey:
This removes two error paths:

- Truncated varints, i.e., the last byte has the continuation bit set
- Varints over the maximum

I think removing these error paths makes parsing slightly easier to implement and more robust.


raphjaph (approved):
These subtleties should be documented somewhere. Probably best right in the code with /// comments

PR #2892, opened 2023-12-23
Encode inscriptions with minimal opcodes

### LINKED ISSUE #1403: Use minimal data-pushes

casey:
Just opening this to see what it would look like, not tests yet. Encodes envelopes with minimal opcodes, which saves 1 byte per tag.

### DIFF
--- a/src/inscription.rs
+++ b/src/inscription.rs
@@ -118,6 +118,51 @@ impl Inscription {
     bytes
   }
 
+  pub(crate) fn payload(&self) -> Vec<Vec<u8>> {
+    let mut payload = Vec::new();
+
+    if let Some(content_type) = self.content_type.clone() {
+      payload.push(envelope::CONTENT_TYPE_TAG.into());
+      payload.push(content_type);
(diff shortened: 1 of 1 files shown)


casey:
Yeah, good point. I think nobody cares about the vindicated charm either way, and so might as well get rid of it.

casey:
This is a good PR but I'm never going to finish it 😭

PR #2904, opened 2023-12-24
feat(redb): add repair progress

nikicat:
This PR adds progress bar for database repair. It doesn't work smoothly though: due to redb limitation, only 30%, 60% and 90% steps exist.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -21,13 +21,13 @@ use {
   log::log_enabled,
   redb::{
     Database, DatabaseError, MultimapTable, MultimapTableDefinition, MultimapTableHandle,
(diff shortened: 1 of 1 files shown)


casey:
Does this need to be in a background thread? Can we update the progress bar directly in the callback?

nikicat:
> Does this need to be in a background thread? Can we update the progress bar directly in the callback?

We can't because progress callback should be an `Fn`, not `FnMut`

casey:
In that case, try wrapping the progress bar in a Mutex.

casey:
Nice, merged!

PR #2912, opened 2023-12-25
Delegation

casey:
Allow inscriptions to nominate a delegate inscription, whose content is displayed instead of the inscription's.


devords:
Could this potentially be used for generative art inscriptions that use inscription ID as seed? 🤔

casey:
Yah, definitely. It would make them very efficient.

lyndoco:
As plain txt, as opposed to wrapping it in an html container?

casey:
@lyndoco What do you mean?

lyndoco:
@casey are you saying that people are inscribing .txt files that simply contain "/content/asdfinsertidherepenisasdfi0"?

and you just wanna resolve that to rendering whatever the pointer is referencing?

Yeah, that would be dope - I'd be using this convention immediately. 

You'd be saving me the scripting and html wrappers that I'm currently using...

and sum sweet sats

casey:
> @casey are you saying that people are inscribing .txt files that simply contain "/content/asdfinsertidherepenisasdfi0"?

Yes, they're creating inscriptions whose content are just the text `/content/ID`. They set the content type to the type of the inscription that they point to.

> and you just wanna resolve that to rendering whatever the pointer is referencing?

I don't want to support it directly, since it's a pretty ugly hack. Instead this PR adds a new tag, which contains a binary inscription ID, which, if present, is displayed instead of the inscription contents. So same thing, but more efficient.

lyndoco:
@casey are you sure you don't wanna just parse all .txt files instead?

Yeah nah maybe we'll go with your idea...

devords:
1. It'd be cool to have a fallback when the delegation tag is present, but the value isn't. Maybe, the first parent?
2. If the delegated inscription doesn't exist, perhaps it's better to attempt to serve the inscription's inscribed content instead of throwing an error. That could make it easy to do delayed reveals.

casey:
> 1. It'd be cool to have a fallback when the delegation tag is present, but the value isn't. Maybe, the first parent?

Can you elaborate? I'm not sure I understand what you mean.

> 2. If the delegated inscription doesn't exist, perhaps it's better to attempt to serve the inscription's inscribed content instead of throwing an error. That could make it easy to do delayed reveals.

That seems reasonable.

devords:
> > 1. It'd be cool to have a fallback when the delegation tag is present, but the value isn't. Maybe, the first parent?
> 
> Can you elaborate? I'm not sure I understand what you mean.

Say, the envelope looks like this:

```
OP_FALSE
OP_IF
  OP_PUSH "ord"
  OP_PUSH 11
OP_ENDIF
```

In that case, could attempt delegating to the first parent or any other sensible default. Delegating to parent would make creating children inscriptions even cheaper, where this makes sense. That said, I am not fully sure if the added obscurity is worth the blockspace savings.

raphjaph (approved):

casey:
> FYI, nested delegation does not appear to be working properly yet, i. e. anointing a delegate that itself has a delegate.

I'm not sure we should support this. If we do, the ord server could possibly wind up having to load a bunch of delegates to return the content of one inscription.

casey:
Did you have a use-case in mind?

casey:
> I think a workaround could be resolving the full delegation path at indexing time, such that if such an inscription got used as a delegate later, instead of following the resolution path again, you just immediately jump to its destination.

I think it's not possible to do this at index time, since an inscription may have a delegate that isn't inscribed yet. (Maybe people will want to do this for delayed reveals.) But we could, whenever looking up a delegate, if it's found, store the resolved delegate in the index, so that we never look up the same delegate twice.

casey:
Created #2922 to track recursive delegation.

PR #2914, opened 2023-12-26
Add cmd to cancel pending transaction

chadwick2143:
Use `ord wallet cancel`  to cancel a pending transaction.

This cmd will try to generate a new transaction using RBF (replace-by-fee).

It's very useful when you want to cancel a pending inscribe. You can cancel the commit transaction with this cmd, than the reveal transaction will fail automatically.

Use this cmd to save your gas.

### DIFF
--- a/src/subcommand/wallet/cancel.rs
+++ b/src/subcommand/wallet/cancel.rs
@@ -26,7 +26,7 @@ impl Cancel {
       )));
     }
 
-    let old_transaction_info = client.get_raw_transaction_info(&self.transaction, None)?;
+    let old_transaction_info = client.decode_raw_transaction(&old_transaction.hex, None)?;
     let vin: Vec<CreateRawTransactionInput> = old_transaction_info
       .vin
       .into_iter()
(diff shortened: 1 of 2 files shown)


src/subcommand/wallet/cancel.rs
```diff
@@ -0,0 +1,74 @@
+use super::*;
+use bitcoincore_rpc::bitcoincore_rpc_json::CreateRawTransactionInput;
```
raphjaph, line 2: Combine these into one use statement.

src/subcommand/wallet/cancel.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Debug, Parser, Clone)]
+pub(crate) struct Cancel {
+  transaction: Txid,
+  #[arg(long, help = "Bump fee by ratio. Default `0.1`.")]
+  bump_fee_ratio: Option<f64>,
```
raphjaph, line 8: You can set the default like so I think.
```suggestion
  #[arg(long, defautl: 0.1, help = "Bump fee by ratio.")]
  bump_fee_ratio: f64,
```

raphjaph (changes requested):
Left some comments but don't understand how this works. I don't think miners/the mempool will accept a transaction that pays less fees than the original one. To cancel a reveal transaction you would have to spend the commit input. This command should also be scoped down to only cancel inscribe transactions and not allow to cancel send as well.

Issue #2736, opened 2023-11-23
Read only server mode


brianorwhatever:
It should be possible to use the wallet while the http server is running as if I understand correctly the server only ever reads. I am currently writing a program that needs to query the server for some things while also running wallet commands. To accomplish this I am creating two `ord` indexes. One for the server to always be running on and another for the wallet.. This is obviously not great and I'd like to be able to use both server and wallet at the same time. Is there anyway to accomplish what I'm trying to do now or is there a path towards that?

raphjaph:
Yeah this would be kind of nice. Will put it in the Backlog

casey:
This works now, you can use the wallet with the server running in the background.

PR #2929, opened 2023-12-31
Wallet uses JSON RPC

raphjaph:
Starts work on converting the wallet to talk to the index over JSON endpoints on the server. This way server and wallet can run concurrently 

#2736


raphjaph:
@casey This is ready for a review. It's kinda gnarly but there was no way to break it up into smaller pieces. Lord have mercy on thy soul.

casey (changes requested):
Let's also modularize, putting Wallet into its own module, `src/wallet.rs`, as well as not spinning up a server when running a wallet runs, requiring that an existing server be running.

raphjaph:
@casey Ready to review.

The integration tests now run really fast. Made our testing harness start threads instead of child processes. Also I remove the `ord preview` command since I think nobody uses that.

casey (approved):

PR #2931, opened 2023-12-31
Try using a single-threaded-runtime

casey:
We probably won't merge this, but I wanted to see what impact a single-threaded runtime had on performance.


casey:
This nuked performance 😅

PR #2938, opened 2024-01-02
Make `FundRawTransactionOptions` public

lateminer:
`FundRawTransactionOptions` struct should be public to be reachable from the other modules.


raphjaph (approved):
lgtm!

PR #2941, opened 2024-01-03
implemented JSON formatting for Rune responses

lugondev:
In this update, JSON formatting has been added to Rune responses to enhance data readability and interoperability.

thanks


lugondev:
Updated the Runes list response to return in JSON format. Thanks in advance for reviewing and merging.

src/templates/rune.rs
```diff
@@ -7,6 +7,15 @@ pub(crate) struct RuneHtml {
   pub(crate) parent: Option<InscriptionId>,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct RuneJson {
+  pub entry: RuneEntry,
+  pub id: RuneId,
+  pub height: u32,
+  pub index: u16,
```
raphjaph, line 15:
```suggestion
```
This is contained in the RuneId so no need to add it here

Since RuneHtml and RuneJson are the same now you could just derive the same traits on RuneHtml and then type alias it to RuneJson. Something like this:

```rust
type RuneJson = RuneHtml;

#[derive(Boilerplate, Debug, PartialEq, Serialize, Deserialize)]
pub(crate) struct RuneHtml {
  pub(crate) entry: RuneEntry,
  pub(crate) id: RuneId,
  pub(crate) parent: Option<InscriptionId>,
}

```

src/index.rs
```diff
@@ -913,6 +913,45 @@ impl Index {
...
+
+    let rune_id = RuneId::load(id);
+    Ok(Some(RuneJson {
+      entry,
+      height: rune_id.height,
+      index: rune_id.index,
+      id: RuneId::load(id),
+      parent,
+    }))
+  }
```
raphjaph, line 954: This function is almost identical to rune_html and combined with the suggestion below you can construct the Html and Json in the same function.

raphjaph (changes requested):
Good start! Left some comments

src/index.rs
lugondev: thanks for your suggestion

raphjaph:
Why did you close this?

lugondev:
@raphjaph please review again my pull request

lugondev:
> Why did you close this?
> 
> If you're having trouble I can just integrate my suggestions

haha, That was a mistake of mine :D

src/subcommand/server.rs
```diff
@@ -617,31 +617,50 @@ impl Server {
...
-    Ok(
+    let rune = index
+      .rune_html(spaced_rune.rune)?
+      .ok_or_not_found(|| format!("rune {spaced_rune}"))?;
+    Ok(if accept_json {
+      Json(RuneJson {
+        parent: rune.parent,
+        id: rune.id,
+        entry: rune.entry,
+      })
```
raphjaph, line 636:
```suggestion
    Ok(if accept_json {
      Json(index
      .rune_html(spaced_rune.rune)?).into_response()
```

raphjaph (changes requested):
Looking good, two tests for the endpoints would also be good.

Can write those for you if you don't know where/how

lugondev:
Do you have any further suggestions for me?

raphjaph:
@lugondev I added some tests.

Still have to fix one thing but, have a look!

raphjaph (approved):
It's good to go now

tests/json_api.rs
```diff
@@ -368,15 +368,21 @@ fn get_block() {
...
-  let rpc_server = test_bitcoincore_rpc::spawn();
+  let rpc_server = test_bitcoincore_rpc::builder()
+    .network(Network::Regtest)
+    .build();
 
   create_wallet(&rpc_server);
+  rpc_server.mine_blocks(1);
 
-  inscribe(&rpc_server);
+  inscribe_regtest(&rpc_server);
```
casey, line 378: Let's get the network from the RPC server and use that in `inscribe`

tests/test_server.rs
```diff
@@ -27,18 +27,19 @@ impl TestServer {
     server_args: &[&str],
   ) -> Self {
     let tempdir = TempDir::new().unwrap();
-    fs::write(tempdir.path().join(".cookie"), "foo:bar").unwrap();
+    let cookie_file = tempdir.path().join(".cookie");
```
casey, line 30: Instead of using `--cookie-file`, can we write the cookie to `regtest/.cookie`?

casey (changes requested):

casey (approved):

Issue #2948, opened 2024-01-03
Batch Inscribing Docs have Sat commented


mj10021:
As written here: https://docs.ordinals.com/guides/batch-inscribing.html the sat field is left commented out for the inscription, which I believe is an error

PR #2949, opened 2024-01-03
fix commented sat field on batch inscriptions guide

mj10021:
fixes issue #2948 , I believe the `#` commenting out the sat field is a mistake and makes the example a little confusing


raphjaph:
This is on purpose since the mode `separate-outputs` does not support using the `sat` field. It only works in `same sat` mode. Wanted the batch file to be internally consistent.

PR #2950, opened 2024-01-03
VINDICATION

casey:
This PR adds a ❤️‍🔥 charm, named "vindicated". The intent is to demonstrate that, with only a small change to ord, we can support BRC-20 indexers that wish to update to a new version of `ord` at a later date, by letting them skip inscriptions with the vindicated charm until then.

This needs tests before being merged. We should also consider whether or not we want to make this charm visible to end users. Users may prefer not to have an additional charm on their inscriptions.

### DIFF
--- a/src/index/updater/inscription_updater.rs
+++ b/src/index/updater/inscription_updater.rs
@@ -175,10 +175,10 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
                 .value(),
             );
 
-            let initial_inscription_is_cursed =
+            let initial_inscription_was_cursed_or_vindicated =
               entry.inscription_number < 0 || Charm::Vindicated.is_set(entry.charms);
 
-            if initial_inscription_is_cursed {
+            if initial_inscription_was_cursed_or_vindicated {
               None
             } else {
               Some(Curse::Reinscription)
--- a/src/index/updater/inscription_updater.rs
+++ b/src/index/updater/inscription_updater.rs
@@ -188,8 +188,6 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
           None
         };
 
-        let vindicated = jubilant && curse.is_some();
-
         let unbound = current_input_value == 0
           || curse == Some(Curse::UnrecognizedEvenField)
           || inscription.payload.unrecognized_even_field;
@@ -211,7 +209,7 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
             pointer: inscription.payload.pointer(),
             reinscription: inscribed_offsets.get(&offset).is_some(),
             unbound,
-            vindicated,
+            vindicated: curse.is_some() && jubilant,
           },
         });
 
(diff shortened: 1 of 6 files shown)


raphjaph:
Why not call it blessed and then use 😇 or 💸 or ✨or 🕍 for the emoji?

src/index/updater/inscription_updater.rs
```diff
@@ -201,13 +203,14 @@ impl<'a, 'db, 'tx> InscriptionUpdater<'a, 'db, 'tx> {
           inscription_id,
           offset,
           origin: Origin::New {
-            reinscription: inscribed_offsets.get(&offset).is_some(),
-            cursed: curse.is_some(),
+            cursed: curse.is_some() && !jubilant,
             fee: 0,
```
jiedo, line 207: Re-inscribing on a vindicated inscription will continue to yield a vindicated inscription, which may inconsistent with the process before Jubilee.

```
            .inscription_number
              < 0;
            if initial_inscription_is_cursed {
```

casey:
> Why not call it blessed and then use 😇 or 💸 or ✨or 🕍 for the emoji?

I dunno, vindication is pretty lindy.

image

image

src/index/updater/inscription_updater.rs
casey: I don't think I understand.

Can you elaborate? The specific sequence of transactions that would cause a problem would be helpful.

DrJingLee:
> 💸

💸 looks cooool lol

src/index/updater/inscription_updater.rs
jiedo: Before Jubilee, first inscribe a cursed inscription using Pushnum. Continuing to inscribe on the same sat will result in a blessed inscription.

After Jubilee, first inscribe a Vindicate inscription using Pushnum, and continuing to inscribe on the same sat will result in a Vindicate inscription, not a blessed one.

samedcildir: We can fix that by also checking the vindicated charm. `initial_inscription_is_cursed` should be true if either inscription_number < 0 or inscription has vindicated charm.

yann-sjtu: > We can fix that by also checking the vindicated charm. `initial_inscription_is_cursed` should be true if either inscription_number < 0 or inscription has vindicated charm.
yes, we also think this is what we want, and `vindicated` will be consistant with release 0.9 after being fixed
```Rust
let entry = InscriptionEntry::load(
  self
    .sequence_number_to_entry
    .get(initial_inscription_sequence_number)?
    .unwrap()
    .value(),
);

let initial_inscription_is_cursed =
  entry.inscription_number < 0 || Charm::Vindicated.is_set(entry.charms);
```

raphjaph (approved):

src/index/updater/inscription_updater.rs
casey: Nice, thanks for the explanation! This is now fixed.

casey:
Yup, that's correct, minimal op codes weren't accepted by the first version of ord (which lead to much confusion), and were cursed when supported.

PR #2988, opened 2024-01-08
Feat/https transport

viethapascal:
Add custom transport using hyper tls in order to enable request data from https server


raphjaph:
I don't understand what this is supposed to do. Is this maybe part of a larger PR and you forgot to commit the other files?

viethapascal:
> I don't understand what this is supposed to do. Is this maybe part of a larger PR and you forgot to commit the other files?

I used it in my local service to connect bitcoin node which running remotely. I think this may helpful, let me add it to an option when start ord

src/index.rs
```diff
@@ -1,3 +1,4 @@
+use bitcoincore_rpc::Auth;
```
raphjaph, line 1: Would be great to combine this with the other `use` statements.

src/index.rs
```diff
@@ -5828,3 +5831,18 @@ mod tests {
...
+  let user = "__cookie__".to_string();
+  let pass = "".to_string();
+  let auth_token = Auth::UserPass(user, pass);
+  let t = transport::CustomTransport::new(rpc_url, auth_token);
+  let http_client = client::Client::with_transport(t);
+
+  let client = Client::from_jsonrpc(http_client);
+
+  let data = client.get_best_block_hash().unwrap_or_else(|err|panic!("failed to send request. detail: {}", err));
+  println!("{}", data)
```
raphjaph, line 5847: This test doesn't really assert that it successfullt connected.

src/options.rs
```diff
@@ -228,9 +230,12 @@ impl Options {
       );
     }
 
-    let client = Client::new(&rpc_url, auth)
-      .with_context(|| format!("failed to connect to Bitcoin Core RPC at {rpc_url}"))?;
+    // let client = Client::new(&rpc_url, auth);
+    //   .with_context(|| format!("failed to connect to Bitcoin Core RPC at {rpc_url}"))?;
+    let t = transport::CustomTransport::new(&rpc_url, auth);
+    let http_client = client::Client::with_transport(t);
```
raphjaph, line 236: This should be confiigured via the options, since we don't always want to use this transport.

raphjaph (commented):
This is nice but this should probably be added to the upstream rust crate `bitcoincore_rpc`. Going to close this for that reason.

Here's the repo: https://github.com/rust-bitcoin/rust-bitcoincore-rpc.

PR #2999, opened 2024-01-09
Add option to retain sat index for spent outputs

casey:
Add `--index-spent-sats` option to keep sat index entries for spent outputs.

This is how the index originally worked, but it was too expensive, taking forever to build and swelling the size of the index to over 10 terabytes, if I recall correctly.

This is untested, and is added for anyone who wants to try it out. It will likely require a crazily beefy machine to run, like 256GB+ of RAM, and 10s or 100s of terabytes of disk.


casey:
I thought this would be easy, but it turned into a pretty big PR. The reason is that we had the weird `List` enum which both contained sat ranges, and indicated whether an output was spent or unspent. `--index-spent-sats` required revisiting that, since you could now have a spent output, but still have the sat ranges.

raphjaph (approved):

GaloisField2718:
Hello @casey,
If we don't index spent sats how is it possible to know who is owning which satoshi? 
> Add `--index-spent-sats` option to keep sat index entries for spent outputs.
> 
> This is how the index originally worked, but it was too expensive, taking forever to build and swelling the size of the index to over 10 terabytes, if I recall correctly.

From your BitcoinTalk post: [Unique serial number for every single satoshi](https://bitcointalk.org/index.php?topic=117224.20), it sounds like index each spent satoshis. 
> 
> This is untested, and is added for anyone who wants to try it out. It will likely require a crazily beefy machine to run, like 256GB+ of RAM, and 10s or 100s of terabytes of disk.

So the current indexer doesn't work concretely with this mechanism? 

What index spent sats allow following this update?

casey:
This option only pertains to whether or not prior locations of sats are stored, and does not affect whether or not the current location of each sat is stored.

GaloisField2718:
Ok thanks, so it's keeping track of all trades of satoshis from day 1 on Bitcoin if I understand correctly?

casey:
It's keeping a record of where sats *used* to be. Without `--index-spent-sats`, when a UTXO is spent, the database entry for that UTXO is deleted, to save space. With `--index-spent-sats`, database entries for spent UTXOs are retained, so you can query where sats used to be.

GaloisField2718:
Ok very clear. Thank you very much for your precisions and your time!

PR #3004, opened 2024-01-12
support tx json response

DaviRain-Su:


src/templates/transaction.rs
```diff
@@ -16,6 +16,36 @@ impl PageContent for TransactionHtml {
   }
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
```
raphjaph, line 19: I think we can just derive these on the `TransactionsHtml` struct directly and then type alias like so:

```rust
pub type TransactionJson = TransactionHtml;
```

src/subcommand/server.rs
```diff
@@ -780,7 +781,17 @@ impl Server {
...
-      Ok(
+      Ok(if accept_json {
+        Json(TransactionJson::new(
+          blockhash,
+          server_config.chain,
+          index.get_etching(txid)?,
+          inscription_count,
+          transaction,
+          txid,
+        ))
```
raphjaph, line 792: Just create the struct directly without going through a `new()` funciton.

raphjaph (changes requested):
This is good start!

Could you add a test for this in `tests/json_rpc.rs`?

tests/json_api.rs
```diff
@@ -389,6 +389,37 @@ fn get_block() {
...
 }
 
+#[test]
+fn get_transaction() {
+  let rpc_server = test_bitcoincore_rpc::spawn();
+
+  rpc_server.mine_blocks(1);
+
+  let response = TestServer::spawn_with_server_args(&rpc_server, &[], &["--enable-json-api"])
+    .json_request("/tx/0");
```
DaviRain-Su, line 399: I don't know how to set this txid? can you help me.

src/subcommand/server.rs
DaviRain-Su: fixed

src/templates/transaction.rs
DaviRain-Su: fixed

raphjaph (approved):
Hey! I just added the tests for you. Have a look if this is good!

Issue #2990, opened 2024-01-09
Feature request: `ord wallet restore` to accept user input for MNEMONIC


oxSaturn:
Currently you have to restore a wallet like below:

```
ord wallet restore "MY SEED HERE"
```
Problem is the shell history will keep the SEED, which is bad practice IMO.

Would like `ord wallet restore` to receive user input from stdin when no `MNEMONIC` is passed through cli argument:

```
$ ord wallet restore
...
Please input your seed below:
...
```
Does it make sense?

so7ow:
Very good idea. 👍🏻

raphjaph:
Agreed, that would be good enhancement :)

PR #3023, opened 2024-01-16
add prompt for seed when restoring wallet

mj10021:
Fixes issue #2990 by allowing a seed to be entered to stdin when restoring a wallet.  I'm not sure if I got the type coercion/checking exactly right here, so would appreciate any feedback/help.  Thanks!

### DIFF
--- a/src/subcommand/wallet/restore.rs
+++ b/src/subcommand/wallet/restore.rs
@@ -31,7 +31,7 @@ impl Restore {
         wallet.initialize_from_descriptors(wallet_descriptors.descriptors)?;
       }
       Source::Mnemonic => {
-        let mnemonic = Mnemonic::parse_normalized(&buffer)?;
+        let mnemonic = Mnemonic::from_str(&buffer)?;
         wallet.initialize(mnemonic.to_seed(self.passphrase.unwrap_or_default()))?;
       }
     }
--- a/src/subcommand/wallet/restore.rs
+++ b/src/subcommand/wallet/restore.rs
@@ -1,20 +1,17 @@
 use super::*;
 
 #[derive(Debug, Parser)]
-#[clap(group(
-  ArgGroup::new("source").required(true).args(&["descriptor", "mnemonic"]))
-)]
 pub(crate) struct Restore {
-  #[arg(long, help = "Restore wallet from <DESCRIPTOR> from stdin.")]
-  descriptor: bool,
-  #[arg(long, help = "Restore wallet from <MNEMONIC> from stdin.")]
-  mnemonic: bool,
-  #[arg(
-    long,
-    requires = "mnemonic",
-    help = "Use <PASSPHRASE> when deriving wallet"
-  )]
+  #[clap(value_enum, long, help = "Restore wallet from <SOURCE> on stdin.")]
+  from: Source,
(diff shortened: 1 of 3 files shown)


src/subcommand/wallet/restore.rs
```diff
@@ -1,9 +1,10 @@
 use super::*;
+use std::io;
```
raphjaph, line 2: combine these two
```rust
use {
  super::*,
  std::io,
};
```

src/subcommand/wallet/restore.rs
```diff
@@ -14,10 +15,20 @@ pub(crate) struct Restore {
 
 impl Restore {
   pub(crate) fn run(self, wallet_name: String, options: Options) -> SubcommandResult {
+    let seed = {
+      if self.mnemonic.is_none() {
+        let mut input = String::new();
+        println!("Please input your seed below:");
+        io::stdin().read_line(&mut input).expect("failed to read mnemonic");
+        let input = input.into_bytes();
+        assert_eq!(input.len(), 64);
```
raphjaph, line 24: not sure if we need this the `Mnemonic` might already perform these checks.

raphjaph (changes requested):
This is good start! Definitely add a an integration test somewhere in `tests/wallet`. If you need help with that let me know.

raphjaph:
@mj10021 Have a look at #3048. I read the descriptor from stdin and also have some tests in there that you can use as a template.

mj10021:
thanks :) i'm away this weekend but should have done early next week

mj10021:
hey @raphjaph , just to clarify, is the stdin meant to accept a descriptor or a mnemonic?

mj10021:
> hey @raphjaph , just to clarify, is the stdin meant to accept a descriptor or a mnemonic?

Oh I see, the --descriptor flag lets you enter a descriptor through stdin without the mnemonic

mj10021:
Ok, sorry, I actually do need some clarification. I see that now there is an option to restore from stdin and tests have already been made for that. Is this PR now redundant? Or is the option to load a mnemonic from stdin in addition to the descriptor flag also wanted? Thanks!

raphjaph:
> Ok, sorry, I actually do need some clarification. I see that now there is an option to restore from stdin and tests have already been made for that. Is this PR now redundant? Or is the option to load a mnemonic from stdin in addition to the descriptor flag also wanted? Thanks!

Yes we want to load the mnemonic from stdin only as well. And def also add a separate test for that

mj10021:
> > Ok, sorry, I actually do need some clarification. I see that now there is an option to restore from stdin and tests have already been made for that. Is this PR now redundant? Or is the option to load a mnemonic from stdin in addition to the descriptor flag also wanted? Thanks!
> 
> Yes we want to load the mnemonic from stdin only as well. And def also add a separate test for that. Let me know when this is ready for review.

Ok got it thanks.  I'm not able to run the tests atm but I think this should be right?

raphjaph:
Tests aren't passing yet. Install `just` and run `just ci` to run all our tests.

mj10021:
> Tests aren't passing yet. Install `just` and run `just ci` to run all our tests.

Got it, thanks.  Should be good now.

raphjaph (approved):
LGTM

PR #3024, opened 2024-01-16
Add api router to support OpenAPI

lesterli:
## Summary

Introduce the crate [utoipa](https://crates.io/crates/utoipa) to auto-generate OpenAPI docs for the API.

## Example

The `status` function in the `src/subcommand/server/api.rs`.

## Test

* Run `ord` and connect with the bitcoin `testnet`: `ord -t server -j`
* Check the OpenAPI docs via the URL: http://localhost/api/v1/api-docs/openapi.json
* Access the REST API: http://localhost:8080/api/v1/status

Output
```json
{"code":0,"msg":"ok","data":{"blessed_inscriptions":0,"cursed_inscriptions":0,"chain":"testnet","height":2573099,"inscriptions":0,"lost_sats":0,"minimum_rune_for_next_block":"ZCUKDPQNT","rune_index":true,"runes":1858,"sat_index":false,"started":"2024-01-16T13:12:44.150040217Z","transaction_index":false,"unrecoverably_reorged":false,"uptime":{"secs":30,"nanos":888723119}}}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1648,6 +1648,7 @@ checksum = "d530e1a18b1cb4c484e6e34556a0d948706958449fca0cab753d649f2bce3d1f"
 dependencies = [
  "equivalent",
  "hashbrown",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


casey:
I think that doing this would be a big change that we wouldn't want to commit to. If we use annotations for the routes, these annotations can get out of sync with the actual paths which are used with the router.

PR #3027, opened 2024-01-16
Allow inscribing a delegate without any batch or a file

devords:
This will allow using `wallet inscribe` without any batch or a file, resulting in smaller tx size.

Todo:
- [ ] Add tests
- [ ] Fail if all 3 options (file, batch and delegate) are missing

### DIFF
--- a/src/inscriptions/inscription.rs
+++ b/src/inscriptions/inscription.rs
@@ -111,6 +111,26 @@ impl Inscription {
     })
   }
 
+  pub(crate) fn without_file(
+    delegate: Option<InscriptionId>,
+    metadata: Option<Vec<u8>>,
+    metaprotocol: Option<String>,
+    parent: Option<InscriptionId>,
+    pointer: Option<u64>,
+  ) -> Result<Self, Error> {
+    Ok(Self {
+      body: None,
+      content_encoding: None,
+      content_type: None,
+      delegate: delegate.map(|delegate| delegate.value()),
+      metadata,
+      metaprotocol: metaprotocol.map(|metaprotocol| metaprotocol.into_bytes()),
+      parent: parent.map(|parent| parent.value()),
+      pointer: pointer.map(Self::pointer_value),
+      ..Default::default()
+    })
+  }
+
   pub(crate) fn pointer_value(pointer: u64) -> Vec<u8> {
     let mut bytes = pointer.to_le_bytes().to_vec();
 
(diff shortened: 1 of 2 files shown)


raphjaph:
I don't understand how this would result in a smaller tx size? We use the same logic to inscribe no matter if it's a single file or a batch file with a single inscription.

src/inscriptions/inscription.rs
```diff
@@ -108,6 +108,26 @@ impl Inscription {
     })
   }
 
+  pub(crate) fn without_file(
```
raphjaph, line 111:
```suggestion
  pub(crate) fn delegate_only(
```

src/subcommand/wallet/inscribe.rs
```diff
@@ -89,8 +89,8 @@ impl Inscribe {
     let parent_info;
     let sat;
 
-    match (self.file, self.batch) {
-      (Some(file), None) => {
+    match (self.file, self.batch, self.delegate) {
+      (Some(file), None, _) => {
```
raphjaph, line 93: This would mean if we have a file and a delegate it would still inscribe the file. Same with the batch below. This should be tested.

src/subcommand/wallet/inscribe.rs
```diff
@@ -148,6 +148,33 @@ impl Inscribe {
...
       }
+      (None, None, Some(delegate)) => {
+        parent_info = wallet.get_parent_info(self.parent, &utxos)?;
+
+        postage = self.postage.unwrap_or(TARGET_POSTAGE);
+
+        ensure! {
+          wallet.inscription_exists(delegate)?,
+          "delegate {delegate} does not exist"
+        }
```
raphjaph, line 159: Let's add a flag called `--allow-non-existant-delegate` and then skip this check if it is set.

raphjaph (changes requested):
The overall logic of delegate should be that you can inscribe a delegate only with 
```
ord wallet inscribe --delegate <INSCRIPTION_ID>
```
You may inscribe a delegate that doesn't exist with:
```
ord wallet inscribe --delegate <INSCRIPTION_ID> --allow-non-existant
```
And you can also inscribe a file in addition to both of these options as a fall back.

src/subcommand/wallet/inscribe.rs
stet: what is the expected response for non-existent-delegate w/o payload and existent-delegate w/ payload?

raphjaph:
What's the status on this?

ep150de:
Had worked around this when inscribing using batch delegate, I use a 1 byte dummy file so I can minimize the extra unnecessary cost and block space, hope this PR gets progress so I don't have to do that going forward, some markets make it think the ordinal is literally a 1 byte blank file when it should show the delegate content.

PR #3029, opened 2024-01-17
chore: fix comment

zhiqiangxu:


src/index/entry.rs
```diff
@@ -354,7 +354,7 @@ impl Entry for SatRange {
 
     let raw_delta = u64::from_le_bytes([b6, b7, b8, b9, b10, 0, 0, 0]);
 
-    // 33 bit delta
+    // 37 bit delta
```
raphjaph, line 357: What makes you think this? 

The delta should be log2(50 * COIN_VALUE) = 32.2

zhiqiangxu: `raw_delta` contains `40` bit and after right shifting `3` bit, it's `37` bit. But I see your reasoning now, `4` bit of the resulting `37` bit is always 0, let's undo this change.

raphjaph (approved):

raphjaph (approved):

raphjaph (approved):

PR #3038, opened 2024-01-21
chore: also forbid `entry.destination` in `same-sat` mode

zhiqiangxu:
Since in `same-sat` mode, the `destination` is also calculated by `get_change_address` as per [here](https://github.com/ordinals/ord/blob/4b95ff655aa051578d26d755ba1532819b1a9da5/src/subcommand/wallet/inscribe/batch.rs#L646), it's more correct to forbid `entry.destination` in `same-sat` mode.


raphjaph (approved):
Thanks, this makes sense!

raphjaph (approved):

Issue #1249, opened 2023-01-16
RPC API


casey:

tiero:
gRPC/protobuf would be handy to self-generate client code in almost all languages out there, to allow to integrate ord in as many apps as possible

veryordinally:
Talked to a wallet developer, a library crate would really help - essentially taking out the "pub crate". @casey we had talked about that months ago (I wanted to use as a library as well), and I remember you had some concerns, that I do not remember.

YaroShkvorets:
Also could use a library crate.

raphjaph:
It technically already is a library crate but we haven't designed it in a way that lends itself to be used as a library. We do want to make it more useful though. 

I'd be interested to know which parts of the codebase would be most useful for you? Just to know where to start

elix1er:
bump

YaroShkvorets:
> I'd be interested to know which parts of the codebase would be most useful for you? Just to know where to start

In my case I want to use `ParsedEnvelope::from_transaction()` but it's private.

raphjaph:
> In my case I want to use `ParsedEnvelope::from_transaction()` but it's private.

Ok thanks, that is a useful one.

Just going to connect this #2964, which would be the first step.

PR #3042, opened 2024-01-24
Make inscriptions module pub

raphjaph:
Made the `inscriptions` module and some structs `pub`. Did it in a pretty haphazard way and would love feedback if I should use `pub use` in `src/inscriptions.rs` so that there is less indirection (less submodules to include in the import statements). Basically any feedback on good crate/library design.

 Let me know if this is useful to you guys @rot13maxi @JeremyRubin


JeremyRubin:
I think you should split out things into a couple different crates -- inscription for example should be really "tiny" to make it includable in Miniscript as an optional dep.

raphjaph:
Ok thanks for the feedback! I think splitting out into crates makes sense but probably a bit too much work for the next release. 

Possible sub crates:

- `ordinals`/`sats`
- `inscriptions`
- `runes`
- `wallet`
- `ord` (binary crate)

casey:
What functionality would a crate need to expose for use in miniscript in order to be useful?

JeremyRubin:
i doubt mainline miniscript will

1) take a non-rust-bitcoin-owned dep
2) add inscriptions to the spec, as I've done in Sapio

That said, were you to get buy in, the functionality is basically just what's in the inscription module, which is pretty minimal.

raphjaph:
I'm going to close this since we are working on #3079 and #1249.

Issue #3015, opened 2024-01-15
Recursive Endpoint for Self-Referencing and Referencing other Inscriptions, reducing costs, enhanced accessibility and efficiency


ORDHeavens:
Proposal to introduce a new recursive endpoint. This feature would enable inscriptions to reference information about themselves without prior knowledge of the inscription IDm and information about other inscriptions. Thus, significantly reducing costs, optimizing block space usage and reduce transactions load further lowering price globally. 

Old command
```
/̶r̶/̶s̶a̶t̶/̶c̶u̶r̶r̶e̶n̶t̶:̶ ̶O̶u̶t̶p̶u̶t̶s̶ ̶t̶h̶e̶ ̶c̶u̶r̶r̶e̶n̶t̶ ̶i̶n̶s̶c̶r̶i̶p̶t̶i̶o̶n̶'̶s̶ ̶s̶a̶t̶ ̶n̶u̶m̶b̶e̶r̶.̶
̶/̶r̶/̶s̶a̶t̶/̶i̶d̶/̶c̶u̶r̶r̶e̶n̶t̶:̶ ̶O̶u̶t̶p̶u̶t̶s̶ ̶t̶h̶e̶ ̶c̶u̶r̶r̶e̶n̶t̶ ̶i̶n̶s̶c̶r̶i̶p̶t̶i̶o̶n̶'̶s̶ ̶f̶u̶l̶l̶ ̶I̶D̶.̶
```

Improved command proposed by @RabbiGains-eth 
`/r/self` returning:
```
{ 
   inscription_id: 371b1fe39a2f0463008ee1f9ab75e0af7b3d80beeb2be6eb5cfed9d848e48b54i0,
   parent_id: 3518fb0f9112542df40f31d4c7dd6e9366acd8401ce3b4cc9b6f5c00d6d21fb3i0,
   sat_number: 45026248133,
   utxo: cb7ca4e9a8c827bbbe4dcc73700b2fda39242c234f2020147d756cc3fe2787f8:2,
   genesis_blockheight: 818927
}
```

**Benefits**
_Cost Reduction:_ Up to circa 75% cost savings for inscriptions requiring their sat number or self-reference.
_Block Space Efficiency:_ Potential reduction in transaction and block space usage by up to 33% on such type of inscriptions.
_Enhanced Flexibility:_ Opens up new possibilities for creative and broader applications of inscriptions, decentralized/user mints (always more and more used).

[...]

PR #3054, opened 2024-01-27
Recursive Endpoint for Self-Referencing and Referencing other Inscriptions /r/info 

jerryfane:
This commit implements a new recursive endpoint (/r/info) to self-reference and reference other inscriptions, providing basic information such as:

- Inscription ID
- Inscription Number
- Inscription Sat Number
- Genesis Block Height
- Parent Inscription ID
- Timestamp
- Fee
- Charms

Fixes issue #3015.


devords:
@jerryfane Believe this is a duplicate of #2628

PR #3063, opened 2024-01-28
Add outgoing info and dry run when sending

### LINKED ISSUE #2887: Expand --dry-run Functionality
When using this flag, can we add an option to output the commit/inscribe PSBTs to verify in Sparrow or deconstruct with Bitcoin-core commands to verify the transactions manually before broadcasting?

Use case:
Say I have a wallet with multiple UTXOs containing uncommon sats. I'd like to manually confirm that these aren't getting accidentally spent.

Alternative:
Add the ability to define "payment" UTXOs for network fees, in addition to --sat-point

raphjaph:
Fixes #2887

### DIFF
--- a/src/outgoing.rs
+++ b/src/outgoing.rs
@@ -107,6 +107,7 @@ mod tests {
     #[track_caller]
     fn case(s: &str, outgoing: Outgoing) {
       assert_eq!(s.parse::<Outgoing>().unwrap(), outgoing);
+      assert_eq!(s, outgoing.to_string());
     }
 
     case(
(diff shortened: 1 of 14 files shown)


src/outgoing.rs
```diff
@@ -69,6 +80,24 @@ impl FromStr for Outgoing {
...
+impl<'de> Deserialize<'de> for Outgoing {
+  fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+  where
+    D: Deserializer<'de>,
+  {
+    Ok(DeserializeFromStr::deserialize(deserializer)?.0)
+  }
+}
+
 #[cfg(test)]
```
casey, line 101: We should have a test for display, and tests for serialize/deserialize that round trip one of each variant.

casey (changes requested):
Will review more later.

casey (approved):
LGTM

PR #3067, opened 2024-01-30
Add OP_RETURN into `ord wallet send`

grdddj:
Fixes https://github.com/ordinals/ord/issues/3007:
- adds a new `op-return` flag to `ord wallet send` that allows for sending some smaller data in `OP_RETURN`

Next steps/improvements:
- adding tests
- addressing TODOs

Possible next steps:
- support even multiple `OP_RETURN`s in one transaction (if possible)
- add `OP_RETURN` possibility even to the minting transaction

### DIFF
--- a/src/subcommand/wallet/send.rs
+++ b/src/subcommand/wallet/send.rs
@@ -11,6 +11,11 @@ pub(crate) struct Send {
     help = "Target amount of postage to include with sent inscriptions. Default `10000sat`"
   )]
   pub(crate) postage: Option<Amount>,
+  #[arg(
+    long,
+    help = "Include OP_RETURN text in the transaction."
+  )]
+  pub(crate) op_return: Option<String>,
 }
 
 #[derive(Serialize, Deserialize)]
@@ -85,7 +90,7 @@ impl Send {
       Target::Postage
     };
 
-    let unsigned_transaction = TransactionBuilder::new(
+    let mut tx_builder =  TransactionBuilder::new(
       satpoint,
       inscriptions,
       unspent_outputs,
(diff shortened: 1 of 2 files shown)


raphjaph:
Let me know when this is ready for review!

grdddj:
> Could you expand a bit on the use cases for this?

The use-case is when transferring the inscription between addresses, create some possibility to **add some metadata/info to the transfer** itself.

Concrete example:
- inscriptions are held by a service in one wallet, but "owned" by multiple customers of this service (on different addresses)
- the service allows for on-chain transfer of these inscriptions/certificates, so the transfer of ownership is transparent and verifiable
- `OP_RETURN` adds the nice possibility to specify e.g. the `price in BTC` that one user paid to buy the certificate from the other user
- the idea is that these OP-returns will be nicely visible e.g. on `mempool.space` (in the future maybe even in `ord` explorer) and will add more transparency to the transfer (or any other on-chain action)

image

Marked it as a draft for now, hopefully, I will have time to finish it soon. Thanks :)

src/subcommand/wallet/send.rs
```diff
@@ -95,8 +100,15 @@ impl Send {
       change,
       self.fee_rate,
       postage,
-    )
-    .build_transaction()?;
+    );
+
+    if let Some(op_return) = self.op_return {
+      // TODO: improve the check for checking byte-length
+      assert!(op_return.len() <= 80);
```
raphjaph, line 107:
```suggestion
      assert!(op_return.len() <= 82);
```

src/wallet/transaction_builder.rs
```diff
@@ -36,6 +36,8 @@ use {
   std::cmp::{max, min},
 };
 
+use bitcoin::blockdata::script::PushBytesBuf;
+
```
raphjaph, line 40: Inline with the other imports

raphjaph:
I'm going to close this but please reopen if you wish to continue work on this!

PR #3069, opened 2024-01-31
Add balances data into rune info

lugondev:
LugonShot 2024-01-31 at 14 11 30@2x

- Add balances data into rune info


datnddot:
Nice move 👍

0xTigerbao:
nice nice nice.... 🧨

lugondev:
LugonShot 2024-01-31 at 22 04 35@2x

👯

lugondev:
@raphjaph @casey help me review this PR

thanks

lugondev:
I hope this pull request will appear in the next release.

raphjaph:
The call to `rune_balances` is very inefficient. We would need to add a multimap table called RUNE_TO_OUTPOINTS. This gets into the guts of the database but you can take a stab at it if you want :)

lugondev:
I think users really want to see specific rune's balance. 
Hope you can review my Pr again

casey:
We cannot merge this as-is. The code as written loads every output with a rune balance from the database. With any nontrivial amount of runes, this will take an extreme amount of compute and RAM, and would be a DoS vector.

To make this efficient enough to be safe, we would need to add a rune -> outpoint balance map which is updated whenever an outpoint is spent or a new outpoint is created.

lugondev:
oh, got it.

svensei:
!wen // really excited about this feature tbh

PR #3070, opened 2024-01-31
Ensure a RUNE balance is set to 0 if it is etched with an expired deadline

victorkirov:
I'm still trying to wrap my head around the Runes code, so not 100% sure, but I _think_ this check is missing 👀

### DIFF
--- a/src/index/updater/rune_updater.rs
+++ b/src/index/updater/rune_updater.rs
@@ -310,7 +310,9 @@ impl<'a, 'db, 'tx> RuneUpdater<'a, 'db, 'tx> {
             rune,
             spacers,
             supply: if let Some(limit) = limit {
-              if end == Some(self.height) {
+              if end == Some(self.height)
+                || (deadline.is_some() && self.timestamp >= deadline.unwrap())
+              {
                 0
               } else {
                 limit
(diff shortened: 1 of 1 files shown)


casey:
Thank's for the PR! This check should indeed be here as is, but this code is going through a lot of flux, and I'm actually going to remove the restriction that etchers must obey the mint limits. When they do this, I'm also going to display a premine percentage on the rune page, so it's clear when the etcher has premined the token.

victorkirov:
Ok, great. Thanks for the feedback 😄 Looking forward to seeing how this evolves

casey:
Definitely very happy that someone is paying attention! Critical review of the low-level details is super welcome.

victorkirov:
I see the new mint changes went in. I think this issue is still present though 👀

casey:
Yah, definitely. The mint changes were  mostly a refactor to require a `mint` flag to be set in order to have an open mint. I'm going to remove all the limits in a follow-up PR. The follow-up PR will actually allow people to create runes that can overflow a u128, and just use saturating addition everywhere. So it's caveat emptor and on the rune creators to pick limits, either to the limit per mint, or the number of mints that are possible, that don't wind up overflowing.

PR #3074, opened 2024-02-02
fix codestyle in test

zhiqiangxu:
All other tests are using `..Default::default()` for default values, better to stick to it..


raphjaph:
I think for now we want to be explicit in the tests. Might add proper defaults later

Issue #3065, opened 2024-01-29
Add `/r/blocktime/<HEIGHT>`


raphjaph:
Or add a `/r/blockinfo/<HEIGHT>

cliobitcoinbank:
We need this to use as a seed in generative stuff and other block based features.

stet:
what will this output contain? entire block data or limited?

jerryfane:
> what will this output contain? entire block data or limited?

I made a pull request for this #3075 

The output as of now contains:

-  Block Hash
-  Block Height
- Timestamp
- Number of Transactions
- Count of Confirmations
- Previous Block Hash
- Next Block Hash
- Inscriptions

Open for feedback though.

stet:
> Open for feedback though.

Thank you. Remind me, what is the total block data fields that Ord indexes?
I will review code but if you know off-hand, thanks.
Ideally, this returns all indexed data. 
Typical block endpoint responses also include:

- ver
- mrkl_root
- bits
- fee
- nonce
- size        
- weight

PR #3075, opened 2024-02-02
Implement /r/blockinfo

jerryfane:
Fix issue #3065


elocremarc:
I think the `inscriptions` will probably be need to be discused. If we were to add it it would probaly need pagination like with the children and sat endpoint. Each block could have hundreds or thousands of inscriptions. 

Also what happens if there is not a `next_block_hash` because we are at the current height. Or what happens if there is a reorg and an inscription changes.

jerryfane:
Yes I agree for the inscriptions field. 

If there is no Next block, then next_block is null.
Ord already handles re-orgs.

raphjaph (approved):
LGTM!

We removed the inscriptions field for now, we can always add it later with proper pagination

casey:
Nice! Added more info, docs, and tests.

PR #3081, opened 2024-02-03
replace clamp with min

zhiqiangxu:
Since `u128>=0` is always satisfied.


raphjaph:
I like clamp but min is more succinct. @casey

casey:
LGTM!

PR #3083, opened 2024-02-04
Update install.sh

SleepTheGod:
Enhancements and new features in this advanced version:

Custom Repository Support: This script allows specifying a custom GitHub repository with the --repo option. This is helpful for installing binaries from forks or alternative versions hosted on different repositories.

JSON Parsing for Tag Name: Instead of using grep and cut, this script uses jq, a powerful JSON processor, to parse the tag name from GitHub's API response. This approach is more reliable and readable.

Improved Platform Detection: The platform detection now explicitly supports arm64 and x86_64 on both Darwin (macOS) and Linux. It also provides a clear error message for unsupported platforms.

Error Handling and Cleanup: The script includes better error handling and uses a trap to ensure temporary files are cleaned up even if the script exits prematurely.

Verbose Installation Message: Upon successful installation, the script outputs a message indicating the binary's installation location.

Remember to install jq before running this script, as it's used for JSON parsing. This version aims to be more adaptable, robust, and user-friendly for various use cases.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,52 +2,52 @@
 
 set -euo pipefail
 
-if [ ! -z ${GITHUB_ACTIONS-} ]; then
-  set -x
-fi
+# Enable tracing in GitHub Actions environment
+[ -n "${GITHUB_ACTIONS-}" ] && set -x
 
 help() {
-  cat <<'EOF'
-Install a binary release of ord hosted on GitHub
+  cat <<EOF
+Advanced install script for ord binary releases from GitHub
 
 USAGE:
-    install.sh [options]
+    install.sh [OPTIONS]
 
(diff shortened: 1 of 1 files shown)


casey:
The install.sh script has to be manually tested, so I think we want to do these as individual smaller PRs, instead of one big one, since they'll be easier to review.

Comments about the individual features:

- Custom Repository Support: This seems good.
- JSON Parsing for Tag Name: We shouldn't use `jq`, since it's an additional dependency, and grep and cut are always installed.
- Improved Platform Detection: This sounds good.
- Error Handling and Cleanup: This sounds good.
- Verbose Installation Message: This sounds good.

PR #3086, opened 2024-02-05
Optimize preview text

oxSaturn:
Currently it's difficult to tell what's the text:

CleanShot-PhZxluE4@2x

Here's how it looks after the update:

CleanShot-wryKGaBy@2x

Let me know what you think.


casey:
I think we don't want to do these kinds of content-based heuristics. People should be using `application/json` for JSON inscriptions.

PR #3088, opened 2024-02-07
chore: remove index parameter from index_block

zhiqiangxu:
The `index` object is already contained in `self`, so no need to pass it to an instance method, this also avoids referencing to the same object in two ways(`index` and `self.index`).


casey:
LGTM!

PR #3112, opened 2024-02-13
chore: fix naming

zhiqiangxu:
`range_cache` contains mapping of `output` to `sat ranges`, this also aligns the naming with other places.


raphjaph (approved):
lgtm

PR #3114, opened 2024-02-13
chore: modify description for inscription sat

zhiqiangxu:
The inscription is not necessarily on the first sat if there's a pointer tag specified.


docs/src/inscriptions.md
```diff
@@ -60,9 +60,9 @@ few restrictions is that individual data pushes may not be larger than 520
 bytes.
 
 The inscription content is contained within the input of a reveal transaction,
-and the inscription is made on the first sat of its input. This sat can then be
-tracked using the familiar rules of ordinal theory, allowing it to be
-transferred, bought, sold, lost to fees, and recovered.
+and the inscription is made on the first sat of its input if there's no pointer
+tag specified. This sat can then be tracked using the familiar rules of ordinal
```
raphjaph, line 64:
```suggestion
and the inscription is made on the first sat of its input if it has no pointer
field. This sat can then be tracked using the familiar rules of ordinal 
```

raphjaph (changes requested):
Looks good!

Minor nit

docs/src/inscriptions.md
```diff
@@ -116,13 +116,13 @@ Inscriptions can either be located in different inputs, within the same input or
 a combination of both. In any case the ordering is clear, since a parser would
 go through the inputs consecutively and look for all inscription `envelopes`.
 
-| Input | Inscription Count | Indices    |
-|:-----:|:-----------------:|:----------:|
-| 0     | 2                 | i0, i1     |
-| 1     | 1                 | i2         |
-| 2     | 3                 | i3, i4, i5 |
-| 3     | 0                 |            |
-| 4     | 1                 | i6         |
```
raphjaph, line 125: I think we just want to stick to the old versoin.

JohnKraken (approved):
Better this way...

raphjaph (approved):

PR #2811, opened 2023-12-04
Emit location update events to optional channel

mi-yu:
Addresses #2789 

Introduces a way for `InscriptionUpdater` to emit location update events on an in-process channel. Example usage can be found here: https://github.com/mi-yu/ord-custom-indexer-example

cc @casey

PR #3137, opened 2024-02-17
Emit location update events to optional channel

mi-yu:
Addresses https://github.com/ordinals/ord/issues/2789

Introduces a way for InscriptionUpdater to emit location update events on an in-process channel. Example usage can be found here: https://github.com/mi-yu/ord-custom-indexer-example

Copy of #2811 to allow edits by maintainers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,4 @@
 /target
 /test-times.txt
 /tmp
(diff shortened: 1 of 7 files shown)


casey:
Thanks, merged!

PR #3144, opened 2024-02-19
Use `image-rendering: auto` when downscaling images

casey:
I think this is an improvement for all image formats and image types. `image-rendering: pixelated` is only important when upscaling pixel art. When upscaling other kinds of images, and when downscaling everything, `image-rendering: auto` is preferred.

This makes the image preview use `auto` rendering, when the body is less than the size of the image, A.K.A., downscaling.

This still leaves things suboptimal when upscaling non pixel art, but I can't think of a non-disruptive change to fix that.

### DIFF
--- a/static/preview-image.js
+++ b/static/preview-image.js
@@ -1,0 +1,15 @@
+function resize() {
+  if (
+    body.clientWidth * window.devicePixelRatio < img.naturalWidth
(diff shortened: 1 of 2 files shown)


casey:
Pixel art native resolution:

pixel art - native


pixel art - downscale auto
pixel art - downscale pixelated

raphjaph (approved):

PR #3151, opened 2024-02-21
top level /inscription_query route

elocremarc:
Closes https://github.com/ordinals/ord/issues/2807

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -260,6 +260,7 @@ impl Server {
           "/inscriptions/block/:height/:page",
           get(Self::inscriptions_in_block_paginated),
         )
+        .route("/:inscription_query", get(Self::inscription))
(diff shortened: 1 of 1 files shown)


raphjaph (changes requested):
Nice, easy peasy!

Could you add a test for this route, just to check that it matches the `/inscription/inscription_query` ?

elocremarc:
@raphjaph I didn't realize this issue https://github.com/ordinals/ord/issues/3099 when I made this PR. Do we want to do more than just the inscription on the home route?

It would be pretty easy if we redirected like /search is doing. Or do we want to keep people on the route for issue 3099? Like if I go to /:tx I wouldn't get rerouted to /tx/:tx?

raphjaph:
Going to close this for now, but we should revisit the generalized search idea

PR #3163, opened 2024-02-23
Preview parent, iframe on inscription page like children.

elocremarc:
Adds a preview for the parent like the children have.


raphjaph (approved):
LGTM!

PR #3181, opened 2024-02-28
Prevent error loading libssl in docker image

aekasitt:
```
ord: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
```
is the error you will get if you don't install `openssl` or `libssl-dev` via apt-get in the final image.

Similar to what the `nodejs` community found out here in the middle of June, 2023.
[node:lts-slim does not include OpenSSL as it is now based on bookworm because of upstream change in debian:bookworm-slim image](https://github.com/nodejs/docker-node/issues/1919)

As stated by example above:

> ...
> The official docker image for [Debian](https://hub.docker.com/_/debian/) is where the change comes from
>
>>   in [bullseye-slim](https://github.com/docker-library/repo-info/blob/master/repos/debian/local/bullseye-slim.md) OpenSSL 1.1.1n is mentioned.
>>   in [bookworm-slim](https://github.com/docker-library/repo-info/blob/master/repos/debian/local/bookworm-slim.md) OpenSSL is not mentioned.
>
> ...

But I tried using [bullseye-slim](https://github.com/docker-library/repo-info/blob/master/repos/debian/local/bullseye-slim.md) in place of [booksworm-slim](https://github.com/docker-library/repo-info/blob/master/repos/debian/local/bookworm-slim.md) but ran into the same error, so I think installing either `openssl` or `libssl-dev` on the final image would suffice.

Reasoning for installing `openssl` in pull request instead of `libssl-dev` is the size difference in built image. Installing either one allows me to run **Ord Server** without a hitch and the final image size with *openssl* installed is 155Mb while the final image size with *libssl-dev* installed is 166Mb, ergo less is more.

---
Findings from my one-command deployment of local Bitcoin & Lightning cluster, [Aesir](https://github.com/krutt/aesir) , now supporting **Ord Server** as a peripheral service. Once I figured out the original Dockerfile was faulty, I thought I would make a contribution. Hope this was useful.


raphjaph (approved):
LGTM

casey (approved):

PR #3196, opened 2024-02-28
Carbonado support

cryptoquick:
Hi, if you're not familiar with [Carbonado](https://carbonado.io), it's a decentralized storage approach we developed for [DIBA](https://diba.io). It was inspired by Nostr relays, by how data is published to multiple nodes in order to ensure data availabiilty. Carbonado adds stream verification, which allows nodes to maintain a certain replication factor through spot-check challenge-response queries without needing to download the entire file over again to verify its integrity.

The idea behind integrating it into ord is to allow for nodes that support a `/carbonado` endpoint to query a preset Carbonado node, or proxy that queries multiple nodes at once. For nodes that don't support the `/carbonado` endpoint, that's fine, the following HTML can be used to fall back to an inscription of lower resolution / fidelity / size from the `/content` endpoint.

```html
<object data="/carbonado/-npub159lm4yl53kdqnug0l2ha6d24csljnauhlutc0te7wewrumrqt3fsuvjkvf/27418bc853c672d246737a0d7d71b5265175db9903ded4c970e3785e83b514b6" style="height: 100%; width: 100%;">
    <img src="/content/15fa71c2959b96d3f7c120af3e3e28df329a73da57229c76d8473a9c5b406d7di0">
</object>
```

An example of this can be found here:

https://testnet.ordinals.com/inscription/be4417fe48d9e4a58db67d4fec6d146b0012668d2f8f9cb2b273dabcdeb51effi0

Additionally, I can add support for calling multiple endpoints for redundancy, and I can also add in a feature flag that disables this functionality by default if that's what's preferred by the community.

Feel free to try this with our Carbonado node:
`CARBONADO_ENDPOINT='https://hut8-ca-on-mississauga.carbonado.io/carb/retrieve'`


cryptoquick:
@raphjaph offered the following feedback:

- Failing test is due to CSP assertion
- Need to validate inputs better
- Should be feature-flagged

I think I'll use Carbonado to verify the data from the endpoint matches the hash. This is an improvement in safety and better guarantees immutability.

cryptoquick:
I just realized, this approach only works on a single node with Inscriptions... I'll need to implement a new Blake3-hashed endpoint in Carbonado to support content-addressability.

cryptoquick:
Content-addressable Carbonado tested with testnet inscription `fd676f18f9e150dc3d9e7f3095c828d82c6125be846dbb6bfda70b44cd1358eai0`

cryptoquick:
Okay, this is fully tested and ready for review, @raphjaph @casey

casey:
I think the issue that I would be concerned about is that although cabonado data is referenced by hash, it is still not on-chain, and so could disappear in the future. Currently, all inscriptions are entirely on-chain, as well as all recursive endpoints that an inscription can access. So I think unfortunately we couldn't support this, because it could lead to inscriptions ultimately breaking when content goes offline, similar to what happens with NFTs which reference data on IPFS.

cryptoquick:
@casey Two things I have to say in defense of Carbonado; 1) in the future, we're working on storage debt markets that incentivize keeping off-chain data and other ways to effect data permanence such as orbital archival, and also, 2) we're going to operate a forked carbonadord instance for DIBA and the various collections that want to drop on Carbonado anyway, so it'd be nice if we didn't have to maintain a fork of ord in order to do so.

I ask that you revisit decision, or at least keep an open discussion, especially since if there are fallback inscriptions like in the example and HTML snippet provided.

Issue #3133, opened 2024-02-15
Fix flaky tests


casey:
Our tests are very flaky at the moment, which is super annoying while developing, and require re-running CI frequently.

mj10021:
Is there a list anywhere of which tests are flaky?

casey:
I think there are a bunch, although there might be a single underlying cause. A good way to find them is to run:

```sh
while cargo test; do echo running; done
```

Which will run the tests repeatedly until one fails. I just ran this, and after a few runs, `etch::send_amount_does_not_select_runic_utxos` failed.

casey:
However, I ran `etch::send_amount_does_not_select_runic_utxos` a bunch on its own, and it never failed, so I think it might be the interaction of some tests, or running a bunch of tests at the same time which is the issue.

raphjaph:
They ain't flaky anymore. KNOCK, KNOCK, KNOCK

PR #3198, opened 2024-02-29
fix: flaky tests error

bingryan:
## Description

for `cargo test `, some error case:

`case 1`: 
```
called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
``` 
> solution: use `ulimit` command to add resource , such as: `ulimit -n 10240`

`case 2`:  spawn too many server

try to add exponential backoff mechanism(sleep time) to retry,  Not sure it can fix the CI error. 
 
## Related Issue

Related #3133

## Proposed Changelog

- Added exponential backoff mechanism(sleep time) to retry for test

### DIFF
--- a/crates/test-bitcoincore-rpc/src/lib.rs
+++ b/crates/test-bitcoincore-rpc/src/lib.rs
@@ -97,17 +97,21 @@ impl Builder {
 
     thread::spawn(|| rpc_server.wait());
 
-    for i in 0.. {
+    let mut attempts = 0;
+    let max_attempts = 600;
+    let mut backoff = 25;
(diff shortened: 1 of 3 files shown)


bingryan:
@casey code review plz :)

casey:
Thanks for the PR! I actually want to solve this in a different way. I'm going to add an `/update` endpoint which updates the index, which is only available during tests. So the client can make a request to `/update` and know that the index will have been updated, without having to poll the index. This should be more reliable, and avoid a loop entirely.

PR #3203, opened 2024-03-01
Support creating and restoring wallets with accounts

### LINKED ISSUE #3204: Support creating and restoring wallets with accounts

casey:
To do:
- Error if restoring from a descriptor and account is set, since descriptors already contain the account
- Add a test that descriptors created when creating and restoring a wallet with an account contain the account
- Add some test cases that simply check that descriptors generated with the same seeds result in the same descriptor


casey:
This is a good PR but I'm never going to finish it 😭

PR #3206, opened 2024-03-02
Encode claims with tag

casey:
This PR changes claims from being a property of edict IDs, to being done with a dedicated tag.

This has a few advantages:
- Only one rune can be claimed at a time with a given runestone. Claim limits make claims costly because the claimant must pay transaction fees. If a given runestone can claim more than one rune, the the transaction fee is amortized across multiple claims. Making a claim a tag prevents this.
- The code for claiming with a claim bit is weird.
- If you want to claim a rune for the maximum amount, and send it to the default output, this can be done with a claim tag without needing an edict.
- The code is actually much cleaner.

I also finally refactored `RuneUpdater::index_runes` into multiple functions.

### DIFF
--- a/src/index/updater/rune_updater.rs
+++ b/src/index/updater/rune_updater.rs
@@ -1,13 +1,14 @@
 use {
   super::*,
-  crate::runes::{varint, Edict, Runestone, CLAIM_BIT},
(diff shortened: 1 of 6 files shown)


raphjaph (approved):
lgtm

PR #3209, opened 2024-03-02
Add documentation to clarify that inscriptions are served from /content/<INSCRIPTION_ID>

### LINKED ISSUE #2903: Make clear inscriptions are served from /content/:id
This topic keeps coming up again and again on twitter and in issues. People seem confused that inscriptions need to be served from a web server and specifically from the content route with the id as a dynamic route. I think this needs to be better defined in the handbook. People keep asking for a recursive route to return the Id when I don't think enough people realize that inscriptions are defined with a server architecture. 

Under inscriptions in the handbook it says. 
> The inscription content model is that of the web. An inscription consists of a content type, also known as a MIME type, and the content itself, which is a byte string. This allows inscription content to be returned from a web server, and for creating HTML inscriptions that use and remix the content of other inscriptions.

We might need to be more clear about the routing here because this has become a standard way of getting the Id and mimetype since we introduced the child and metadata endpoint. Also since we introduced recursion /content/ has been solidified as well.

### LINKED ISSUE #3208: Proper consensus on how inscriptions should self reference
I know this has been discussed many times before but I feel like this is still not properly resolved, so raising another issue.

I'm aware this exact issue was raised in #3015, and that #2628 closed that issue, but I disagree that this issue is resolved. 

At the moment if an inscription wants to fetch details about itself, such as metadata, children, sat number, etc, it first needs to know it's own inscription ID. Raph and Casey have both suggested in the past that you can get this from the url with something like this:
```js
let id = window.location.pathname.split("/").pop();
```
However, this is not documented anywhere in the handbook, so, as I see it, there is no guarantee that the inscription ID is in the url, or that it is the correct ID. I know pretty much every ordinals explorer does have the ID in the url, but without this being a requirement as part of the part of the protocol, there is no guarantee. 

This is actually already a problem, as ord.io behaves differently to ordinals.com in this regard. If you inscribe a HTML inscription, then inscribe a delegate of that HTML, ord.io will render the delegate in an iframe with the inscription ID of the original inscription, not the delegate, making it impossible for an inscription to retrieve information about itself. Here is an example inscription that demonstrates the problem:
733ca7d69f286ab295f09952f0c4d55daeae4b2fd8c1e3683fd377bbef215282i100

You can see in this screenshot that the ID in the iframe is different to the actual inscription ID
309445850-f0cf4af9-bf08-420a-965c-685dd2b2c3f0

While this behaviour from ord.io is frustrating, I don't consider it a bug, because the ordinals protocol does not specify that the url of the iframe must contain the inscription ID of the current inscription. I am just showing ord.io's behaviour to demonstrate that this is a serious issue.

I can see several possible solutions to this:
1. Define the structure of the url used to display HTML inscriptions in the ordinals handbook, making it part of the ordinals protocol. By doing this, ord.io would no longer be following the protocol and would be required to fix this
2. Add a new inscription endpoint (e.g. `/r/self`) that returns the same content as the `/r/inscription/:id` endpoint, but without ever needing to know it's own id (as suggested in #3015)
3. For any recursive endpoint that takes an inscription id in the path (`/r/children/:id`, `/r/inscription/:id`, `/r/metadata/:id`), you can use the keyword `self` instead of an actual inscription ID and it will return the corresponding payload for the current inscription. (e.g. `/r/metadata/self`, `/r/children/self`, etc)

In my opinion, option 3 is the best one, as it means you can get the children, metadata, etc, in a single fetch, rather than having to fetch the id from `/r/self`, then fetching the metadata/children as you would have to do in option 2. I think option 1 is the least desirable, as I think the ordinals protocol should concern itself as much as possible with the inscription/ordinals side of things, and not the explorers. I feel like it is preferable to allow explorers to choose how they want to design their systems, and if that means not using the standard conventions regarding the url, it shouldn't break a bunch of inscriptions.

I'm happy to raise a PR for this, but also happy for someone else to, because I've got a lot of other shit to do. But I wanted to get others opinion and consensus on the best solution before raising a PR

Vanniix:
Closes #2903
Closes #3208

Adds a few lines to the documentation to make it clear that inscriptions are served from /content/<INSCRIPTION_ID>. Many inscriptions rely on this fact to get their own inscription ID, so it is important that this is properly documented so we know we can rely on this.

Also added clarification regarding the url of inscriptions that have delegates. Happy to remove this if people think it is implicit in the other sentences, but since ord.io is currently making this mistake, I thought it is better to clarify.

Also added an example for how inscriptions can self reference and fetch their own inscription ID. Also happy to remove this, as it isn't directly about the protocol, but rather how to use it. But a lot of people ask how to do this, so having it in the handbook would be helpful.

### DIFF
--- a/docs/src/inscriptions.md
+++ b/docs/src/inscriptions.md
@@ -70,6 +70,9 @@ Content
 The data model of inscriptions is that of a HTTP response, allowing inscription
 content to be served by a web server and viewed in a web browser.
 
+Inscriptions are served from the URL `/content/<INSCRIPTION_ID>`, where 
+`<INSCRIPTION_ID>` is the ID of the inscription to fetch.
+
 Fields
 ------
 
@@ -148,7 +151,12 @@ off-chain content, thus keeping inscriptions immutable and self-contained.
 
 This is accomplished by loading HTML and SVG inscriptions inside `iframes` with
 the `sandbox` attribute, as well as serving inscription content with
-`Content-Security-Policy` headers.
+`Content-Security-Policy` headers. 
+
+The iframe has a URL of `/content/<INSCRIPTION_ID>`, where `<INSCRIPTION_ID>` is 
+the ID of the rendered inscription. If an inscription has a delegate, 
+`<INSCRIPTION_ID>` is the ID of the inscription being rendered, *not* the ID of 
+the delegate.
 
 Reinscriptions
 --------------
--- a/docs/src/inscriptions.md
+++ b/docs/src/inscriptions.md
@@ -61,7 +61,7 @@ bytes.
 
 The inscription content is contained within the input of a reveal transaction,
 and the inscription is made on the first sat of its input if it has no pointer
-field. This sat can then be tracked using the familiar rules of ordinal 
+field. This sat can then be tracked using the familiar rules of ordinal
 theory, allowing it to be transferred, bought, sold, lost to fees, and recovered.
 
 Content
@@ -69,9 +69,6 @@ Content
 
 The data model of inscriptions is that of a HTTP response, allowing inscription
 content to be served by a web server and viewed in a web browser.
-
-Inscriptions are served from the URL `/content/<INSCRIPTION_ID>`, where 
-`<INSCRIPTION_ID>` is the ID of the inscription to fetch.
 
 Fields
 ------
@@ -151,12 +148,27 @@ off-chain content, thus keeping inscriptions immutable and self-contained.
 
 This is accomplished by loading HTML and SVG inscriptions inside `iframes` with
 the `sandbox` attribute, as well as serving inscription content with
-`Content-Security-Policy` headers. 
+`Content-Security-Policy` headers.
 
-The iframe has a URL of `/content/<INSCRIPTION_ID>`, where `<INSCRIPTION_ID>` is 
-the ID of the rendered inscription. If an inscription has a delegate, 
-`<INSCRIPTION_ID>` is the ID of the inscription being rendered, *not* the ID of 
-the delegate.
+Self-Reference
+--------------
+
+The content of the inscription with ID `INSCRIPTION_ID` must served from the
+URL path `/content/<INSCRIPTION_ID>`.
+
+This allows inscriptions to retrieve their own inscription ID with:
+
+```js
+let inscription_id = window.location.pathname.split("/").pop();
+```
+
+If an inscription with ID X delegates to an inscription with ID Y, that is to
+say, if inscription X contains a delegate field with value Y, the content of
+inscription X must be served from the URL path `/content/X`, *not*
+`/content/Y`.
+
+This allows delegating inscriptions to use their own inscription ID as a seed
+for generative delegate content.
 
 Reinscriptions
 --------------
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -62,9 +62,9 @@ plain-text responses.
 ### Self Referencing
 
 An inscription may fetch data about itself using the endpoints above by 
-using its own inscription ID. Since all inscriptions are rendered in an 
-iframe with a URL of `/content/<INSCRIPTION_ID>`, it can retrieve its own 
-inscription ID from `window.location.pathname` using the following line (or similar)
+using its own inscription ID. Since all inscriptions are rendered with 
+a URL of `/content/<INSCRIPTION_ID>`, it can retrieve its own inscription 
+ID from `window.location.pathname` using the following line (or similar)
 
 ```js
 const inscriptionId = window.location.pathname.match(/\/content\/([\da-f]{64}i\d+)/i)[1]
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -59,17 +59,6 @@ plain-text responses.
 - `/blockhash/<HEIGHT>`: block hash at given block height.
 - `/blocktime`: UNIX time stamp of latest block.
 
-### Self Referencing
-
-An inscription may fetch data about itself using the endpoints above by 
-using its own inscription ID. Since all inscriptions are rendered with 
-a URL of `/content/<INSCRIPTION_ID>`, it can retrieve its own inscription 
-ID from `window.location.pathname` using the following line (or similar)
-
-```js
-const inscriptionId = window.location.pathname.match(/\/content\/([\da-f]{64}i\d+)/i)[1]
-```
-
 Examples
 --------
 
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -59,6 +59,17 @@ plain-text responses.
 - `/blockhash/<HEIGHT>`: block hash at given block height.
 - `/blocktime`: UNIX time stamp of latest block.
 
+### Self Referencing
+
+An inscription may fetch data about itself using the endpoints above by 
+using it's own inscription ID. Since all inscriptions are rendered in an 
+iframe with a URL of `/content/<INSCRIPTION_ID>`, it can retrieve its own 
+inscription ID from `window.location.pathname` using the following line (or similar)
+
+```js
+const inscriptionId = window.location.pathname.match(/\/content\/([\da-f]{64}i\d+)/i)[1]
+```
+
 Examples
 --------
 
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -62,7 +62,7 @@ plain-text responses.
 ### Self Referencing
 
 An inscription may fetch data about itself using the endpoints above by 
-using it's own inscription ID. Since all inscriptions are rendered in an 
+using its own inscription ID. Since all inscriptions are rendered in an 
 iframe with a URL of `/content/<INSCRIPTION_ID>`, it can retrieve its own 
 inscription ID from `window.location.pathname` using the following line (or similar)
 
--- a/docs/src/inscriptions/delegate.md
+++ b/docs/src/inscriptions/delegate.md
@@ -2,8 +2,9 @@ Delegate
 ========
 
 Inscriptions may nominate a delegate inscription. Requests for the content of
-an inscription with a delegate will instead return the content and content type
-of the delegate. This can be used to cheaply create copies of an inscription.
+an inscription with a delegate will instead return the content, content type 
+and content encoding of the delegate. This can be used to cheaply create copies 
+of an inscription.
 
 ### Specification
 


devords:
Looks great!

gmart7t2:
It would also be helpful if the handbook documented that the content-type and content-encoding are delegated along with the body when delegating. 

Some sites don't implement this. If you view a delegate inscription on MagicEden it won't display correctly unless the delegate inscription has its own copy of the content-type tag, which is obviously wasteful.

leonidasord:
It is too opinionated to make it part of the protocol that when you build a client an inscription needs to "retrieve its own inscription ID from `window.location.pathname`". Not a fan of this PR. People should be able to build clients however they want as long as they implement the recursive endpoints. If people want to self reference then add a recursive endpoint. We should not start mandating that clients implement very specific arbitrary client side code to match ordinals dot com.

Vanniix:
Thanks for the feedback @leonidasord. I actually agree with you. I made similar comments towards the end of the issue I raised about this: #3208. But after talking with a few people, I feel like the pathname is the only option for self referencing. 

A self endpoint is a great idea in principle, but I don't see how it would work in practice. If the server receives a request to `/r/self`, how on earth does it know which inscription is requesting it, and how does it know what payload to send back. None of the endpoints so far are context dependent. They all have every piece of information they need in the path, but an `/r/self` would need some extra context (the id of the requesting inscription) and I can't see how it would get this. If anyone has any ideas about how to achieve this, I would love to know, because I prefer the idea of a self endpoint too.

I know that dictating the structure of the url is not ideal, but I think it's important we work out some way to self reference. Without it, you can't make dynamic/recursive collections with delegates (unless every item is identical), and for certain projects, this would be a significant cost saving.

Would love to hear other proposals for how inscriptions can self reference.

devords:
@leonidasord Problem is, server would have no way of figuring out what `self` is for the recursive endpoint you're proposing. Only workaround is using HTTP Referral field, which would defeat the purpose.

leonidasord:
What I am imagining is a /self recursive endpoint where all it does is return the inscription ID of the inscription that is calling it. This can happen completely client side. I haven't thought through the edge cases very hard though like if you recursively bring in an inscription and that inscription uses /self the client would need to know to give it its id still.

What are the best reasons for why this wouldn't work?

The alternative of calling window like this and relying on clients implementing several things a specific way for that to work feels overly constraining to devs. If for some reason it doesn't end up making sense to do the /self endpoint client side idea I would like to see an improved version of this PR that mandates that all clients attach an object with a specific set of key value pairs to the window that includes the inscription id so that there is a very clear spec for for developers to implement. The current form of this PR is yucky to me because it's basically telling devs they have to structure their client a very specific way in order for people to do this hacky window call. If we are going to do the window call just make devs be required to provide a specific set of data in an object and that is all.

devords:
> /self recursive endpoint where all it does is return the inscription ID of the inscription that is calling it.

As I said, the caller of `/self` would have to pass the inscription ID along with the request in order for the ord server to know what `self` is.

Vanniix:
Yeah, I'm not sure having the `/self` endpoint completely client side is possible. The client can't intercept certain routes and handle them client side (that I know of), all requests will automatically go to the server, and the server won't be able to handle self requests.

I do like the idea of injecting some global object into the window/iframe though. Depending on how it is designed, it may be possible that the inscription code runs before the global object is set, so you may also need to fire an event to indicate it's been set, and the inscription code can listen for the event. Similar to the design pattern used for the `DOMContentLoaded` event. i.e. You might do the following in an inscription (if the global var is called `inscription`)
```js
if (window.inscription === undefined) {
  document.addEventListener("InscriptionContentLoaded", () => console.log(window.inscription))
} else {
  console.log(window.inscription)
}
```

The variable would need to be called something non-obvious though. If we chose a name that an existing inscription is already using as a global variable, then that inscription would break, because we just overwrote it with something it wasn't expecting. Maybe add a couple of underscores or something like that (i.e. `__inscription`)

devords:
@Vanniix Seems very convoluted for explorers to inject the state like that into 3rd party content, just in order to avoid relying on specific URL paths.

"Recursion" is essentially a CSP directive, which is a very web/HTTP specific thing. I am not sure why protocol cannot take a position on how content/preview URLs must look like, considering all recursive endpoints are tied to specific URLs + the fact that using `window.location.pathname` has been officially suggested by the protocol team numerous times.

elocremarc:
> inscription ID from window.location.pathname". Not a fan of this PR. People should be able to build clients however they want as long as they implement the recursive endpoints. If people want to self reference then add a recursive endpoint. We should not start mandating that clients implement very specific arbitrary client side code to match ordinals dot com.

I agree this PR is to specific should not enforce any examples of front end implementation. The intent of the issue that raised this PR was that its clear from the rules of recursion and prior that an inscription is returned from a server using the `/content/:id`.

From the handbook this just needs updating to be more explicit that it needs to be returned from `/content/` that is all:

> The inscription content model is that of the web. An inscription consists of a content type, also known as a MIME type, and the content itself, which is a byte string. This allows inscription content to be returned from a web server, and for creating HTML inscriptions that use and remix the content of other inscriptions.

docs/src/inscriptions.md
```diff
@@ -148,7 +151,12 @@ off-chain content, thus keeping inscriptions immutable and self-contained.
 
 This is accomplished by loading HTML and SVG inscriptions inside `iframes` with
 the `sandbox` attribute, as well as serving inscription content with
-`Content-Security-Policy` headers.
+`Content-Security-Policy` headers. 
+
+The iframe has a URL of `/content/<INSCRIPTION_ID>`, where `<INSCRIPTION_ID>` is 
+the ID of the rendered inscription. If an inscription has a delegate, 
+`<INSCRIPTION_ID>` is the ID of the inscription being rendered, *not* the ID of 
+the delegate.
```
elocremarc, line 159: This is actually wrong for the ord explorer the /inscription page the iframe src is under /preview/ for some mime-types like html. The html preview template uses the /content/
However all this is redundant because it was stated above I would remove these changes.

docs/src/inscriptions/recursion.md
```diff
@@ -48,6 +48,17 @@ plain-text responses.
...
+
+An inscription may fetch data about itself using the endpoints above by 
+using its own inscription ID. Since all inscriptions are rendered with 
+a URL of `/content/<INSCRIPTION_ID>`, it can retrieve its own inscription 
+ID from `window.location.pathname` using the following line (or similar)
+
+```js
+const inscriptionId = window.location.pathname.match(/\/content\/([\da-f]{64}i\d+)/i)[1]
+```
+
```
elocremarc, line 61: I don't think we need any of this. We shouldn't enforce how people use the window object or even give them regex none of the docs have any examples of inscription implementation.

Vanniix: Yeah, I'm fine with removing this as it's not strictly necessary. However, I don't think it's bad to include examples in documentation. Heaps of software documentation includes examples, they're designed to make developers lives easier, so I don't see the issue with it. This was meant to just be an example of how you could do it, not enforcing people to do it this way, as it is a common question that new ordinals developers have. 

Maybe it belongs somewhere in the guides section instead, but like I said, this isn't necessary, so I have removed it for now

docs/src/inscriptions.md
Vanniix: Oops, forgot about it having the preview url

I know you want this removed, but I think this part is necessary. You want it to just be that the inscription is served from `/content/:id`, and thats all we say on the matter, but just because the inscription is served from that url, doesn't mean you can use the path name to get the id (which is the whole point of this PR). Your point about `/preview/` partially demonstrates this problem. ordinals.com serves its inscriptions from `/content/:id`, but it also serves them from `/preview/:id`, which is used for the iframes in the front end. They are different urls, implying the explorer is free to choose the iframe url however they choose (provided it serves the right inscription). So, if we don't say anything about the iframes in the handbook, who's to say some explorer doesn't decide to use some other weird format for the iframe url?

The question then is what do we enforce? Do we just say it must end in `/:id`, and leave the start of the path up to the implementation?

Vanniix:
> It would also be helpful if the handbook documented that the content-type and content-encoding are delegated along with the body when delegating.
@gmart7t2, yep, good point. `delegate.md` does actually mention the content type, but doesn't mention the content encoding, so I added that in there too

leonidasord:
The more I think about this the less I like my ideas of a client-side /self endpoint or appending an object to window. Yall are making good points. I also agree that it would be very helpful for inscriptions to be able to know their ID. All of that said it still feels yucky to me to make it part of the Ordinals protocol that for the next 100 years clients have to support allowing people to call `window.location.pathname` to parse out the inscription id. Maybe that is the best option though if we decide inscriptions need to be self aware of the id.

sanjfomojis:
> The more I think about this the less I like my ideas of a client-side /self endpoint or appending an object to window. Yall are making good points. I also agree that it would be very helpful for inscriptions to be able to know their ID. All of that said it still feels yucky to me to make it part of the Ordinals protocol that for the next 100 years clients have to support allowing people to call `window.location.pathname` to parse out the inscription id. Maybe that is the best option though if we decide inscriptions need to be self aware of the id.

This is a similar conclusion to what I came to as well, initially I wasn’t a fan of it not being fully contained within the inscription as it does feel a little hacky but the benefits of self referencing would unlock a lot of cool development and it seems like there aren’t many better options.

Unless @rot13maxi @casey or @raphjaph have any ideas on cleaner ways to achieve an inscription self referencing?

Vanniix:
Yeah, I think we are agreed on this @leonidasord. I have felt for a long time that getting the id from the url felt "wrong" in some way, but since everyone was doing it, I just went along with it. I would be awesome to have a better solution, but I can't think of anything better. The idea of adding a global object would work, but as @devords mentions, its convoluted, and ultimately it has similar problems to the url in that it relies on the client doing something specific.

It's probably also going to be the case that ordinals implementations will find it easier and simpler to just use include the id in the url than adding an extra endpoint or doing some other complex workaround.

thinginab:
> The more I think about this the less I like my ideas of a client-side /self endpoint or appending an object to window. Yall are making good points. I also agree that it would be very helpful for inscriptions to be able to know their ID. All of that said it still feels yucky to me to make it part of the Ordinals protocol that for the next 100 years clients have to support allowing people to call `window.location.pathname` to parse out the inscription id. Maybe that is the best option though if we decide inscriptions need to be self aware of the id.

There really isn't a better way than the url, which is already mandated to work for the /inscription/ endpoint anyways because of recursion.  Stateless web means the server can't know, and any other way of client getting it is messier for server to push it.

While it would be nice to store a delegate once on say a CDN....there you are also stuck in that you still have a way to inject the id in the response, url or otherwise.

So - if the key problem with a site like ord.io is the CDN duplication - maybe they have a way to push a alt/vanity url of some type that doesn't redirect but returns the content stored another place from more than one endpoint? 

This might be what is needed to achieve unique inscription urls for delegates on ord.io @leonidasord 
https://vercel.com/docs/edge-network/rewrites

ZedZeroth:
I've thought of one solution that may sound convoluted at first but it avoids URLs and feels more self-contained. It's similar to the method I'm using currently (passing target sat number via JS) but would be much more efficient. You could implement a special type of delegation whereby a tiny "sat_number:reinscription_point" file/metadata is inscribed for each batch/delegated inscription. The inscriber would need to inscribe onto specific sats, but this is fairly straightforward with batch inscribing, and could even be coded into the batch process behind the scenes, so that the inscriber doesn't need to think/plan it out. That way each inscription knows which sat it's on and which reinscription it is on that sat. It should only be a few extra bytes per inscription.

e.g. I inscribe on sat range 10000 to 20000 and set postage to 500. Inscription 1 is passed "10000:0", Inscription 2 is passed "10500:0" etc. The number after the colon designates reinscription number, so this would be incremented if inscriptions already exist on the sat. Ord still delegates the inscriptions to the delegation target (is there a proper name for this?), but it passes along the sat_number:reinscription_point data that can be read via an endpoint. Then the inscriptions can refer to themselves using /r/sat/SAT_NUMBER/at/REINSCRIPTION_POINT.

In short, ord knows which sats it's inscribing on, so it attaches that sat number to each delegated inscription at the time of inscribing. The delegation target then uses this data to identify each inscription. Thought?

raphjaph (approved):

PR #3211, opened 2024-03-03
Document how inscription contents must be loaded by explorers

devords:
This is a directive for explorers re: how they must load inscriptions and streamline how inscriptions may find their own ID.


devords:
Nevermind. I just saw #3209

PR #3212, opened 2024-03-03
Commit reveal

casey:


casey:
How should I prevent inscribe creating mixed outputs that contain both runes and inscriptions?

This only happens when the premine is nonzero. If the premine is zero, then you're etching a rune, but not allocating it, so there is no danger of runes landing in the same output as an inscription.

What can we do?
- If there is already a change output with no inscriptions, then the rune can land there.
- If there isn't a change output, we can create one, and put the rune there.
- It should always be possible to append change outputs to an inscribe transaction's outputs after we select inputs and outputs, and not disrupt inscription assignment, so let's do that.

Morgandri1:
Hey @casey! Love your work! 
is this the current runes merge?

raphjaph (approved):
LGTM!

PR #3217, opened 2024-03-04
Adjust docker file to work with `just`

carlosaicrag:
Here I've adjusted the docker file to allow developers to work in the container and use the just command.  I've also create a docker-compose file for those that prefer docker-compose and allow for easier multi-container interactions in the future


carlosaicrag:
@casey @raphjaph could I get a review when y'all get a chance plz :) not sure if mentioning is the standard to get a review or if I should just wait for a review.

Thanks :)

Issue #2494, opened 2023-10-02
Allow multiple parents


raphjaph:
#2278 

The indexer should allow multiple parents and also show it in front-end.

Later wallet support should be added

woiwoiwoiwoi:
Is this something that might come in the near future or is it rather far off?

Multiple parents would be so perfect for something we're working on that we're willing to wait a while until they become possible. The question though is how long we would have to wait...

woiwoiwoiwoi:
I just re-read the issue description above and realised it's about the indexer. Does that mean that creating an inscription with multiple parents is already possible, just won't be correctly identified by the indexer?

If so, how would I go about creating an inscription with multiple parents?

UrbanBorNovak:
Multiple parents is awesome!

PR #3227, opened 2024-03-06
Index multi-parent inscriptions

arik-so:
This PR enables indexing and rendering multiple parents in a single inscription, addressing #2494.


src/inscriptions/tag.rs
```diff
@@ -44,45 +54,75 @@ impl Tag {
...
-          .push_slice::<&script::PushBytes>(value.as_slice().try_into().unwrap());
+        TagCodecStrategy::Chunked => {
+          for chunk in value.chunks(MAX_SCRIPT_ELEMENT_SIZE) {
+            tmp = tmp
+              .push_slice::<&script::PushBytes>(self.bytes().try_into().unwrap())
+              .push_slice::<&script::PushBytes>(chunk.try_into().unwrap());
+          }
+        } // TagParsingStrategy::Array => {
+          //   unimplemented!()
+          // }
```
raphjaph, line 71: Do we still have to do something here?

src/inscriptions/tag.rs
```diff
@@ -44,45 +54,75 @@ impl Tag {
...
+      }
+      TagCodecStrategy::Array => {
+        panic!("Array-type fields must not be removed as a simple byte array.")
       }
     }
   }
+
+  pub(crate) fn remove_array_field(self, fields: &mut BTreeMap<&[u8], Vec<&[u8]>>) -> Vec<Vec<u8>> {
+    let values = fields.remove(self.bytes()).unwrap_or_default();
+    // .cloned() doesn't work for Vec<&[u8]>
```
raphjaph, line 125:
```suggestion
  
```

raphjaph (changes requested):
This is looking really good, tests as well. Here are a couple of tests I can still think of:
- [ ] order of parents is correct in database and in explorer display
- [ ] parents paginated correctly

src/wallet/inscribe/batch_file.rs
```diff
@@ -134,7 +134,7 @@ impl Batchfile {
         entry.delegate,
         entry.metadata()?,
         entry.metaprotocol.clone(),
-        self.parent,
+        self.parent.map_or(Vec::new(), |parent| vec![parent]),
```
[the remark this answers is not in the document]
raphjaph: I think this should be fine, just moved it up one level

raphjaph (approved):
LGTM!

let the breeding begin

PR #3244, opened 2024-03-08
Update dump.rs for public descriptor dumping.

cliobitcoinbank:
Added ability to dump public keys by default instead of private keys, needs --dump_private_keys for the private keys to dump instead.  Kept the warning because dumping your public descriptor is dangerous to privacy the same way private keys are dangerous to security.  Not a rust dev, sorry.

### DIFF
--- a/src/subcommand/wallet/dump.rs
+++ b/src/subcommand/wallet/dump.rs
@@ -1,14 +1,25 @@
 use super::*;
+use std::collections::BTreeMap;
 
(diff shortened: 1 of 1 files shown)


raphjaph:
Did you maybe forget to commit all files? I can only see one file and no way to actually read the command-line option. I like having the default be private and then just print public like so:

`ord wallet dump --public-only`

raphjaph:
This PR is unfortunately incomplete and doesn't have any tests. I'll close it for now but please reopen when you add the review comments.

PR #3254, opened 2024-03-10
In case Mint entry contains both term and deadline, test if choose the case that occurs first

sondotpin:


raphjaph (approved):
LGTM, good test!

PR #3256, opened 2024-03-11
Add note field to inscriptions with tag 15

casey:
Add a `note` field to inscriptions. [chisel.xyz](https://chisel.xyz/) puts a note that the inscription was inscribed by chisel in tag 15. This seems reasonable, and similar to BitTorrent's `created_by` field. This PR adds the field to the `Tag` enum, to ensure we don't accidentally use it in the future. This field will probably never be displayed, since users probably don't want the tool which created the inscription to display on /inscription, and they can include that in the metadata if they do.


raphjaph (approved):
LGTM

PR #3259, opened 2024-03-11
HTML/SVG code preview

elocremarc:
Ord io and Geniidata both have a code preview on the inscription pages for html and svg. This would be a nice feature for the inscription page.
We can use the highlight Js with the XML to hightlight HTML and SVG. I initialized a new code-preview template.

We could swap the Iframe of the html preview with the html code preview. However we try and minimize client JS so maybe there is a different way of doing it? Open to suggestions.

PR is far from working just throwing down some ideas


raphjaph:
What's the status on this?

PR #3262, opened 2024-03-11
ci: add pre-commit hook

bingryan:
## Description

add git `pre-commit` hook 

## Usage

`step1`:  install pre-commit for you local repo

```
just git-pre-commit
``` 

`step2`: auto format submitted `**.rs` files for your `git commit xxxx`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,6 +114,7 @@ mod inscriptions;
 mod object;
 pub mod options;
 pub mod outgoing;
+mod re;
 mod representation;
 pub mod runes;
 mod server_config;
--- a/src/outgoing.rs
+++ b/src/outgoing.rs
@@ -24,8 +24,6 @@ impl FromStr for Outgoing {
 
   fn from_str(s: &str) -> Result<Self, Self::Err> {
     lazy_static! {
-      static ref SATPOINT: Regex = Regex::new(r"^[[:xdigit:]]{64}:\d+:\d+$").unwrap();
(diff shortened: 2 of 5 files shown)


bingryan:
After several PRs, i sometimes submit some unformatted code, then get ` CI lint error `.  i think add `pre-commit` hook is useful for this case.

bin/pre-commit
```diff
@@ -0,0 +1,12 @@
...
+
+set -euo pipefail
+
+REPO_ROOT_DIR=$(git rev-parse --show-toplevel)
+
+files=$(git diff --cached --name-only --diff-filter=ACMR | grep -Ei "\.rs$" || true)
+
+if [[ -n $files ]]; then
+    cargo fmt --all
+    git add -- $(echo "$files" | paste -s -d " " -)
```
raphjaph, line 11:
```suggestion
    cargo fmt --check
```

bingryan: This only automatically formats the submitted rust file.

for only `cargo fmt --check`, if there are unformatted files, you will need to execute `cargo fmt --all` and then `git add xxx` again.

bingryan:
only format the submitted Rust file now

raphjaph:
It's not a good idea to change files in a pre-commit hook. So we only want to check things not write to any files

raphjaph (approved):
lgtm

Made it very simple and generic for now, we can add more complex hooks later.

Issue #3269, opened 2024-03-12
Better error message when running wallet commands without server running in the background


casey:

raphjaph:
Closed by #3272

Issue #3263, opened 2024-03-11
bug: Unable to create a reception address


starker-xp:
versions :
windows 11 prox64
btc cli 25.0
ord 0.16.0-rc1

rpc.conf
```cnf

##
## bitcoin.conf configuration file. Lines beginning with # are comments.
##
daemon=0
regtest=1

listen=1

# JSON-RPC options (for controlling a running Bitcoin/bitcoind process)

#rpcport=8332

# server=1 tells Bitcoin-Qt and bitcoind to accept JSON-RPC commands
server=1
#prune=5500
txindex=1
datadir=H:\btc\data

fallbackfee=0.000002
minrelaytxfee=0.000002
```

bitcoind -conf="H:\btc\rpc.conf"
bitcoin-cli -conf="H:\btc\rpc.conf" createwallet dev
bitcoin-cli -conf="H:\btc\rpc.conf" generatetoaddress 101 $(bitcoin-cli -conf="H:\test\rpc.conf" -rpcwallet=dev getnewaddress)

ord -r --cookie-file "H:\test\data\regtest\.cookie" wallet --name=ord_dev create

ord -r --cookie-file "H:\test\data\regtest\.cookie" wallet --name=ord_dev receive
```sh
thread 'main' panicked at src\wallet.rs:112:54:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

raphjaph:
Did you start the `ord server`?

The new wallet requires running it in parallel

raphjaph:
Added a warning in #3272

PR #3272, opened 2024-03-12
refactor: better error message with without server running

bingryan:
## Description

add better error log for `ord server` is not running

## Related Issue

Related #3263 #3269


raphjaph (approved):
looks good! Simplified it a bit

PR #3289, opened 2024-03-14
Return effective content type in JSON API.

### LINKED ISSUE #3055: Delegator/delegate content type mismatch
When an inscription has a delegate, the content and preview endpoints are served with the delegate's mime type, as one would expect.

However, the delegating inscription's details page shows the mime type corresponding to the inscription itself which, while technically correct, doesn't tell the whole tale. It might be prudent to show the mime type of the inscription whose content is actually being returned, possibly alongside the inscription's own mime type.

arik-so:
For inscriptions with delegates, the content type of the underlying delegate should be exposed. Fixes #3055.


raphjaph (approved):
LGTM!

PR #3301, opened 2024-03-17
Allow multiple delegates.

arik-so:
This PR modifies server behavior to return the content of the first delegate it finds, absent which it returns the inscription's own content.

https://github.com/ordinals/ord/issues/4114

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -118,7 +118,7 @@ pub struct InscriptionRecursive {
   pub charms: Vec<Charm>,
   pub content_type: Option<String>,
   pub content_length: Option<usize>,
-  pub delegate: Option<InscriptionId>,
+  pub delegates: Vec<InscriptionId>,
   pub fee: u64,
   pub height: u32,
   pub id: InscriptionId,
--- a/src/index.rs
+++ b/src/index.rs
@@ -2087,16 +2087,14 @@ impl Index {
       Charm::Lost.set(&mut charms);
     }
 
-    let effective_mime_type = if let Some(delegate_id) = inscription.delegate() {
-      let delegate_result = self.get_inscription_by_id(delegate_id);
-      if let Ok(Some(delegate)) = delegate_result {
-        delegate.content_type().map(str::to_string)
-      } else {
-        inscription.content_type().map(str::to_string)
-      }
-    } else {
-      inscription.content_type().map(str::to_string)
-    };
+    let effective_mime_type = inscription
+      .delegates()
+      .iter()
+      .find_map(|delegate| self.get_inscription_by_id(*delegate).unwrap_or(None))
+      .as_ref()
+      .unwrap_or(&inscription)
+      .content_type()
+      .map(str::to_string);
 
     Ok(Some((
       api::Inscription {
--- a/src/index.rs
+++ b/src/index.rs
@@ -2426,7 +2426,7 @@ mod tests {
     {
       let context = Context::builder().build();
       context.mine_blocks(1);
-      let txid = context.core.broadcast_tx(template.clone());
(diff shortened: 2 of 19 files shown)


src/subcommand/server.rs
```diff
@@ -1287,11 +1287,14 @@ impl Server {
...
-        inscription = index
-          .get_inscription_by_id(delegate)?
-          .ok_or_not_found(|| format!("delegate {inscription_id}"))?
-      }
+      // return the first inscription where the delegate exists
+      // probably worth considering some sort of limitation to mitigate "DoSability"
+      let inscription_override = inscription
+        .delegates()
+        .iter()
+        .find_map(|delegate| index.get_inscription_by_id(*delegate).unwrap_or(None));
```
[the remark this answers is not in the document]
cryptocj: Can we add a length number to avoid the Dos concern?
And extract a function to remove the repeat code below.
```rust
fn find_inscription<'a>(delegates: &'a [usize], index: &'a Index, max_length: usize) -> Option<&'a Inscription> {
    delegates
        .iter()
        .take(max_length)
        .find_map(|delegate| index.get_inscription_by_id(*delegate).unwrap_or(None))
}

// Usage
let inscription_override = find_inscription(inscription.delegates(), &index, max_length);
```

raphjaph:
For reference #4114

src/index.rs
```diff
@@ -2087,16 +2087,14 @@ impl Index {
...
-      } else {
-        inscription.content_type().map(str::to_string)
-      }
-    } else {
-      inscription.content_type().map(str::to_string)
-    };
+    let effective_mime_type = inscription
+      .delegates()
+      .iter()
+      .find_map(|delegate| self.get_inscription_by_id(*delegate).unwrap_or(None))
```
casey, line 2093: We should handle the error here instead of using `unwrap_or`.

src/subcommand/server.rs
```diff
@@ -1583,11 +1583,12 @@ impl Server {
...
 
-      if let Some(delegate) = inscription.delegate() {
-        inscription = index
-          .get_inscription_by_id(delegate)?
-          .ok_or_not_found(|| format!("delegate {inscription_id}"))?
-      }
+      let inscription_override = inscription
+        .delegates()
+        .iter()
+        .find_map(|delegate| index.get_inscription_by_id(*delegate).unwrap_or(None));
```
casey, line 1589: We should handle the error here instead of using `unwrap_or`. All errors must be handled >:]

src/subcommand/server.rs
```diff
@@ -1707,11 +1708,12 @@ impl Server {
...
```
casey, line 1714: Handle errors.

casey (changes requested):
Seems mostly reasonable to me! We should always handle errors instead of doing `unwrap_or`, or some other suppression.

I think we should impose a limit of 100 delegates, and any delegates after are just ignored. I can't imagine someone needing more, and it's more painful to add a restriction later than to relax a restriction.

src/api.rs
```diff
@@ -118,7 +118,7 @@ pub struct InscriptionRecursive {
   pub charms: Vec<Charm>,
   pub content_type: Option<String>,
   pub content_length: Option<usize>,
-  pub delegate: Option<InscriptionId>,
+  pub delegates: Vec<InscriptionId>,
```
raphjaph, line 121: We cannot change the return type since it's a recursive endpoint. We'll have to return the effective delegate here instead in an `Option<InscriptionId>`.

casey:

arik-so:
Thanks, Casey! And no worries, it used antiquated authorship mechanics anyway.

PR #3307, opened 2024-03-18
ignore invalid instruction of rune commit tx

nowherekai:
fix https://github.com/ordinals/ord/issues/3306


casey:
We get transactions from bitcoin core, so I don't think it's possible to encounter a transaction with an invalid instruction. Have you found a case where it is possible?

PR #3325, opened 2024-03-19
Invalid runestones are cenotaphs

casey:


raphjaph (approved):

jerryfane:
What was wrong with the term "Burn"?

casey:
Because "burn" is used as a term for "burning bitcoin" or "burning runes". Here, cenotaph is type of runestone, which in addition to other things, burns input runes.

PR #3329, opened 2024-03-19
Handle previous_output store is not exists

lugondev:


lugondev:
yeh, could @raphjaph  review this PR for me?

lugondev:
@casey thankss

casey:
This is not currently the way to fix this. The bug here is that we aren't processing transactions before `Chain::first_rune_height`, which we need to do, instead of just suppressing the bug.

Issue #3318, opened 2024-03-19
Rune gambling


casey:
Use block hash as a source of randomness. Huge concern is miners being able to manipulate bets by grinding favorable block hashes. A possible mitigation for that is to make the bet in block N and resolve the bet in block M > N.

Gambling could be simple odds, i.e., double or nothing, or a form of collaborative betting where runes are assigned to a random output of the transaction, which was @sondotpin's idea.

cryptoni9n:
some gambol ideas:

**standard flip/war** (_2 player_): some number of runes are locked into bet by each party (can be x for x, or  x for y).   Variations could be flagged for best 2 out of 3, allow double or nothing, and so on.  Flip would then assign a random number 1-10 to each party - high number wins the rune pot and is assigned total output for bet resolution.

> **example 1**: Player 1 bets 2 UNCOMMON.GOODS and Player 2 bets 2 UNCOMMON.GOODS.  Flip initiates and assigns Player 1 the value `5` and Player 2 is assigned the value `7`.  Player 2 wins the pot and wins total of 4 UNCOMMON.GOODS (2 from Player 1, and 2 that he put into the pot himself)

> **example 2**: Player 1 bets 2 UNCOMMON•GOODS and Player 2 bets 20,000 GAME•WORLD•COIN.  Flip initiates and assigns Player 1 the value `5` and Player 2 is assigned the value `2`.  Player 1 wins the pot and wins total of 2 UNCOMMON.GOODS and 20,000 GAME•WORLD•COIN.

**BTC price bet (1-many):** bet creator bets Y runes that BTC/USD price will or will not break X by block height Z, according to some designated and agreed upon source of truth.  Bet creator would specify a TTL block  height that keeps the bet open for the takers to participate by.  Prior to TTL height, the Taker would take opposite side and need to match the rune bet with a like and in kind rune stack.  Can be configured to accept one or more takers per bet instantiation, depending on initial bet configuration and rune total posted.  If no takers by TTL height, bet closes and runes are returned to bet creator.

[...]

PR #3330, opened 2024-03-19
Add randomness ouput

lugondev:
Rune gambling #3318 
@casey this is my solution


casey:
Thanks for the PR! We're not ready to consider adding this yet, and the design is still uncertain. In particular, I have concerns that a simple implementation would be trivially gamable by miners.

PR #3339, opened 2024-03-22
Allow runestone to be deserialized from json

emilcondrea:
Allow runestone to be deserialized from json.

This would be super-helpful intermediary step until we get a separate library for encoding/decoding payloads.

@casey @raphjaph , I know there are some versions of these that are already deserializable and used by the wallet(eg. wallet/inscribeetch.rs), but if wallet does not expose customization for some fields, they will not be included there. This is needed for protocol convertors, to actually have the exact same structures.


emilcondrea:
Thanks for merging this!

Issue #3030, opened 2024-01-18
Add charms to /sat page


casey:
Many charms are based on the sat, so we should display them on the sat page. Rare, uncommon, coin, block 9, etc.

PR #3340, opened 2024-03-22
Add sat related charms to /sat page

markovichecha:
Added sat based charms to the /sat page. Resolves: #3030 

Also made a minor tweak in forbid.sh to handle the silent failure when ripgrep is missing.


markovichecha:
@casey thanks for looking into my PR, what are your thoughts on doing something like that instead of using u16?
 
`#[derive(Debug, Clone, PartialEq, Eq, Default)]
pub(crate) struct Charms(pub u16);

impl Charms {
  pub(crate) fn new() -> Self {
    Self(0)
  }

  pub(crate) fn set(&mut self, charm: Charm) -> Charms {
    self.0 |= 1 << charm as u16;

    self.clone()
  }

  pub(crate) fn is_set(&self, charm: Charm) -> bool {
    self.0 & (1 << charm as u16) != 0
  }

  pub(crate) fn is_empty(&self) -> bool {
    self.0 == 0
  }

  pub(crate) fn unset(&mut self, charm: Charm) -> Charms {
    self.0 &= !(1 << charm as u16);

    self.clone()
  }

  pub(crate) fn iter(&self) -> impl Iterator<Item = Charm> + '_ {
    Charm::ALL
      .iter()
      .copied()
      .filter(move |&charm| self.is_set(charm))
  }
}`

markovichecha:
@casey thanks for looking into PR, what are your thoughts on something like that instead of using u16? https://gist.github.com/markovichecha/03f491c990520bc5988175fdd2b0a11d

casey:
Thanks for the PR! I changed some stuff, since I wanted to serialize `Charm`s directly, instead of converting to strings, as well as add charm for the mythic sat, which didn't previously have one, because charms were only displayed for inscriptions.

casey:
> @casey thanks for looking into PR, what are your thoughts on something like that instead of using u16? https://gist.github.com/markovichecha/03f491c990520bc5988175fdd2b0a11d

This looks great, free free to open a PR!

PR #3341, opened 2024-03-22
Update redb dependency to 2.0.0

cberner:
I implemented this while testing the new redb release, so figured I'd send the PR. Please note that the file format is not backwards compatible, so you'll want to add some upgrade logic if you need it to be backwards compatible.


casey:
Excellent PR! Look at those lifetimes disappear!

PR #3348, opened 2024-03-23
Remove unnecessary code

lugondev:
1. Remove unnecessary parentheses
2. Remove unnecessary path prefix


casey:
I added `Self` back. We import the variants of `Language` and `ImageRendering` so `Media::Table` will be more readable. I otherwise don't really like to use enum variants directly.

lugondev:
> I added `Self` back. We import the variants of `Language` and `ImageRendering` so `Media::Table` will be more readable. I otherwise don't really like to use enum variants directly.

thanks <3

PR #3358, opened 2024-03-24
Make rune parent clickable

### LINKED ISSUE #3355: Make rune parent clickable on /rune

raphjaph:
Fixes #3355


casey (commented):
LGTM. We can figure out if we want it to be full size later.

casey (approved):

PR #3365, opened 2024-03-25
Allow supply-capped mints

casey:


revofusion:
`create_rune_entry` sets `burned: balance`, which at that point in time would be equal to all runes unallocated.

However, we still have the catch all case of `*allocated[vout].entry(id).or_default() += balance;` which may allocate unallocated balance after the create_rune_entry call

src/index/entry.rs
```diff
@@ -38,30 +38,48 @@ pub struct RuneEntry {
...
     if let Some(deadline) = mint.deadline {
       if block_time >= deadline {
-        return Err(MintError::Deadline((self.spaced_rune.rune, deadline)));
+        return Err(MintError::Deadline(deadline));
       }
     }
 
-    Ok(mint.limit.unwrap_or(runes::MAX_LIMIT))
+    if let Some(count) = mint.count {
+      if self.mints > count {
```
revofusion, line 64: @casey I believe this should be self.mints >= count, will allow 1 mint over count otherwise

casey: Good catch! Fixed.

raphjaph (approved):

Issue #3367, opened 2024-03-25
Update the required rust version for build in readme


cryptoni9n:
Currently, the readme states: 

> ord requires rustc version 1.67.0 or later. Run rustc --version to ensure you have this version. 

When trying to build ord from source, the following message is received: 
```

error: package `clap_builder v4.5.2` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.72.0
Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap_builder@4.5.2 --precise ver
where `ver` is the latest version of `clap_builder` supporting rustc 1.72.0
```
running rustup update and updating to 1.77.0 bypasses this error.

PR #3370, opened 2024-03-25
docs: add rust-toolchain.toml

bingryan:
## Description

Use `rust-toolchain.toml` to describe rust toolchain information

## Related Issue

Related #3367


rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.75.0"
```
markovichecha, line 2: It should be rust 1.77.0 as for now

casey:
I like having CI use the latest stable, so we notice breakages due to new releases of Rust.

Issue #2619, opened 2023-11-01
Index inscriptions by address


77656233:
Is it possible to add a search by address endpoint to get a list of all inscriptions an address holds?

raphjaph:
It's definitely possible but I'm hesitant to add this because it adds a huge (more than 60gb) database table. It could be added behind an optional flag though

ChristianGrieger:
> It's definitely possible but I'm hesitant to add this because it adds a huge (more than 60gb) database table. It could be added behind an optional flag though

That would be amazing as i guess a lot of users would love to use it :) Is there a chance you can take a look into it?

elocremarc:
This would be a nice recursive endpoint too. 
Would also be nice to navigate to a address page on the inscription page when clicking it. 
Thing to consider how big will this table grow if we get millions more addresses in inscriptions or something?
If we max min this and every 460 million bitcoin addresses hold inscriptions thats a big table.

ChristianGrieger:
True ,would be a big table but if it's an optional flag it should be fine for everyone :)

ChristianGrieger:
@raphjaph, could you provide an estimate for this request? We, and undoubtedly others, would greatly appreciate having access to that endpoint. Thanks a lot!

raphjaph:
Do you mean as a recursive endpoint or just a normal html page?

At the moment we're working on database optimizations so want to control the amount of tables we have. I think I'll code something quick up and do some testing to see how it impacts performance.

ChristianGrieger:
> Do you mean as a recursive endpoint or just a normal html page?
> 
> At the moment we're working on database optimizations so want to control the amount of tables we have. I think I'll code something quick up and do some testing to see how it impacts performance.

I talk about the json api endpoint that the server offers.

elocremarc:
> Do you mean as a recursive endpoint or just a normal html page?

That would be nice but we can add that later I think a html page would be a good start.

markovichecha:
Hey, may I handle this task? To clarify, do we need to create an index that includes addresses and their corresponding inscription IDs? And add an html page that shows all inscriptions for given address after that.

raphjaph:
we got this now

PR #3372, opened 2024-03-25
Index inscriptions by address

markovichecha:
Added an address to inscriptions index and address endpoint with inscriptions. Resolves: #2619 

image

Address index relies on transactions index, how should we handle this? Fail to startup if there are no --index-transactions flag?

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -9,6 +9,12 @@ pub use crate::templates::{
 };
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct Address {
+  pub address: bitcoin::Address<NetworkUnchecked>,
+  pub inscriptions: Vec<InscriptionId>,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct Block {
   pub hash: BlockHash,
   pub target: BlockHash,
@@ -119,7 +125,7 @@ pub struct Inscriptions {
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct Output {
-  pub address: Option<Address<NetworkUnchecked>>,
+  pub address: Option<bitcoin::Address<NetworkUnchecked>>,
   pub indexed: bool,
   pub inscriptions: Vec<InscriptionId>,
   pub runes: Vec<(SpacedRune, Pile)>,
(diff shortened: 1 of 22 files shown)


markovichecha:
@raphjaph *actively harassing to check on this PR as were suggested on stream*, thanks

ynohtna92:
> Address index relies on transactions index, how should we handle this? Fail to startup if there are no --index-transactions flag?

I don't think it should fail, I implemented the address index onto my own branch a while ago without the transaction index. Users shouldn't need to double up on storing tx information as it is already done with -txindex.

Perhaps use it if it is available otherwise use the rpc getTransaction.

https://github.com/ynohtna92/ord-litecoin/blob/30825f67ccca4921c8f96e6a7d212cd3660f1f70/src/index/updater/inscription_updater.rs#L413-L419

Klyaus:
Hello! I’m experimenting with your version when indexing testnet (I launched Ord in Docker with the command “ord --data-dir /index-data --index-transactions --index-addresses server”). And I encountered this behavior:
```
[2024-04-18T11:37:24Z INFO  ord::index::updater] Block 2441766 at 2023-07-14 03:35:26 UTC with 21 transactions…
[2024-04-18T11:37:24Z INFO  ord::index::updater] Block receiver disconnected: sending on a closed channel
[2024-04-18T11:37:24Z INFO  ord::index] script is not a p2pkh, p2sh or witness program
[2024-04-18T11:37:24Z WARN  ord::subcommand::server] Updating index: script is not a p2pkh, p2sh or witness program
```
after that, the indexer starts scanning a few days before the specified date, and again hits the same block, and so on in a circle. Have you seen this kind of behavior?

markovichecha:
Hey @Klyaus. I haven't met this error, but may be it was due to not indexing mainnet till the end. I'll return to this PR soon and will change transaction index behavior and try to fix this issue.

Klyaus:
> Hey @Klyaus. I haven't met this error, but may be it was due to not indexing testnet. I'll return to this PR soon and will change transaction index behavior and try to fix this issue.

Great! Thanks for the answer. I also think this is due to the fact that this is a testnet.

raphjaph:
We recently merged #3757, which adds a more comprehensive address to output tracker. If you look at it you'll see the html page is just a list of outputs. If you want you could add more descriptive and aggregated information like inscriptions, btc, sats and runes. Also your PR links addresses we show in places to the address html page. If you feel like you could probably use some of the work you did here and just build on top of what I coded up. Best to chunk these up into different PRs. Let me know if you're up for that :)

PR #3380, opened 2024-03-27
Document runes

casey:


rot13maxi:
yo dog. i heard you like prs. so i pr'd your pr so you can review while you review

https://github.com/casey/ord/pull/2

raphjaph (approved):

raphjaph (approved):
LFG!

docs/src/runes.md
```diff
@@ -0,0 +1,158 @@
...
+#### Start Offset
+
+A mint is open starting in the block whose height is equal to the start offset
+plus the height of the block in which the rune was etched.
+
+#### End Offset
+
+A rune may not be minted in or after the block whose height is equal to the end
+offset plus the height of the block in which the rune was etched.
+
```
twhit223, line 112: What is the behavior if neither a premine nor terms are specified? It seems the default for each of those is 0/do nothing. Does that etch a valid rune with 0 max supply (effectively "burning" the name), or does it create a cenotaph (and burn all runes that are also on that tx)?

casey: If neither premine nor terms are supplied, the supply is zero, but the runestone is not a cenotaph.

PR #3405, opened 2024-03-29
Add docs for runes send

rot13maxi:
it was not intuitive to me how to send runes. added to the docs.


docs/src/guides/wallet.md
```diff
@@ -386,6 +386,36 @@ running:
...
+
+```
+ord wallet transactions
+```
+
+Once the send transaction confirms, the recipient can confirm receipt by
+running:
+
+```
+ord wallet inscriptions
```
raphjaph, line 416:
```suggestion
ord wallet balance
```

docs/src/guides/wallet.md
```diff
@@ -386,6 +386,36 @@ running:
...
+ord wallet receive
+```
+
+Send the runes by running:
+
+```
+ord wallet send --fee-rate <FEE_RATE> <ADDRESS> <RUNES_AMOUNT>
+```
+where `RUNES_AMOUNT` is the number of runes to send, a `:` character, and the name of the rune. For example if you want to send
+1000 of the EXAMPLE rune, you would use `1000:EXAMPLE`.
```
casey, line 404: Maybe also add the full command:

```
ord wallet send --fee-rate 1 SOME_ADDRESS 1000:EXAMPLE
```

PR #3437, opened 2024-04-01
Implement burn

onchainguy-btc:
After my first burn PR (#2766) got a bit outdated, I created a new one. Tried to make as less changes as possible. This PR does not support burn messages yet. Hope we can get the general burning merged first and then I'm happy to add support for an optional burn message via data pushes.


src/wallet/transaction_builder.rs
```diff
@@ -99,9 +121,10 @@ pub struct TransactionBuilder {
   inputs: Vec<OutPoint>,
   inscriptions: BTreeMap<SatPoint, Vec<InscriptionId>>,
   locked_utxos: BTreeSet<OutPoint>,
+  network: Network,
   outgoing: SatPoint,
-  outputs: Vec<(Address, Amount)>,
-  recipient: Address,
+  outputs: Vec<(OutputType, Amount)>,
```
raphjaph, line 126:
```suggestion
  outputs: Vec<TxOut>,
```

Let's change this to a vec of TxOut since that struct encapsulates these two things.

raphjaph (changes requested):
Finally having a look at this. There's just two things but they probably require a bit of refactoring. Especially the new command `ord wallet burn` I think is important because burning is quite drastic and we want it to be a very distinct action.

onchainguy-eth:
@raphjaph thanks for the feedback, will get that done asap!

src/wallet/transaction_builder.rs
onchainguy-eth: Refactored ✅

src/subcommand/wallet/send.rs
```diff
@@ -1,4 +1,23 @@
...
+
+fn parse_address(arg: &str) -> Result<ParsedAddress, bitcoin::address::Error> {
+  if arg == "burn" {
+    let builder = script::Builder::new()
+      .push_opcode(opcodes::all::OP_RETURN);
+    Ok(ParsedAddress::ScriptBuf(builder.into_script()))
+  } else {
+    Ok(ParsedAddress::Address(Address::from_str(arg)?))
+  }
+}
```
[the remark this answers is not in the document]
onchainguy-eth: Refactored ✅

onchainguy-eth:
@raphjaph just implemented the changes. Let me know if I can improve on anything else. Can also write some more tests if you have ideas for some good ones.

src/subcommand/wallet/burn.rs
```diff
@@ -0,0 +1,31 @@
...
+  #[arg(long, help = "Don't sign or broadcast transaction")]
+  pub(crate) dry_run: bool,
+  #[arg(long, help = "Use fee rate of <FEE_RATE> sats/vB")]
+  fee_rate: FeeRate,
+  #[arg(
+    long,
+    help = "Target <AMOUNT> postage with sent inscriptions. [default: 10000 sat]"
+  )]
+  pub(crate) postage: Option<Amount>,
+  outgoing: Outgoing,
```
raphjaph, line 14:
```suggestion
  inscription_id: InscriptionId,
```
For now we only want to enable burning inscriptions, so let's make sure only that is possible. We'll add burning functionality to the other types (sats, btc, runes) later.

src/subcommand/wallet/send.rs
```diff
@@ -1,18 +1,32 @@
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum ParsedAddress {
+  Address(Address<NetworkUnchecked>),
+  ScriptBuf(ScriptBuf),
+}
+
+fn parse_address(arg: &str) -> Result<ParsedAddress, bitcoin::address::Error> {
+  Ok(ParsedAddress::Address(Address::from_str(arg)?))
+}
```
raphjaph, line 14:
```suggestion
```
Let's get rid of all of these. The burn command should just use the transaction builder directly with the correct `TxOut`, which can include an OP_RETURN script.

src/subcommand/wallet/send.rs
```diff
@@ -1,18 +1,32 @@
...
+  pub(crate) fee_rate: FeeRate,
   #[arg(
     long,
     help = "Target <AMOUNT> postage with sent inscriptions. [default: 10000 sat]"
   )]
   pub(crate) postage: Option<Amount>,
-  address: Address<NetworkUnchecked>,
-  outgoing: Outgoing,
+  #[arg(value_parser = parse_address)]
+  pub(crate) address: ParsedAddress,
```
raphjaph, line 28:
```suggestion
  pub(crate) address: Address,
```

raphjaph (changes requested):
In OP_RETURNs there is no dust_limit and I see that you correctly disable the check for that. One thing you could try doing is only burn the exact sat that the inscription is on instead of the whole UTXO. This will require some careful changes to the `TransactionBuilder`, which should then be tested as well in the unit tests of the `transaction_builder.rs` file.

onchainguy-eth:
> In OP_RETURNs there is no dust_limit and I see that you correctly disable the check for that. One thing you could try doing is only burn the exact sat that the inscription is on instead of the whole UTXO. This will require some careful changes to the `TransactionBuilder`, which should then be tested as well in the unit tests of the `transaction_builder.rs` file.

@raphjaph Do you think that's worth it? If I'm not mistaking doing so would add additional inputs to the transaction for extracting the sat. The fee for that could potentially be higher than what the inscription sats are worth. 

As an alternative idea: What if we exit the `ord wallet burn` command with an error if the inscription is on a utxo with more than 10_000 sats. This way we can prevent burning large bitcoin amounts by mistake. If people still want to burn the inscriptions, they can first transfer it to a different, smaller output.

crates/ordinals/src/charm.rs
```diff
@@ -62,6 +64,7 @@ impl Charm {
       Self::Unbound => "🔓",
       Self::Uncommon => "🌱",
       Self::Vindicated => "❤️‍🔥",
+      Self::Burned => "💀🔥",
```
raphjaph, line 67:
```suggestion
      Self::Burned => "🔥",
```

raphjaph (changes requested):

raphjaph:
> @raphjaph Do you think that's worth it? If I'm not mistaking doing so would add additional inputs to the transaction for extracting the sat. The fee for that could potentially be higher than what the inscription sats are worth.

That's true, then let's just leave it how it is for now.

> As an alternative idea: What if we exit the `ord wallet burn` command with an error if the inscription is on a utxo with more than 10_000 sats. This way we can prevent burning large bitcoin amounts by mistake. If people still want to burn the inscriptions, they can first transfer it to a different, smaller output.

Sounds reasonble

onchainguy-eth:
> > @raphjaph Do you think that's worth it? If I'm not mistaking doing so would add additional inputs to the transaction for extracting the sat. The fee for that could potentially be higher than what the inscription sats are worth.
> 
> That's true, then let's just leave it how it is for now.
> 
> > As an alternative idea: What if we exit the `ord wallet burn` command with an error if the inscription is on a utxo with more than 10_000 sats. This way we can prevent burning large bitcoin amounts by mistake. If people still want to burn the inscriptions, they can first transfer it to a different, smaller output.
> 
> Sounds reasonble

Cool, will adapt to that!

raphjaph:
Let me know when this is ready for review :)

onchainguy-btc:
@raphjaph just saw you merged master? Was that on purpose? I can also just rebase if you like.

raphjaph:
@onchainguy-btc don't worry about rebasing, I'm going to squash merge this anyway. Gonna have a look at this today or tomorrow

raphjaph (changes requested):
See comments above

onchainguy-btc:
> @onchainguy-btc don't worry about rebasing, I'm going to squash merge this anyway. Gonna have a look at this today or tomorrow.
> 
> The tests and linting also don't seem to pass yet. If you could fix, that would be great :)
> 
> You can run our CI if you install `just` and run `just ci`.

@raphjaph thanks for the hint. Just made `just ci` pass.

src/subcommand/wallet/burn.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl Burn {
+  pub(crate) fn run(self, wallet: Wallet) -> SubcommandResult {
+    let unsigned_transaction = match self.outgoing {
+      Outgoing::InscriptionId(id) => {
+        let inscription_info_map = wallet.inscription_info();
+        let inscription_info = inscription_info_map
```
raphjaph, line 22:
```suggestion
        let inscription_info = wallet.inscription_info()
```

src/subcommand/wallet.rs
```diff
@@ -47,6 +50,8 @@ pub(crate) enum Subcommand {
   Balance,
   #[command(about = "Create inscriptions and runes")]
   Batch(batch_command::Batch),
+  #[command(about = "Burn sat or inscription")]
```
raphjaph, line 53:
```suggestion
  #[command(about = "Burn an inscription")]
```

src/subcommand/wallet/burn.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        Self::create_unsigned_burn_transaction(
+          &wallet,
+          inscription_info.satpoint,
+          self.postage,
+          self.fee_rate,
+        )?
+      }
+      _ => panic!("Outgoing type is not an inscription"),
```
raphjaph, line 39:
```suggestion
      _ => panic!("Only inscriptions can be burned for now"),
```

src/wallet/transaction_builder.rs
```diff
@@ -90,6 +92,11 @@ impl Display for Error {
 }
 
 impl std::error::Error for Error {}
+impl From<bitcoin::address::Error> for Error {
```
raphjaph, line 95:
```suggestion

impl From<bitcoin::address::Error> for Error {
```

raphjaph (changes requested):
Just did a cursory review and except for a couple nits this looks good!

I'm going to do a more in-depth review with @casey to make sure the burning mechanism makes sense in the long-term vision of the protocol.

onchainguy-btc:
> Just did a cursory review and except for a couple nits this looks good!
> 
> I'm going to do a more in-depth review with @casey to make sure the burning mechanism makes sense in the long-term vision of the protocol.

Thanks, just merged your suggestions. Just had to add one `.clone()` to make the build work (see [this commit](https://github.com/ordinals/ord/pull/3437/commits/f66e5cb5cbed4587de4adcabaf16a06de01c2568)).

After we get this merged I'm happy to continue with the burn feature. For example, by working on [this](https://github.com/ordinals/ord/issues/3025).

casey (changes requested):
There's a problem with how `is_burned` is set in the index. `is_burned` is set inside a loop which goes over every inscription in the transaction's inputs. However, that value is then used for all the inscriptions which are moving. See my comment for more details.

The wallet changes here are very large, so I think I want to split this into two PRs. First, a much simpler PR which sets the burned charm, and then a second PR which covers the wallet changes.

The first PR will also need a test which is specific to the index / server. I started writing a test, which looks like this:

```rust
  #[test]
  fn burned_charm() {
    let server = TestServer::builder().chain(Chain::Regtest).build();

    server.mine_blocks(1);

    let inscription = Inscription {
      content_type: Some("text/html".into()),
      body: Some("foo".into()),
      ..default()
    };

    let txid = server.core.broadcast_tx(TransactionTemplate {
      inputs: &[(1, 0, 0, inscription.to_witness())],
      outputs: 0,
      op_return_index: Some(0),
      op_return_value: Some(10000),
      op_return: Some(
        script::Builder::new()
          .push_opcode(opcodes::all::OP_RETURN)
          .into_script(),
      ),
      ..default()
    });

    server.mine_blocks(1);

    let id = InscriptionId { txid, index: 0 };

    pretty_assert_eq!(
      server.get_json::<api::InscriptionRecursive>(format!("/r/inscription/{id}")),
      api::InscriptionRecursive {
        charms: vec![Charm::Burned],
        content_type: Some("text/html".into()),
        content_length: Some(3),
        delegate: None,
        fee: 0,
        height: 2,
        id,
        number: 0,
        output: OutPoint { txid, vout: 0 },
        sat: None,
        satpoint: SatPoint {
          outpoint: OutPoint { txid, vout: 0 },
          offset: 0
        },
        timestamp: 2,
        value: Some(50 * COIN_VALUE),
      }
    );
  }
```

The `TransactionTemplate` will also need a new field, `op_return_value: Option<u64>` which allows making the value of the OP_RETURN output nonzero, so the burned inscriptions lands on it.

onchainguy-btc:
@casey I'm stupid 🥲 Guess my overall understanding of Rust + the inscription updater was not good enough when starting the PR. Hopefully it gets better with [this new PR](https://github.com/ordinals/ord/pull/3836). This only implements the indexing.

raphjaph:
@onchainguy-btc should we review this as the wallet PR or do you want to open a new PR and make changes?

onchainguy-btc:
@raphjaph just rebased this and did some cleanups so that it now only includes the wallet changes. Appreciate a new review :)

onchainguy-btc:
In my last commit I also protected last postage from being burned. Not 100% sure if that makes sense though.

raphjaph (changes requested):
@onchainguy-btc This looks almost ready to go!

Could you add some more tests:

1. Multiple inscriptions in the same utxo and then try to burn and make sure it either fails or only burns the correct inscription.
2. An inscription and a rune in the same utxo and then it should fail.
3. Try `ord wallet burn <RUNE_ID>` and it should fail.

onchainguy-btc:
@raphjaph just implemented the tests you requested ✅

raphjaph (approved):
Awesome, LGTM!

PR #3440, opened 2024-04-02
Lookup rune number

lugondev:


src/subcommand/server.rs
```diff
@@ -2661,6 +2664,45 @@ mod tests {
...
+        }],
+        etching: Some(Etching {
+          rune: Some(rune),
+          ..default()
+        }),
+        ..default()
+      },
+      1,
+      None,
+    );
```
raphjaph, line 2695: try etching like 10 runes and query the first the 5th and the last

src/index.rs
```diff
@@ -867,6 +867,25 @@ impl Index {
...
+    {
+      Some(result) => {
+        let rune_result =
+          result.map(|(_id, entry)| RuneEntry::load(entry.value()).spaced_rune.rune);
+        Ok(rune_result.ok())
+      }
+      None => Ok(None),
+    }
+
+    // Ok(rune_result)
```
raphjaph, line 886:
```suggestion
```

raphjaph (changes requested):
Looks good so far.

Try etching multiple runes and check that the order is correct

raphjaph (approved):
Great PR!

PR #3441, opened 2024-04-02
Render closed-interval-ranges for human legibility.

arik-so:
Currently, a satoshi range of [10; 12], which contains the elements {10, 11, 12}, would be rendered as 10-13. It's convenient for calculating its value, but highly misleading because it suggests that 13 is part of the interval, which it isn't. 

The issue stems from the asymmetry of interval openness: the beginning of the interval is closed, and the ending is open.

To facilitate human legibility and convenience, the ends of intervals are now rendered as closed, too.

This PR might be controversial, so feel free to close.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -610,6 +610,13 @@ impl Server {
           .ok_or_not_found(|| format!("output {outpoint}"))?
       };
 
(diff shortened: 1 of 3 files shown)


raphjaph:
I like this. How about we just render it in the front-end as `[10, 13)` instead of `10-13`?

casey:
This is stale, so closing.

casey:
Fixing in #4075.

Issue #3396, opened 2024-03-28
should `inscriptions` be optional in batchfile?


rot13maxi:
Right now a batch file [requires inscriptions ](https://github.com/ordinals/ord/blob/6f7653e6213708791ca072d8c5eeaa0d6e2bdede/src/wallet/batch/file.rs#L21-L24). Since the interface for etching runes is through the `inscribe` command, should we allow this to be optional if you only want to etch runes and are not trying to inscribe something? I'm happy to write the PR, want to make sure there's not an intent here that I'm missing before I go make all the updates

raphjaph:
I think this would be a great addition!

Since we use the inscription envelope for the commit/reveal of the rune name. It would have to create an inscription with no content. And then on the runes page, if the inscription has no content, it shouldn't be shown

casey:
I think if we do this, the `inscriptions` field should still be present, but marked somehow as having no content, since creating a rune will still create an inscription, to store the commitment.

GiantDole:
+1 on this feature :)

0xprames:
+1

raphjaph:
Closed by #3451

gus4rs:
@raphjaph ```inscriptions``` are still required in the batch file

```
error: missing field `inscriptions` at line 23 column 1
```

raphjaph:
you can do:

```
inscriptions:
-
```

It's a bit janky for now but works.

gus4rs:
It creates an inscription s 

> you can do:
> 
> ```
> inscriptions:
> -
> ```
> 
> It's a bit janky for now but works.

It actually creates an inscription without a file along with the etchin, so no, it doesn't work

ensfn:
> you can do:
> 
> ```
> inscriptions:
> -
> ```
> 
> It's a bit janky for now but works.

@raphjaph It doesn't work, still creates inscription with outrageous fees (at least for me for some reason). Can you provide guidance on how can I remove validations for that part of the code so I can do it at my own risk?

Thanks

PR #3451, opened 2024-04-02
Allow inscribing delegate without file

### LINKED ISSUE #3445: When etching a rune, batch requires a file even when using a delegate.
When etching a rune, batch requires a file even when using a delegate.  I had thought this bug was fixed in previous versions but maybe have been reintroduced with runes?  Using delegates with runes icon inscriptions means we can save on fees by inscribing our images long before the halvening and the delegating to them later, this feature mostly works but may confuse some people trying to delegate in this way for runes.
Related:
https://github.com/ordinals/ord/pull/3027/files/7fc9271a5cf9190d3e20b7986932fab084869a67

raphjaph:
Fixes #3445, #3396

### DIFF
--- a/src/subcommand/wallet/inscribe.rs
+++ b/src/subcommand/wallet/inscribe.rs
@@ -1,6 +1,7 @@
 use super::*;
 
 #[derive(Debug, Parser)]
+#[clap(group(ArgGroup::new("input").required(true).args(&["delegate", "file"])))]
 pub(crate) struct Inscribe {
   #[command(flatten)]
   shared: SharedArgs,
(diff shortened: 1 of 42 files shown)


src/subcommand/wallet/inscribe.rs
```diff
@@ -1,6 +1,7 @@
 use super::*;
 
 #[derive(Debug, Parser)]
+#[clap(group(ArgGroup::new("input").required(true).multiple(true).args(&["delegate", "file"])))]
```
casey, line 4: Let's put this on multiple lines

src/subcommand/wallet/inscribe.rs
```diff
@@ -15,7 +16,7 @@ pub(crate) struct Inscribe {
   #[arg(long, help = "Send inscription to <DESTINATION>.")]
   pub(crate) destination: Option<Address<NetworkUnchecked>>,
   #[arg(long, help = "Inscribe sat with contents of <FILE>.")]
```
casey, line 18: Maybe add a comment here, like, may be omitted if `--delegate`…

casey (changes requested):
Test that it's an error to inscribe without a delgate or file, and also a test that it's allowed with the batch command.

casey (approved):
LGTM

PR #3461, opened 2024-04-04
Test validity of all opcodes

casey:
Partially for documentation purposes, I thought it would be good to include a test which explicitly tests all 256 opcodes for validity.

Opcodes 0 through 78 inclusive are data push opcodes, and can be used in runstones. Opcodes 79 and greater are non pushdata opcodes, and will result in a cenotaph if used in a runestone.

### DIFF
--- a/crates/ordinals/src/runestone.rs
+++ b/crates/ordinals/src/runestone.rs
@@ -2160,19 +2160,18 @@ mod tests {
   #[test]
   fn all_non_pushdata_opcodes_are_invalid() {
     for i in 79..=u8::MAX {
-      let mut script_pubkey = Vec::new();
-
-      script_pubkey.push(opcodes::all::OP_RETURN.to_u8());
-      script_pubkey.push(Runestone::MAGIC_NUMBER.to_u8());
-      script_pubkey.push(i);
-
       assert_eq!(
         Runestone::decipher(&Transaction {
           version: 2,
           lock_time: LockTime::ZERO,
           input: default(),
           output: vec![TxOut {
-            script_pubkey: script_pubkey.into(),
+            script_pubkey: vec![
+              opcodes::all::OP_RETURN.to_u8(),
+              Runestone::MAGIC_NUMBER.to_u8(),
+              i
+            ]
+            .into(),
             value: 0,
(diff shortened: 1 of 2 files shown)


crates/ordinals/src/runestone.rs
```diff
@@ -2094,4 +2094,84 @@ mod tests {
...
+      match i {
+        0..=75 => {
+          for j in 0..i {
+            script_pubkey.push(if j % 2 == 0 { 1 } else { 0 });
+          }
+
+          if i % 2 == 1 {
+            script_pubkey.push(1);
+            script_pubkey.push(1);
+          }
```
summraznboi, line 2116: This is how I implemented this unit test in Typescript:

```ts
for (const j of _.range(0, i)) {
          scriptPubKeyBuffers.push(Buffer.from([j % 2]));
        }

        if (i >= 2) {
          const additionalIntegersCount = (77 - i) % 4;
          const additionalIntegers = _.reverse(
            _.range(0, additionalIntegersCount).map((i) => i % 2)
          );
          scriptPubKeyBuffers.push(Buffer.from([additionalIntegersCount, ...additionalIntegers]));
        }
```

crates/ordinals/src/runestone.rs
```diff
@@ -2094,4 +2094,84 @@ mod tests {
...
+          },],
+        })
+        .unwrap(),
+        Artifact::Runestone(Runestone::default()),
+      );
+    }
+  }
+
+  #[test]
+  fn all_non_pushdata_opcodes_are_invalid() {
```
summraznboi, line 2151: For demonstration purposes, I would specifically test OP_PUSHNUM in contexts that would otherwise be correct if replaced with its equivalent OP_PUSHBYTES, ideally a single edict whose amount is controlled by the OP_PUSHNUM opcode.

raphjaph (approved):

crates/ordinals/src/runestone.rs
casey: I'm hesitant to add such a test, just because I know that the code doesn't care where the opcode is in the runestone.

PR #3476, opened 2024-04-07
record the final mint transaction

robinsdan:
It is not easy to check an onchain transaction mint successfully, because of the mint cap term. Saving the final mint transaction id, we can ensure all transactions before that tranaction id satisfy the cap condition.

### DIFF
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -49,6 +49,7 @@ pub struct RuneEntry {
   pub mints: u128,
   pub number: u64,
   pub premine: u128,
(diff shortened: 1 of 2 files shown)


casey:
Users can see if a mint succeeded by looking up the output on `/output/TXID:VOUT`.

robinsdan:
@casey I think the output is only temporary available? When an output is spent, it gets deleted from the database.
https://github.com/ordinals/ord/blob/744a3d6bec4422b2b1d599f0fa98bf69df04e4ad/src/index/updater/rune_updater.rs#L419-L421

PR #3499, opened 2024-04-10
chore: ord env add index log

bingryan:
## Description

`ord env` add  update log

## Screenshot

log


raphjaph:
Thanks for the PR but I don't think we need those updates in the console. If you want to see the blocks indexed you can just have a look at the explorer which is listening on port 9000. 

I could see a case for printing the `RUST_LOG=info` output in the `ord env` command. That could be made configurable.

PR #3500, opened 2024-04-10
Better error message when bitcoind doesn't start

twosatsmaxi:
Problem:
- if `bitcoind` isn't there, the command `ord env` or `cargo run env` fails with an error message - `No such file or directory`
image

Now it shows the reason:
image

### DIFF
--- a/src/subcommand/env.rs
+++ b/src/subcommand/env.rs
@@ -67,7 +67,8 @@ rpcport={bitcoind_port}
       Command::new("bitcoind")
         .arg(format!("-conf={}", absolute.join("bitcoin.conf").display()))
         .stdout(Stdio::null())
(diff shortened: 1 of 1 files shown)


src/subcommand/env.rs
```diff
@@ -67,7 +67,7 @@ rpcport={bitcoind_port}
       Command::new("bitcoind")
         .arg(format!("-conf={}", absolute.join("bitcoin.conf").display()))
         .stdout(Stdio::null())
-        .spawn()?,
+        .spawn().expect("failed to start bitcoind")?,
```
twosatsmaxi, line 70:
```suggestion
        .spawn().expect("failed to start bitcoind"),
```
lint fix

raphjaph (approved):

PR #3503, opened 2024-04-10
fixes inconsitency in terms.cap naming

### LINKED ISSUE #3502: Naming inconsistency in the error message
image
It shows terms.count but in etch yaml file, name is kept as `terms.cap`, I don't if there if any reason behind this or this looks like a small inconsistency in the error message.

twosatsmaxi:
Fixes #3502


raphjaph:
Ah sorry just saw this. I fixed this in #3505. Thanks for the PR though!

PR #3504, opened 2024-04-10
Lock runes commit output

raphjaph:


src/wallet/batch/plan.rs
```diff
@@ -128,6 +129,12 @@ impl Plan {
       .send_raw_transaction(&signed_commit_tx)?;
 
     if let Some(ref rune_info) = rune {
+      eprintln!("Locking rune commitment output");
```
casey, line 132: I think we can remove this

casey (approved):
LGTM modulo nits

PR #3505, opened 2024-04-10
Decimal balance

raphjaph:
The changes the output for `ord wallet balance`, so it shows the decimal amount instead of just the amount without the divisibiliity included


casey:
LGTM

PR #3506, opened 2024-04-10
Update data carriersize to match with ord

twosatsmaxi:


raphjaph (approved):
LGTM!

Any reason you opened this? Did you run into errors on regtest?

PR #3509, opened 2024-04-10
Add default content proxy and decompress to env command

jahvi:
## Description

Added `https://ordinals.com` as default content proxy to `env` command. I find myself using this all the time so I thought it'd be useful to have as the default.

Seems pretty safe since the proxy is only used if the regtest inscription doesn't exist but let me know if there's any downsides.


raphjaph:
I think we don't want to encourage people to by default using the content proxy. The way to do this would be to run `ord env` then killing it, going into the `env` directory and putting the content-proxy into the settings file found within. If you want add a line to the docs that explains.

jahvi:
@raphjaph Fair enough, do you mean the `ord.yaml` file? Because at the moment `content_proxy` isn't allowed as a setting so this was the easiest way I could find to add without changing things too much.

raphjaph:
Ah I see. Then let's make this an optional flag on `ord env` for now:

`ord env --content-proxy https://ordinals.com`

jahvi:
Is it worth having the same options as `server` available for `env`? For example `--decompress` would be useful as well.

jahvi:
Actually I think that'd be an overkill, I just updated this to include content-proxy and decompress to as args to `env` but let me know if that's not desirable.

raphjaph (approved):
LGTM

PR #3526, opened 2024-04-11
add --no-wait flag to not wait for etching to mature

rot13maxi:
if you're going to C-c anyway, you can instead pass the `--no-wait` flag when you do an etching. Makes it easier to wrap this in a script, if you're into such degeneracy.

not in love with the way this gets passed through, so marking it a draft for now.

### DIFF
--- a/src/wallet.rs
+++ b/src/wallet.rs
@@ -302,13 +302,14 @@ impl Wallet {
     commit: Transaction,
     reveal: Transaction,
     output: batch::Output,
+    no_wait: bool,
   ) -> Result<batch::Output> {
     eprintln!("Waiting for rune commitment {} to mature…", commit.txid());
 
     self.save_etching(rune, &commit, &reveal, output.clone())?;
 
     loop {
-      if SHUTTING_DOWN.load(atomic::Ordering::Relaxed) {
+      if SHUTTING_DOWN.load(atomic::Ordering::Relaxed) || no_wait {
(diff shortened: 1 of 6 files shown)


rot13maxi:
control-c

rot13maxi:
Control-c

src/wallet.rs
```diff
@@ -302,13 +302,14 @@ impl Wallet {
...
     output: batch::Output,
+    no_wait: bool,
   ) -> Result<batch::Output> {
     eprintln!("Waiting for rune commitment {} to mature…", commit.txid());
 
     self.save_etching(rune, &commit, &reveal, output.clone())?;
 
     loop {
-      if SHUTTING_DOWN.load(atomic::Ordering::Relaxed) {
+      if SHUTTING_DOWN.load(atomic::Ordering::Relaxed) || no_wait {
```
twosatsmaxi, line 312: why not instead have a separate message for `no_wait` instead of combining it with `suspending batch....` message, else branch saying "commitment submiited, reveal is pending" something along that line.

twosatsmaxi (changes requested):

casey (changes requested):
I don't think this needs to be in `ShareArgs`, since `ord wallet inscribe` never needs to wait.

raphjaph:
I'm going to close, please reopen if you want to continue work on this!

PR #3538, opened 2024-04-15
Supercharge Inscriptions with Runes

lifofifoX:
This PR adds a new recursive endpoint - `/r/inscription/:inscription_id/rune/:rune_id` - for fetching balance for a given rune token assigned to the inscription UTXO. This would allow combining inscriptions and runes in new ways, and allow artists to incorporate runes in their art.

Opening this PR primarily to get the conversation going. I am not fully convinced if this is necessarily the best approach. So, I am curious to hear more thoughts.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -236,6 +236,10 @@ impl Server {
           "/r/inscription/:inscription_id",
           get(Self::inscription_recursive),
         )
+        .route(
+          "/r/inscription/:inscription_id/rune/:rune_id",
+          get(Self::inscription_rune),
+        )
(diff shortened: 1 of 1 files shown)


raphjaph:
Yeah, we've been thinking along similar lines. Would be great to collect some concrete ideas here to see what people want to do with this.

elocremarc:
Another approach here by adding the `/r/output`
https://github.com/ordinals/ord/pull/3654

raphjaph:
Thanks for PR but I think if we do this we'll go for the `/r/output` method. It seems more general and cleaner.

PR #3556, opened 2024-04-16
Add test Rune cannot be minted less than limit amount

lugondev:
```rust
pub fn supply(&self) -> u128 {
    self.premine
      + self.mints
        * self
          .terms
          .and_then(|terms| terms.amount)
          .unwrap_or_default()
  }
  ``` 
Added a test case to validate the supply calculation logic.

### CODE REVIEW

File: src/runes.rs [hunk 1]
```diff
@@ -3909,6 +3909,99 @@ mod tests {
...
+
+    let txid1 = context.core.broadcast_tx(TransactionTemplate {
+      inputs: &[(2, 0, 0, Witness::new())],
+      op_return: Some(
+        Runestone {
+          mint: Some(id),
+          edicts: vec![Edict {
+            id,
+            amount: 111,
+            output: 0,
```
raphjaph, line 3963: Could you make it a two output transaction? That would make this a better test.

raphjaph (changes requested):

src/runes.rs
```diff
@@ -3909,6 +3909,100 @@ mod tests {
...
+        },
+      )],
+      [],
+    );
+
+    let txid1 = context.core.broadcast_tx(TransactionTemplate {
+      inputs: &[(2, 0, 0, Witness::new())],
+      outputs: 2,
+      op_return: Some(
+        Runestone {
```
lugondev, line 3959: is it right?

lugondev: @raphjaph  thanks

raphjaph (approved):

PR #3571, opened 2024-04-17
Revert runes balances json

lugondev:


src/subcommand/server.rs
```diff
@@ -740,6 +741,29 @@ impl Server {
     })
   }
 
+  async fn runes_balances(Extension(index): Extension<Arc<Index>>) -> ServerResult {
```
raphjaph, line 744: It should have the AcceptJson extension in the function arguments, so that it can be disabled with `--disable-json` flag.

raphjaph (changes requested):

src/subcommand/server.rs
raphjaph: And also it should return a 404 if the client does not set the JSON application header

lugondev:
@raphjaph done with testcase

casey (approved):
LGTM

raphjaph (approved):
LGTM

PR #3580, opened 2024-04-17
Show rune unlock height

### LINKED ISSUE #3524: Provide more info on rune unlock heights
1. Allow visiting `/block/N` for some future block height, and see the range of runes which will unlock on that block.
2. Allow visiting `/rune/X` for some future rune, and see when it will unlock.

raphjaph:
Fixes #3524


raphjaph:
Use binary search instead

casey (approved):
LGTM

PR #3602, opened 2024-04-18
Add pagination for runes and fix size query

lugondev:


raphjaph:
We've made the default 100 for DoS protection reasons. Specifying the query size would circumvent that protection. So unfortunately I cannot merge this. 

What's wrong with making multiple requests until you reach the amount you want?

lugondev:
> We've made the default 100 for DoS protection reasons. Specifying the query size would circumvent that protection. So unfortunately I cannot merge this.
> 
> What's wrong with making multiple requests until you reach the amount you want?

yeah,  we can disable page size for query to prevent DoS.
currently, I return wrong page size. 51 per page (now size is 50).
I will make default is 100.

raphjaph:
I don't think we should allow this since this circumvents the DoS protection. If you want more than 100 runes then it's best to just make multiple requests.

PR #3615, opened 2024-04-19
Allow higher rpcworkqueue limit conf

JeremyRubin:
It's really standard to configure bitcoin core to support e.g. 1000 inflight requests, enabling a configurable parameter can speed up the index updating substantially.

Some of the other parallelization in this area around block fetching is "stinky" and needs further refactor and configuration as well, but this is a low hanging fruit.


raphjaph:
> It's really standard to configure bitcoin core to support e.g. 1000 inflight requests, enabling a configurable parameter can speed up the index updating substantially.

Do you think we should increase the default? 


> Some of the other parallelization in this area around block fetching is "stinky" and needs further refactor and configuration as well, but this is a low hanging fruit.

Agreed, it is kind of haphazard. We haven't touched this part of the code in a long time. If you have some good refactor ideas I'd be happy to review them.

raphjaph (approved):
LGTM

PR #3635, opened 2024-04-20
Update sparrow-wallet.md --name flag update

taha-abbasi:
The --wallet flag no longer works, to import wallet with name we need --name flag instead with --from mnemonic flag instead of the current listed: `ord --wallet ord_from_sparrow wallet restore "BIP39 SEED PHRASE"`


raphjaph (approved):
Thanks!

PR #3644, opened 2024-04-21
change operator to <= for etching when start block height set

jnapp18:
change operator to <= for etching when start block height set so that mints can be fired off when the next block is the start of a rune mint. This allows users in order to get tx's in the mempool to be confirmed in the same block as the mint start

### DIFF
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -63,7 +63,7 @@ impl RuneEntry {
     };
 
     if let Some(start) = self.start() {
-      if height < start {
+      if height <= start {
         return Err(MintError::Start(start));
       }
     }


bFanek:
Thank you for creating this PR (beat me to it lol), I can confirm i missed a mint due to this bug on the ord codebase (0.18.3), and this PR fixes this. Cheers!

coinspiracynut:
I believe that you would have to account for start block -1

`if let Some(start) = self.start() {
    if height < start - 1 {
        return Err(MintError::Start(start));
    }
}`

to resolve this specific issue, although the optimal solution is a flag that allows the removal of block checks for power users

jnapp18:
> I believe that you would have to account for start block -1
> 
> `if let Some(start) = self.start() { if height < start - 1 { return Err(MintError::Start(start)); } }`
> 
> to resolve this specific issue, although the optimal solution is a flag that allows the removal of block checks for power users

you are right, let me fix

PR #3645, opened 2024-04-21
 Allow mint submissions as start block is pending

coinspiracynut:
Current implementation only allows mint submissions after the start block is confirmed meaning ord users can not attempt to compete in that block. This implementation allows for submissions at start - 1, which is a good solution for most users. However, a flag to ignore these checks for power users might be a more composable solution.

### DIFF
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -63,7 +63,7 @@ impl RuneEntry {
     };
 
     if let Some(start) = self.start() {
-      if height < start {
+      if height <= start {
         return Err(MintError::Start(start));
       }
     }
--- a/src/index/entry.rs
+++ b/src/index/entry.rs
@@ -63,9 +63,9 @@ impl RuneEntry {
     };
 
     if let Some(start) = self.start() {
-      if height <= start {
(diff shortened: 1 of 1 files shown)


0xcruci:
Referencing this issue for more context https://github.com/ordinals/ord/issues/3643

gmart7t2:
Isn't the end check similarly off by one?

If the minting ends at block 80 and block 79 is the current block, new mints won't be accepted in block 80 so don't let people mint after block 79 has been mined

utxo-detective:
`if height < (start - 1) {` can be simplified to `if height <= start {`

benbuschmann:
> Isn't the end check similarly off by one?
> 
> If the minting ends at block 80 and block 79 is the current block, new mints won't be accepted in block 80 so don't let people mint after block 79 has been mined

yes

raphjaph (changes requested):
The projection into the next block should be done on a higher level.

casey:
Fixed in #3659.

0xcruci:
Thanks @casey!

PR #3654, opened 2024-04-22
Recursive Runes Endpoint Output

elocremarc:
`/r/output/:output`

Its time for the output endpoint

image

Example of runes with an inscription  https://ordinals.com/output/b84a1dd3cfb9ebb26f240afe7f17bf73f8a1b6cc9774ea26e360fbe077a26669:0

I just proxied the current `/output` response to play around with it. What do we want to keep or remove from the fields? We want  at least `runes` `inscriptions`. This allows artist to choose if they want to activate an inscription via a rune or via an inscription on the output. 

Another cool thing to have would be sat ranges. Then you can start activating inscriptions based on sats in the output if you don't want to activate an inscription with runes or inscriptions. Output gives you all 3 options activating inscriptions with sats, inscriptions or runes.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -250,6 +250,10 @@ impl Server {
           "/r/sat/:sat_number/at/:index",
           get(Self::sat_inscription_at_index),
         )
+        .route(
+          "/r/output/:output",
+          get(Self::output_recursive),
+        )
         .route("/range/:start/:end", get(Self::range))
         .route("/rare.txt", get(Self::rare_txt))
         .route("/rune/:rune", get(Self::rune))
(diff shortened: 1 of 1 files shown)


PirateCodeWeb3:
this is amazing, LFG

devords:
One of the reasons I had used `/r/inscription/:inscription_id/rune/:rune_id` in #3538 over output was so that the recursive endpoints would not be ephemeral. Output endpoints would probably also need to be paginated, scoping by `rune_id` avoids that.

alexamizz:
I was able to produce the output with both an inscription and rune. However, it looks like the inscription and rune is now soulbound. It will not let me separate.  I was thinking this could be a cool solution for upcoming projects to include runes with the inscriptions when someone goes to "mint". Kinda like a package deal.  

https://signet.ordstuff.info/output/e00ef72ec8aea33553cce8b41bfc0c7cbf7044563c24b02d11b21eadfc6da538:0
image

raphjaph:
We have the `/r/utxo/<OUTPOINT>` now. Have a look at #4148.

PR #3656, opened 2024-04-22
Do not show runic outputs in cardinals command

### LINKED ISSUE #3648: `wallet cardinals` should not return runic utxos
As of right now `ord wallet cardinals` returns runic utxos.

This is an unexpected behaviour, as spending those utxos results in spent rune balances.

`wallet cardinals` should return only common non-rune, non-inscription utxos

raphjaph:
Fixes #3648


casey (approved):
LGTM

PR #3673, opened 2024-04-24
show progress for etching

### LINKED ISSUE #3548: Display commitment maturation progress
Probably with some fire emojis for each block.

### LINKED ISSUE #3674: ord wallet resume shouldn't bail fully when there is more than one pending etchings..
# Scenario
## Expected
- It should skip the one for which commitment is spent.
- Resume other pending etchings.

twosatsmaxi:
Fixes #3548 
Fixes #3674 

image

Now it shows a progress bar too:
image

Fancier
image


twosatsmaxi:
@casey @raphjaph: can someone review it?

alpeshvas:
@raphjaph ^

raphjaph (approved):
This looks good to me. 

Just made some simplifying changes. I'll let @casey look over it tomorrow before we merge.

alpeshvas:
@casey any update ^

src/wallet.rs
```diff
@@ -280,37 +289,42 @@ impl Wallet {
...
 
-    if let Some(transaction) = transaction {
-      if u32::try_from(transaction.info.confirmations).unwrap() + 1
-        >= Runestone::COMMIT_CONFIRMATIONS.into()
-        && rune
-          >= Rune::minimum_at_height(
-            self.chain().network(),
-            Height(u32::try_from(self.bitcoin_client().get_block_count()? + 1).unwrap()),
-          )
+  pub(crate) fn is_mature(&self, rune: Rune, commit: &Transaction) -> Result<Maturity> {
```
twosatsmaxi, line 300: Should we change the method name too here @raphjaph ? Now that it doesn't return bool, `is_mature` is less suitable here. We can name it `check_maturity` or something to reflect enum return type.

raphjaph: good point, will change

src/subcommand/wallet/resume.rs
```diff
@@ -27,8 +27,13 @@ impl Resume {
...
 
-        if wallet.is_mature(rune, &entry.commit)? {
-          etchings.push(wallet.send_etching(rune, &entry)?);
+        match wallet.check_maturity(*rune, &entry.commit)? {
+          Maturity::Mature => etchings.push(wallet.send_etching(*rune, entry)?),
+          Maturity::CommitSpent(txid) => {
+            eprintln!("Commitment for rune etching {rune} spent in {txid}");
+            etchings.remove(i);
+          }
+          _ => continue,
```
casey, line 36: I think it would be nice to explicitly enumerate all the cases, instead of using `_`, so that it's obvious that it's okay to ignore the other cases.

src/wallet.rs
```diff
@@ -280,37 +289,42 @@ impl Wallet {
     self.settings.integration_test()
   }
 
-  pub(crate) fn is_mature(&self, rune: Rune, commit: &Transaction) -> Result<bool> {
-    let transaction = self
-      .bitcoin_client()
-      .get_transaction(&commit.txid(), Some(true))
-      .into_option()?;
+  fn is_after_minimum_height(&self, rune: Rune) -> bool {
```
casey, line 292:
```suggestion
  fn is_above_minimum_at_height(&self, rune: Rune) -> bool {
```

src/wallet.rs
```diff
@@ -280,37 +289,42 @@ impl Wallet {
...
+    rune
+      >= Rune::minimum_at_height(
+        self.chain().network(),
+        Height(u32::try_from(self.bitcoin_client().get_block_count().unwrap() + 1).unwrap()),
```
casey, line 296: This shouldn't unrwap:

```suggestion
        Height(u32::try_from(self.bitcoin_client().get_block_count()? + 1).unwrap()),
```

src/wallet.rs
```diff
@@ -280,37 +289,42 @@ impl Wallet {
...
+    Ok(
+      if let Some(commit_tx) = self
+        .bitcoin_client()
+        .get_transaction(&commit.txid(), Some(true))
+        .into_option()?
       {
-        let tx_out = self
-          .bitcoin_client()
-          .get_tx_out(&commit.txid(), 0, Some(true))?;
+        let current_confirmations = u16::try_from(commit_tx.info.confirmations).unwrap();
```
casey, line 307: This unwrap could panic if we're resuming after more than 2^16 blocks.

src/wallet.rs
```diff
@@ -324,14 +338,37 @@ impl Wallet {
       entry.commit.txid()
     );
 
+    let mut pending_confirmations = Runestone::COMMIT_CONFIRMATIONS;
+
+    let pb = ProgressBar::new(pending_confirmations.into()).with_style(
```
casey, line 343:
```suggestion
    let progress = ProgressBar::new(pending_confirmations.into()).with_style(
```

casey (changes requested):

src/subcommand/wallet/resume.rs
twosatsmaxi: makes sense, now that I have added them, just realized that we should also clear etching if commit is not found (for CommitNotFound arm)?

twosatsmaxi: done.

src/wallet.rs
twosatsmaxi: done.

src/wallet.rs
twosatsmaxi: done.

src/wallet.rs
twosatsmaxi: done.

raphjaph (approved):
LGTM

To remove an etching from the loop we need to use the clear_etching function. With that I was able to remove the enumerate and i in the loop. If the commit is not found I'd rather not delete the entry for now, since maybe you would want to recover that. We can change it back later but this way we don't accidentally delete something.

twosatsmaxi:
> LGTM
> 
> To remove an etching from the loop we need to use the clear_etching function. With that I was able to remove the enumerate and i in the loop. If the commit is not found I'd rather not delete the entry for now, since maybe you would want to recover that. We can change it back later but this way we don't accidentally delete something.

alright.

Issue #3640, opened 2024-04-21
Feature request: command to burn runes


kusan-thana:
Would be nice to have a burn command that would either send the runes to a special address or that would make a cenotaph to burn them. The number of burned runes would be displayed on the rune page.

pauldemarco:
For all the shameful pre-miners, this could allow for a redemption arc if some / all of the premine can be burned -- and subsequently show a lesser percentage of premine.

This perhaps should be a separate issue: "Premine percentage should account for burned runes"

nghuyenthevinh2000:
I think this could be fairly easily done by:
* sending runes to a recognized burn address X by Rune protocol
* try to create Cenotaph to make sure it stays burned. Cenotaphs may be created if a runestone contains an edict with an output number greater than the number of inputs

Hmm, I will try to produce a working logic for this one if you don't mind @raphjaph

PR #3683, opened 2024-04-25
feat: burn rune through forcing Cenotaph

nghuyenthevinh2000:
fixes: #3640 

add a command to burn all specific runes in a wallet

**Logic**
1. Add Burn impl in subcommand/wallet: forcing burn through Cenotaph EdictOutput Flaw

**Test**
1. Add a happy case test burning_rune_works(): 
* send 100 rune to address B
* burn rune in address A (which contains 900 rune), confirming that balances only return 100 rune
* confirm that rune info return burned amount of 900

I think it can be fine - grained further to burn exact amount in a wallet through protocol recognized burn addresses. However, changes to updater are required to recognize the burn addresses. Changes to updater can be a lenghtly review process as it touches the core, so probably in another PR. Which burn addresses can be a topic for debates @raphjaph 

However, in this PR, I still include burn address constants for reviews
```
// satoshi first block address
 pub const MAINNET_BURN_ADDRESS: &'static str = "1FvzCLoTPGANNjWoUo6jUGuAG3wg1w4YjR";
 // just a random testnet address
 pub const TESTNET_BURN_ADDRESS: &'static str = "tb1prqr6lpxeqcyfff8n7c748ad7szyh8hwmf9nly598r5uqewkm4kms0nevay";
 // just a random regnet address
 pub const REGNET_BURN_ADDRESS: &'static str = "bcrt1pyf99p0qrt8dsx8a8hnd7edzpnnpsndf9ce6hhx55s9f9v84e48eqdlyd6n";
```


onchainguy-eth:
I implemented a generic burn in #3437. Can extend this to support runes?

raphjaph:
I'll have a look at generic burning first (#3437) but probably we don't want to do protocol level burning through cenotaphs but rather by sending to an `OP_RETURN`.

raphjaph:
I merged #3437, since burning through OP_RETURN is more generic and can be used for inscriptions, sats and runes.

gus4rs:
@raphjaph After #3437 is it possible to burn runes via ```ord wallet```? I am thinking about two scenarios:
* Burn a certain amount of a Rune
* Burn the "etching"

onchainguy-btc:
Happy to work on this as well @raphjaph

raphjaph:
> @raphjaph After #3437 is it possible to burn runes via `ord wallet`? I am thinking about two scenarios:
> 
> * Burn a certain amount of a Rune
> * Burn the "etching"

Yes, burning a rune should be part of `ord wallet burn` command as well. It's already defined on the protocol level here: https://docs.ordinals.com/runes.html?highlight=burn#burning. So all we have to do is implement it. You should be able to specify the amount you want to burn.

For "etching" do you mean the optional parent inscription of the rune? This can already be done with the new command. @onchainguy-btc is tweaking it so it only burns a single sat instead of the whole utxo it is in.

gus4rs:
> For "etching" do you mean the optional parent inscription of the rune

Also this, but more destroying the rune itself, so it is totally trashed, no one can mint/transfer it anymore, would it make sense?

raphjaph:
An etching can't retroactively destroy all runes it "minted". Once there on Bitcoin they're no longer owned by the original creator

gus4rs:
> An etching can't retroactively destroy all runes it "minted"

The idea was not to destroy the already minted runes, but the etching itself preventing new mints

gus4rs:
So once an etching is confirmed, it can't prevent new mints, or can't be modified in any shape or form, or even be "transferred" to another owner?

PR #3685, opened 2024-04-25
add github templates for issues and prs for easier tracking

nghuyenthevinh2000:
I notice that ord is receiving a lot of community contributions, questions, thus a need for a template to easily follow and manage. Things are definitely going wild after runes launch 🤣

This pr adds:
1. Issue templates
* bug report
* documentation request
* feature request

2. pull request template

it will look like so:
Screenshot 2024-04-25 at 15 11 30
Screenshot 2024-04-25 at 15 24 29


raphjaph:
Personally I'm not a big fan of issue templates since they add a lot of friction and boilerplate when I want to add an issue. I'm open to having my mind changes though. Do you maybe have an example repo where these kind of templates are used? That way I can get a feeling for how it is used in practice.

nghuyenthevinh2000:
let's reason from the maintainer perspective, 

Currently Ord has 390 issues, a huge number and will continue to increase. Some of it may contain critical bugs but poor issue description, some may contain a brilliant feature idea but unattractive issue name. As a maintainer, how to quickly scan all these issues and maximize the chance of discovery, also spending less time reading all these strange looking issues? 

With issue templates, people now have a framework to describe what they want (ofc some will not follow but you can refer them to a template that they need to follow, so now they have to spend time actually describing their problems seriously). Plus, it will auto - assign a tag upon issue creation. Now, you lose less time scanning which are "bug", "documentation", "features". Upon clicking one, you can already see a clearly defined problem (with environment, ord version, inputs, ...) instead of just a name and some summaries.

With pull request templates, people now have to clearly define what they want to submit, and relevant test before your review.

I understand the pain of maintaining an open source project (here is mine: https://github.com/classic-terra/core), looking through all poorly defined issues, PRs can be real crazy sometimes. So, these templates help your sanity. All of these are based on cosmos-sdk: https://github.com/cosmos/cosmos-sdk, they have really good system of github actions to clean issues, prs, auto tagging exactly which components are affected based on changed codes. All of these is to help save time of maintainers.

PR #3694, opened 2024-04-26
Patch some omissions in the Chinese translation

shadowv0vshadow:
1. Fix `OOP_RETURN ` to `OP_RETURN `
2. Add some translcation for chinese


raphjaph (approved):
Thanks, looks good to me

PR #3703, opened 2024-04-28
Add Taproot Assets Bridge (WIP)

linden:
Hi @casey & @raphjaph.

I'm writing this PR to formally propose adding bridge support to the Runes protocol, starting with Taproot Assets and potentially expanding to Ethereum, Solana and others to make adoption easy (and bi-directional, unlike teleburning alleviating a lot of risk).

## Why Taproot Assets?
1. Lightning Support: Taproot Assets is set to provide lightning support in the coming months (potentially weeks), which will bring near-instant FT trading and transfers. Lightning support in Taproot Assets has taken a fairly large and well-funded team quite a while (6+ months) to implement; while Runes could implement lightning support independently I don't see a good reason to repeat this work.
2. Batch Support: While Runes does support batching from a single-party, Taproot Assets will support batching from multiple parties into a single transaction on-chain, trustlessly. Adopting this system would alleviate a lot of the congestion caused by trading.

## How?
This PR adds a new Runestone called the Bridge Runestone which contains an output (or input), protocol and wether or not the bridge is a lock or unlock.

### Lock (aka bridge out, Runes -> Taproot Assets).
1. Runes: a transaction is funded, similar to a send but with a Bridge Runestone instead.
2. Taproot Assets: a mint is created inside this same transaction. The mint contains an asset, amount and metadata matching the Rune.
3. Transaction is broadcasted.
4. Taproot Assets: once confirmed, the proof is shared with a universe.

### Unlock (aka bridge in, Taproot Assets -> Runes).
1. Runes: a transaction is with a Bridge Runestone and an input containing the sats that'll become bridged Runes.
2. Taproot Assets: a burn is created inside the same transaction.
3. Transaction is broadcasted.
4. Taproot Assets: once confirmed, the proof is shared with a universe.

### Index (Runes <-> Taproot Assets):
1. Runes: a Bridge Runestone for a lock and then later a unlock is found.
2. Taproot Assets: we validate the proofs from the lock to the unlock to ensure the burn is valid. 
3. Runes: the sats from the input are indexed as the Rune.

I've currently implemented a MVP for lock-in functionality and would appreciate any feedback. I'd be happy to jump on a call and discuss the implementation further.

I'll pause implementation for now and continue as soon as the high-level overview is given the OK.

## Early Adopters
I'm the creator of [Joltz Wallet](https://wallet.joltz.app/), the first non-custodial Taproot Assets wallet outside of tapd. I'm also doing this work in collaboration with Nostr Assets, the largest exchange on Taproot Assets. We'd both love to implement this as we think it'll benefit both ecosystems greatly.


casey:
Thanks for the PR!

I think that adding support for taproot assets to ord represents a level of complexity that I'm not comfortable with.

Specifically, adding support for verifying that an incoming taproot asset is a valid taproot asset requires adding support to ord for understanding the entire taproot assets protocol. So, a bridge out might be okay, but it would be a one way bridge, since we would never add support for verifying that an incoming asset was valid. And if bridges are one-way only, they can be done with a burn mechanism, which sends the runes to an OP_RETURN containing details of the destination taproot assets, which doesn't need any specific support from ord.

If for some reason, there's massive demand for this feature, than I would consider adding it, but I suspect that in the long run either taproot assets will be very successful, and nobody will care about runes, or runes will be very successful and nobody will care about taproot assets, and a bridge between the two would ultimately be a bridge to nowhere.

linden:
Hi @casey, thanks for the feedback.

> Specifically, adding support for verifying that an incoming taproot asset is a valid taproot asset requires adding support to ord for understanding the entire taproot assets protocol.

That's understandable, it does add to the complexity of indexing which was my primary concern with this approach. 

>  [...] suspect that in the long run either taproot assets will be very successful, and nobody will care about runes, or runes will be very successful and nobody will care about taproot assets [...]

I agree in the long term one protocol will probably become the defacto protocol but in the short term bridging can allow users to share the benefits of each meta-protocol without lock-in, benefitting the community as whole. Bitcoin vs. World instead of Runes vs. Taproot Assets vs. meta-protocol.

> Edit: Or, of course, third possibility, nobody cares about either 😅

I hope not haha. let's work to make sure that doesn't happen with undeniable value (i.e. instant trading via lightning).

I've got an idea on how this could be implemented in a way that takes the strain off of the ord indexer, relying more heavily on tapscript for verification instead of Taproot Assets itself - very roughly equivalent to an HTLC. I'd initially not gone with this approach as it was a lot more conceptually complex.

I'm happy to write up this approach over the next few days if you think this is something Runes would be open to adopting, assuming A. ord doesn't need to depend on Taproot Assets and B. complexity for the indexer is minimal.

casey:
How would ord verify that a taproot asset was present in an input without having to implement the full taproot assets protocol?

PR #3710, opened 2024-04-30
docs: Translation of Runes into Portuguese

Dione-b:
@raphjaph 

You could do a little review. I am a rune protocol and ordinals enthusiast and would like to share the knowledge in my native language. Thank you for your attention.


Dione-b:
> Thanks for the translation!
> 
> Unfortunately I can't merge this because it does not conform to our translation template. Please have a look at it here: https://github.com/ordinals/ord#translations. Then you can open a new PR with the correct template.

Thanks for the tip. I will redo it following the specified rules

PR #3722, opened 2024-05-03
FIXED lazy load iframe

losingle:
1. Will not cause the browser to crash in a poor network environment
2. Load iframe according to window

But it is not the best solution at the moment. I will modify it later to try to avoid using iframes to improve performance and security.


casey:
We add `loading="lazy"` to iframes. Does that not work? I'm not sure I'm a fan of this approach, since it breaks the explorer entirely if JavaScript is not enabled.

losingle:
> We add `loading="lazy"` to iframes. Does that not work? I'm not sure I'm a fan of this approach, since it breaks the explorer entirely if JavaScript is not enabled.

The loading="lazy" of iframe is not efficient,
It will only load the iframe in the user window,  [isIntersecting](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/isIntersecting)
This solution mainly improves the front-end experience. 
I have not changed the logic of inscription, just home.html、blocks.html、inscriptions.html

casey:
I don't understand what the improvement is. It seems okay if it only loads inscriptions when they're in view, and I'd rather not break preview if JavaScript is disabled. Thank you for the PR though!

Issue #3620, opened 2024-04-19
A useless postage was added to the sending runes tx?


ZhiGuang:
Query Balance:

PS D:\Projects\bitcoin\ord> ordtest wallet balance
` 
{
  "cardinal": 46553843,
  "ordinal": 20000,
  "runes": {
    "xxxxxxx": "1000000",
    "MEME•OF•RUNES": "24",
    "xxxxxxx": "381000000",
    "xxxxxxx": "20000"
  },
  "runic": 1450000,
  "total": 48023843
}
`

send runes: 
PS D:\Projects\bitcoin\ord> ordtest wallet send --fee-rate 100 tb1pt0v7eye37l38e50lnywfu6atj2ec6dzteduqa4xasqnn6ndckcuqx83fsn 24:MEME•OF•RUNES
`
{
  "txid": "07537f2643f790535e12a2211f78ec82e2cfef866b694ba6d653a2caecf592f9",
  "psbt": "xxxx",
  "outgoing": "24:MEME•OF•RUNES",
  "fee": 27600
}
`

The payload in OP_RETURN:
`
0099ee9d0110e01202
`

Dechiper result:
`
{"edicts":[{"id":{"block":"2586393","tx":"16"},"amount":"2400","output":"2"}]}
`

[Tx in mempool](https://mempool.space/zh/testnet/tx/07537f2643f790535e12a2211f78ec82e2cfef866b694ba6d653a2caecf592f9)

This transaction only requires one postage, why are there two postage outputs in the transaction?

gmart7t2:
I have the same problem. 

When you send runes, ord always creates 3 outputs: 1 for the op_return, one for the change runes, and one for the outgoing runes.

If you are sending all the runes from the input utxos, there is no need for a change output but it is created anyway.

gmart7t2:
There should be a check for the case where you are sending the whole runic balance of the input utxos, because in that case no change or op_return is required. We can just send the inputs to the output and rely on the fact that the lack of an edict makes all the input runes go to the first output.

Issue #3025, opened 2024-01-16
Sacrifices


devords:
Allow burning/sacrificing inscriptions as an offering to another inscription.

I can think of two ways to do this:

- Generate a burn address specific to the inscription using the inscription ID, and send sacrificial inscriptions there
- Spend sacrificial inscription to `op_return` and include inscription ID in the data. This was what Raph/Casey suggested on the discord coding club.

This, combined with recursive endpoints for fetching sacrifices, will allow for permission-less state updates and open up other possibilities.

onchainguy-eth:
[This](https://github.com/ordinals/ord/pull/3437) could be a good start for getting this feature mid-term.

gmart7t2:
> I can think of two ways to do this

The 2nd is better, because nodes know what op_return outputs are unspendable. Sending to a (not provably unspendable) burn address bloats the UTXO set

cbspears:
I love this idea, I have also been of the opinion that the best way to disassociate an inscription with a sat is to spend it to opreturn.

However, we should consider the maturity of the overall marketplace landscape for ordinals _and_ the reality that most major marketplaces have not integrated other important ordinals consensus changes into their platforms. Adding this disruptive consensus change to ordinals could result in disassociating an inscription with a sat by ord clients which is not recognized by major marketplaces such as MagicEden.

I think this change could be very disruptive to the overall market if it includes inscription-sat-disassociation.

lifofifoX:
@cbspears This won't disassociate inscription from sats. It'd just be special kind of burning. It came up today in the conversation re: on-chain gaming, and how something like this could be used to perform a game action against another player.

cbspears:
Ah! Then my comment is irrelevant.

casey:
I thought about this, and I think the most useful thing is actually displaying the amount of bitcoin sacrificed. I was going to hijack this issue but I would just wind up completely rewriting the OP, so I created #4268 instead.

PR #3728, opened 2024-05-05
Refactoring send.rs by introducing transfer trait

twosatsmaxi:
#### Catching bugs early for example: 
- If transfer trait has a fixed contract, bugs like this could be cought early by making sure that the postage is used for new transfer implementation - https://github.com/ordinals/ord/issues/3620
  
#### Code Quality
- More extensibility, for example, the most requested features in ordicord like `splitting rune` or `send to many` can have its own transfer trait implementation without bloating [send.rs](https://github.com/ordinals/ord/blob/master/src/subcommand/wallet/send.rs) file.
- Cleaner code, for example:
    - [`sending_inscription`](https://github.com/ordinals/ord/blob/b0435e8cc5718773e2b97c01a25a96dc5af01abe/src/subcommand/wallet/send.rs#L167) boolean can be moved out of the `create_unsigned_send_satpoint_transaction` method and to its trait implementation (not a big deal but just a bit cleaner code)
    - [send.rs](https://github.com/ordinals/ord/blob/9c903e43bc8388efb30b6ff4ce439f1325365c39/src/subcommand/wallet/send.rs) is now just 52 lines of code after the refactor and super readable ❤️

### DIFF
--- a/src/subcommand/wallet/send.rs
+++ b/src/subcommand/wallet/send.rs
@@ -1,4 +1,10 @@
-use {super::*, crate::outgoing::Outgoing, base64::Engine, bitcoin::psbt::Psbt};
+use {super::*, crate::outgoing::Outgoing};
+use crate::wallet::transfer::amount_tranfer::AmountTransfer;
+use crate::wallet::transfer::inscription_transfer::InscriptionTransfer;
+use crate::wallet::transfer::rune_transfer::RuneTransfer;
+use crate::wallet::transfer::sat_transfer::SatTransfer;
+use crate::wallet::transfer::satpoint_transfer::SatPointTransfer;
+use crate::wallet::transfer::Transfer;
 
 #[derive(Debug, Parser)]
 pub(crate) struct Send {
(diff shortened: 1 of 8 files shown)


onchainguy-eth:
Maybe we can merge #3437 before? 👀 It allows for different features to be built on top like #3025 or #3640

twosatsmaxi:
@raphjaph @casey this ^ is ready for review now, no functionality changes and just refactoring.

casey:
I personally don't think this is a huge increase in readability, since there's just more indirection in figuring out what's going on.

twosatsmaxi:
> I personally don't think this is a huge increase in readability.

But do you think though that it's a bit more readable?

twosatsmaxi:
> since there's just more indirection in figuring out what's going on.
It simplify things a bit for new folks who want to read about what's going on in send.rs.

@casey some more Context on why I went ahead with this refactor: I wanted to implement the even split command for rune where I can split 10k of runes into 100 each, send.rs was a big and going through that code with different branches was lots of to-and-fro. If it had been like this, understanding what's going through would be much less time consuming for new contributors and can start checking in more features like split command for example.

image

twosatsmaxi:
I feel that even if it adds a bit more readability and simplicity, there is no harm in checking in to main branch unless it adds more complexity, which I don't see.

twosatsmaxi:
We were a way past ["Rule Of Three"](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) and the refactoring was a due in my opinion. Currently, there are 5 sends.

casey:
To be clear: I closed this PR because I think it's a net negative and makes the code harder to read.

The rule of three is about deduplication.

The PR is net +107 lines of code, and replaces a match statement which calls three different functions, with a trait with five different implementations. I think it's much worse and more complicated.

casey:
> * If transfer trait has a fixed contract, bugs like this could be cought early by making sure that the postage is used for new transfer implementation - [A useless postage was added to the sending runes tx? #3620](https://github.com/ordinals/ord/issues/3620)

This bug was fixed with a change to `create_unsigned_send_runes_transaction`. In this PR, this function was just moved from one place to another. I don't see that would have helped find that bug.

> * More extensibility, for example, the most requested features in ordicord like `splitting rune` or `send to many` can have its own transfer trait implementation without bloating [send.rs](https://github.com/ordinals/ord/blob/master/src/subcommand/wallet/send.rs) file.

Having six functions in a file is fine.

>   * [send.rs](https://github.com/ordinals/ord/blob/9c903e43bc8388efb30b6ff4ce439f1325365c39/src/subcommand/wallet/send.rs) is now just 52 lines of code after the refactor and super readable ❤️

It's super readable, but reading it doesn't tell you anything about what it does.

twosatsmaxi:
> The PR is net +107 lines of code.

but does # of lines code really matter in such cases? # of lines of code would increases if a piece of code is moved to different files with new imports, new functions definitions, empty lines tends to add to net # of loc. 

> This bug (`the postage not being used for send runes`) was fixed with a change to create_unsigned_send_runes_transaction. In this PR, this function was just moved from one place to another. I don't see that would have helped find that bug.

It's true that It wouldn't have helped finding that bug, but while writing the "send rune" implementation itself, it would have enforced the author to use postage (due to trait contract for `transer` being there) In the absense of the transfer contract, there is no way to enforce/nudge that `postage` need to be used. The bug was later fixed by adding `postage` to `create_unsigned_send_runes_transaction` method [here](https://github.com/ordinals/ord/pull/3484/files#r1599636491)

PR #3679, opened 2024-04-24
Add --rune flag to resume command to enable resuming a specific etching

ldiego08:
Adds a new `--rune <RUNE>` flag to allow resuming a specific pending etching. This is useful in automated scenarios (if anyone else is into that kind of stuff) where you'd like to etch something, suspend (perhaps via `--no-wait` from #3526), do something else, and then resume after the 6 confirmations.

PR #3732, opened 2024-05-06
Add command to list pending etchings

ldiego08:
Add a new `ord wallet pending` command to list pending etchings, as mentioned by @raphjaph on #3679.


src/subcommand/wallet/pending.rs
```diff
@@ -0,0 +1,20 @@
...
+}
+#[derive(Debug, Parser)]
+pub(crate) struct Pending {}
+
+impl Pending {
+  pub(crate) fn run(self, wallet: Wallet) -> SubcommandResult {
+    let etchings = wallet
+      .pending_etchings()?
+      .into_iter()
+      .map(|(_, entry)| entry)
```
ldiego08, line 15: I'm just returning the entire etching entry, making the output quite verbose. Please, let me know if another output schema is more desirable.

src/subcommand/wallet/pending.rs
ldiego08, line 1: Not sure about the name as it doesn't specify it means pending _etchings._ Feel free to suggest a different name.

raphjaph: `resume` is also kind of ambiguous so I think this should be fine. We can always change it if we find a better name.

src/subcommand/wallet/pending.rs
```diff
@@ -0,0 +1,20 @@
+use {super::*, crate::wallet::entry::EtchingEntry};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub struct PendingOutput {
+  pub etchings: Vec<EtchingEntry>,
+}
```
raphjaph, line 6:
```suggestion
pub struct PendingOutput {
  pub rune: SpacedRune,
  pub commit: Txid,
}
```

Let's just return the Rune name and the commit Txid for now.

src/subcommand/wallet/pending.rs
```diff
@@ -0,0 +1,20 @@
...
+
+impl Pending {
+  pub(crate) fn run(self, wallet: Wallet) -> SubcommandResult {
+    let etchings = wallet
+      .pending_etchings()?
+      .into_iter()
+      .map(|(_, entry)| entry)
+      .collect();
+
+    Ok(Some(Box::new(PendingOutput { etchings }) as Box<dyn Output>))
```
raphjaph, line 18:
```suggestion
    Ok(Some(Box::new(PendingOutput { etchings }) as Box<dyn Output>))
```

Down here then return a Vec<PendingOutput>

raphjaph (changes requested):
Looks good so far. Definitely also needs an integration test. You could put that in the `tests/wallet/resume.rs` file.

src/subcommand/wallet/pending.rs
```diff
@@ -0,0 +1,28 @@
...
+#[derive(Debug, Parser)]
+pub(crate) struct Pending {}
+
+impl Pending {
+  pub(crate) fn run(self, wallet: Wallet) -> SubcommandResult {
+    let etchings = wallet
+      .pending_etchings()?
+      .into_iter()
+      .map(|(rune, entry)| {
+        let spaced_rune = rune.to_string().parse::<SpacedRune>().unwrap();
```
ldiego08, line 17: @raphjaph what's the best way to get a `SpacedRune` from a `Rune`?

raphjaph: It's buried inside the entry struct. Something like this:
```
let spaced_rune = entry.output.rune.unwrap().rune
```

raphjaph (approved):
Nice, LGTM!

PR #3733, opened 2024-05-07
Add decode api

shadowv0vshadow:
Based on the issue https://github.com/ordinals/ord/issues/3676, I have added an API named "decode" to parse the OP_RETURN in transactions (tx).

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -3,9 +3,12 @@ use {
   serde_hex::{SerHex, Strict},
 };
 
-pub use crate::templates::{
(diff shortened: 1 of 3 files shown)


raphjaph (changes requested):
Looks good so far. Could you add a test in `tests/json_api.rs`?

shadowv0vshadow:
The `decode/{txid}`api reuses the code from the `decode` CLI, and the result is consistent with the CLI's result.

raphjaph (approved):
LGTM

Issue #3729, opened 2024-05-06
Add a `ord wallet runics` commmand.


raphjaph:
Works similiar to `ord wallet cardinals`. Should return

- output
- rune name
- rune id
- rune balance
- btc balance

ldiego08:
@raphjaph this has been resolved.

raphjaph:
Thanks!

PR #3734, opened 2024-05-08
Add `ord wallet runics` command

ldiego08:
Add a new `ord wallet runic` command showing runic unspent outputs and balances. Solves #3729.


src/subcommand/wallet/runics.rs
ldiego08, line 1: @raphjaph, does something like this satisfy #3734? It's pretty scrappy code - more interested in functionality atm. I'll polish it if green-lighted.

src/subcommand/wallet/runics.rs
```diff
@@ -0,0 +1,37 @@
+use super::*;
+
+#[derive(Serialize, Deserialize)]
+pub struct RunicUtxo {
+  pub output: OutPoint,
+  pub rune: SpacedRune,
+  pub rune_balance: u128,
+  pub amount: u64,
```
raphjaph, line 8:
```suggestion
```

I know I said to add this but after thinking about it I think we should just show the runes balance and not the btc amount.

src/subcommand/wallet/runics.rs
```diff
@@ -0,0 +1,37 @@
+use super::*;
+
+#[derive(Serialize, Deserialize)]
+pub struct RunicUtxo {
+  pub output: OutPoint,
+  pub rune: SpacedRune,
+  pub rune_balance: u128,
```
raphjaph, line 7:
```suggestion
  pub runes: BTreeMap<SpacedRune, Decimal>,
```
In theory an output can contain multiple runes, so let's make this a map. Have a look at `src/subcommand/wallet/balance.rs` for how to create this struct.

raphjaph (changes requested):
Good start, you have the green light. Don't forget to also add an integration test :)

raphjaph (approved):
LGTM

PR #3735, opened 2024-05-08
tests should fail this commit

twosatsmaxi:


twosatsmaxi:
@gmart7t2: how do I forcefully spend the commitment here to write a failing test? can do it using `bitcoin-cli` but not sure, can use `ord wallet send` but it doesn't provide a way to add specific UTXO.

gmart7t2:
> @gmart7t2 @raphjaph : how do I forcefully spend the commitment here to write a failing test? can do it using `bitcoin-cli` but not sure, can use `ord wallet send` but it doesn't provide a way to add specific UTXO.

Look at how the code sends the etch reveal tx after waiting for maturity. Can you use the same method to try sending it early?

raphjaph:
You can use the TransactionTemplate to construct a transaction with the commit input.

PR #3740, opened 2024-05-09
Fix fuzz testers 

jeasonstudio:


raphjaph (approved):
Thanks for fixing!

LGTM

PR #3745, opened 2024-05-11
Update runes spec

gmart7t2:
We should be clear about what happens when there aren't enough runes available to give the same amount to every output


raphjaph (approved):
Looks good

Issue #3677, opened 2024-04-24
Flag to restrict minting rights


HubertusVIE:
For Bitcredit Protocol, we need a "serious" digital asset to represent guarantee capital which secures Bitcoin Credit Money redemption by permissionless Wildcat mints (based on Cashu). It does nothing, apart from existing positively to be verifiable and tradable. Think of it like a bond. It must sit on Bitcoin mainchain to inherit its security, so Runes would be great. 

It is also necessary that the issuance extends over approximately 130 years, similar to Bitcoin itself. Issuance is a certain number of new guarantee tokens per quarter, declining by a certain percentage each time. Only the project's Governance should have the right to mint the token, it handles further distribution according to work and contributions, not unlike BISQ's coloured coins.

At the moment, we could only do 130x4 declining mints immediately after deployment. Problem is: the total supply would immediately exist from a regulatory / legal / tax PoV, which triggers a rat's tail of real world risks and complications in some countries with overreaching securities regulators. 

Solution: 
1. Add a flag for the etching `governance` yes/no. 
2. When minting, first check if flag is set and only allow minting by the original deployer. 

We'd be most grateful for your help with this requirement for Bitcredit project. Any chance?

cryptoni9n:
I'm not sure I understand the specific use case presented, but I think I can see general uses for such a feature.

It could be sort of a ticketing system rune - with governance, maybe a mechanism is created that watches an address for payments, and that triggers a mint with destination to the payee.  Concert ticket runes could maybe replace something like ticketmaster this way?  Am I off on a tangent?

HubertusVIE:
> watches an address for payments, and that triggers a mint with destination to the payee. Am I off on a tangent?

Agree that ticketing is another "serious" use case, just without the regulatory urgency. 

Bitcredit Protocol plans BISQ like decentralised governance, votes decide number and recipient and automatically trigger a mint.

casey:
> At the moment, we could only do 130x4 declining mints immediately after deployment. Problem is: the total supply would immediately exist from a regulatory / legal / tax PoV, which triggers a rat's tail of real world risks and complications in some countries with overreaching securities regulators.

Can you give a little color on the regulatory / legal / tax problems that creating the supply in advance would cause? I want to make sure that I understand the motivation for this, and that it can't be solved through some other mechanism. If the tokens don't have a guaranteed, fixed value, can't they sit on the books without a tax / legal / whatever event happening on issuance?

[...]

PR #3748, opened 2024-05-13
premine percentage and supply fields update for rune.html

### LINKED ISSUE #3630: Premine percentage formula should use the total supply
Correct me if I am wrong, but this premine percentage formula that is being used in the portal is wrong:
((self.entry.premine as f64 / self.entry.supply() as f64) * 10000.0) as u128

The self.entry.supply() is actually calculating the circulating supply (premine + minted) instead of the total supply (premine + (cap * amount per mint))

cryptoni9n:
presents options to fix #3630

* this is my first attempt at making any sort of real changes using Rust.  I am still in the very early stages of learning Rust and so I wanted to add a disclaimer that these changes were not done with much proficiency - there are probably smarter and better ways to have made them.  I am eager to learn so please provide feedback on what I can do better.  I don't think that all of these fields are necessary for the rune page, but I wanted to display them all so that they could be reviewed and a final decision as to which to keep can be made while considering all possibilities.

changes made:
 - renamed `supply` to `current supply`
 - added `maximum supply` field
 - renamed `premine percentage` to`premine percentage of current supply`
 - added `premine percentage of maximum supply` formula

### DIFF
--- a/templates/rune.html
+++ b/templates/rune.html
@@ -50,12 +50,35 @@
 %% } else {
   <dd>no</dd>
 %% }
-  <dt>supply</dt>
+  <dt>current supply</dt>
   <dd>{{ self.entry.pile(self.entry.supply()) }}</dd>
+  %% if let Some(terms) = self.entry.terms {
+    %% if let Some(symbol) = self.entry.symbol {
+      <dt>maximum supply</dt>
+      <dd>{{ (terms.cap.unwrap_or_default() * terms.amount.unwrap_or_default()) + self.entry.premine }}{{ symbol }}</dd>
+    %% } } else {
+      <dt>maximum supply</dt>
+      <dd>Unable to calculate maximum supply</dd>
+      %% }
+  %% if let Some(terms) = self.entry.terms {
+    %% if let Some(symbol) = self.entry.symbol {
+      <dt>supply progress</dt>
+  <dd>{{ self.entry.pile(self.entry.supply()) }} of {{ (terms.cap.unwrap_or_default() * terms.amount.unwrap_or_default()) + self.entry.premine }}{{ symbol }} </dd>
+  %% } } else {
+    <dt>supply progress</dt>
+    <dd>Unable to calculate supply progress</dd>
+    %% }
   <dt>premine</dt>
   <dd>{{ self.entry.pile(self.entry.premine) }}</dd>
-  <dt>premine percentage</dt>
+  <dt>premine percentage of current supply</dt>
   <dd>{{ Decimal { value: ((self.entry.premine as f64 / self.entry.supply() as f64) * 10000.0) as u128, scale: 2 } }}%</dd>
+  %% if let Some(terms) = self.entry.terms {
+  <dt>premine percentage of maximum supply</dt>
+  <dd>{{ Decimal { value: ((self.entry.premine as f64 / ((terms.cap.unwrap_or_default() as f64 * terms.amount.unwrap_or_default() as f64) + self.entry.premine as f64)) * 10000.0) as u128, scale: 2 } }}%</dd>
+  %% } else {
+    <dt>premine percentage of maximum supply</dt>
+    <dd>Unable to calculate premine percentage of maximum supply</dd>
+  %% }
   <dt>burned</dt>
   <dd>{{ self.entry.pile(self.entry.burned) }}</dd>
   <dt>divisibility</dt>


HubertusVIE:
Thanks for this intiative! Do you think your'd be able to add #3677? It goes well with this one.

cryptoni9n:
I've reverted these changes in order to better understand the test failures they're triggering.  Still working on this, just need to reevaluate the best way to move forward.  Renaming the supply field in the .html page causes over 90 tests to fail.

casey:
I think doing these one at a time would be a good idea. So, for example, start with `supply progress`, which I would actually call `mint progress` in a single PR. It's usually a good idea to make PRs simple and focused, so they can be reviewed easily, and then follow up with additional small PRs.

cryptoni9n:
> I think doing these one at a time would be a good idea. So, for example, start with `supply progress`, which I would actually call `mint progress` in a single PR. It's usually a good idea to make PRs simple and focused, so they can be reviewed easily, and then follow up with additional small PRs.

Thank you, Casey - I appreciate your guidance and patience.  I've taken your advice and just added the `mint progress` field to begin with.  Please let me know what you think and what the next best step would be.

casey:
Nice! I think mint progress should be expressed as a percentage with two digits after the decimal point. So it would be something like `format!("{}%", minted as f64 / total_supply as f64}`.

cryptoni9n:
> Nice! I think mint progress should be expressed as a percentage with two digits after the decimal point. So it would be something like `format!("{}%", minted as f64 / total_supply as f64}`.

how's this look to you?

raphjaph (approved):
Thanks for simplifying this PR, this is much easier to review!

LGTM

raphjaph:
Another tip for follow-up PRs:

It's always best to work on feature branches. Assuming you're on your `master` branch, this is what I recommend:

First pull in the latest changes from the `ord` (`origin`) repo by doing:

```
git pull origin master
```

Then create a local branch:

```
git checkout -b show-current-supply
```

Then make you changes and do `git add`, `git commit`, etc. Try to do a `git push`. You'll probably be prompted to something like this:

```
 git push --set-upstream cryptoni9n show-current-supply
```

Open a PR from that local branch which would have the name `cryptoni9n/show-current-supply`. After the PR has been reviewed and successfully merged you can delete the local branch by doing.

```
git checkout master
git branch -D show-current-supply
```

Repeat this for every feature/PR and you'll have a good work flow :)

cryptoni9n:
Thanks for the tip, @raphjaph - I've been really struggling with figuring out that process and so that is immensely helpful for me. I was already able to use this process once successfully on a new PR I just submitted.  

@casey thank you again for your help on this. what would you recommend for the first additional follow up PR?  I feel like the `Premine Percentage` isn't very intuitive in many cases.  I would like to see it change or replace it with a different metric expressing the percentage as (premine/max_supply).  In my previous attempts, I had renamed the `supply` field to `current supply` and renamed the `premine percentage` field to `premine percentage of current supply`.  I also added new fields for `maximum supply` and `premine percentage of maximum supply` but I'm not sure all of that is necessary.  

I hope memeing is supported on this platform.

7bfb1cc0245d6aefa017517d0e7e29a3361805c9787afb7e52686a82473820b9

casey:
You bet!

I'm not actually sure I would want to expose "premine percentage of maximum supply", because it might not correspond to anything meaningful. The maximum supply can be very high, but if it never actually mints out, then that percentage is a bit misleading.

cryptoni9n:
> You bet!
> 
> I'm not actually sure I would want to expose "premine percentage of maximum supply", because it might not correspond to anything meaningful. The maximum supply can be very high, but if it never actually mints out, then that percentage is a bit misleading.

I see that point and agree.  I feel the heart of the issue is when people would go to this [mintable example rune page](https://ordinals.com/rune/1000 ) and see that the `premine percentage` is 99.83%, it seems to convey two incorrect ideas: 1) there is less than 1% left to mint and, 2) the etcher premined almost 100% of the rune.  I understand that both of these are misconceptions based off of an incorrect understanding of the metric, but a lot of people don't get it visually.

At the same time, going to an [unmintable example rune page](https://ordinals.com/rune/100) and seeing the `premine percentage` of 100% makes perfect sense because it isn't mintable anymore and the same assumptions above are now correct, 1) there is 0% left to mint and 2) the etcher premined 100% of the rune.  

I'm sure this is over complicating things, but it's almost as if there should be different stats for mintable vs unmintable/expired mint runes.  Mintable runes could show current and maximum possible supply stats and percentages, to showcase the growth capability.  Unmintable could have `percent premined` and other hard locked statistics, (supply instead of current or max supply, etc.).  Also, I feel like showing the terms for an unmintable rune is largely unnecessary and should be de-emphasized somehow.

I'm sorry I'm not presenting any great solution suggestions, but this is the crux of what I think needs to be addressed.  If there are no good ideas right now, we can park it for awhile and I will continue to noodle it.

casey:
I do think it would be a good idea, once a rune is unmintable, to display the premine percentage as a percentage of the final supply.

PR #3751, opened 2024-05-14
Add delegate value to inscription endpoint

phorkish:
Add the delegate ID to the `/inscription/[id]` endpoint. If there is no delegate then the response doesn't include that field.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -103,6 +103,8 @@ pub struct InscriptionRecursive {
   pub charms: Vec<Charm>,
   pub content_type: Option<String>,
   pub content_length: Option<usize>,
+  #[serde(skip_serializing_if = "Option::is_none")]
+  pub delegate: Option<InscriptionId>,
   pub fee: u64,
   pub height: u32,
   pub id: InscriptionId,
--- a/src/api.rs
+++ b/src/api.rs
@@ -103,7 +103,6 @@ pub struct InscriptionRecursive {
   pub charms: Vec<Charm>,
   pub content_type: Option<String>,
   pub content_length: Option<usize>,
-  #[serde(skip_serializing_if = "Option::is_none")]
   pub delegate: Option<InscriptionId>,
   pub fee: u64,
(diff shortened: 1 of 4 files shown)


elocremarc:
@gmart7t2 This PR is attempting the same thing as yours gonna link these together. 
https://github.com/ordinals/ord/pull/3761

src/api.rs
```diff
@@ -103,6 +103,8 @@ pub struct InscriptionRecursive {
   pub charms: Vec<Charm>,
   pub content_type: Option<String>,
   pub content_length: Option<usize>,
+  #[serde(skip_serializing_if = "Option::is_none")]
```
raphjaph, line 106:
```suggestion
```
This is a good idea but in order to stay consistent with the other Option fields we should remove it

raphjaph (changes requested):
LGTM, just a small change

gmart7t2:
@phorkish That "this branch is out-of-date with the base branch" means that your PR no longer merges cleanly to the upstream repo because something incompatible was merged. You need to merge your change to the master branch and push the change to your branch. 

It's kind of annoying to have to keep doing that, but someone needs to do it.

raphjaph:
> @phorkish That "this branch is out-of-date with the base branch" means that your PR no longer merges cleanly to the upstream repo because something incompatible was merged. You need to merge your change to the master branch and push the change to your branch. 
> 
> It's kind of annoying to have to keep doing that, but someone needs to do it.

There's a feature on GitHub which auto updates branches if they can be merged cleanly. I'll try to find it and activate

raphjaph:
#3784

raphjaph (changes requested):
Sorry, forgot to add the we should add this field to to docs. Could you add the `delegate:  None` field to the examples in the `recursion.md` docs?

phorkish:
Yep, done. Sorry about missing that.

raphjaph (approved):
LGTM

PR #3756, opened 2024-05-15
Fix typo in mint.rs

gmart7t2:
Postage should be allowed to be equal to the dust limit, just not smaller than it.


raphjaph (approved):
LGTM

PR #3761, opened 2024-05-16
Add `/r/children/:inscription_id/inscriptions` endpoint

gmart7t2:
It also adds `delegate` to the inscription details.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -76,6 +76,13 @@ pub struct Children {
   pub page: usize,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct ChildInscriptions {
+  pub inscriptions: Vec<InscriptionRecursive>,
+  pub more: bool,
+  pub page: usize,
+}
+
 #[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
 pub struct Inscription {
   pub address: Option<String>,
@@ -103,6 +110,7 @@ pub struct InscriptionRecursive {
   pub charms: Vec<Charm>,
   pub content_type: Option<String>,
   pub content_length: Option<usize>,
+  pub delegate: Option<InscriptionId>,
   pub fee: u64,
   pub height: u32,
   pub id: InscriptionId,
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -240,8 +240,14 @@ impl Server {
           "/r/children/:inscription_id/:page",
           get(Self::children_recursive_paginated),
         )
-        .route("/r/children/:inscription_id/inscriptions", get(Self::child_inscriptions_recursive))
-        .route("/r/children/:inscription_id/inscriptions/:page", get(Self::child_inscriptions_recursive_paginated))
+        .route(
+          "/r/children/:inscription_id/inscriptions",
+          get(Self::child_inscriptions_recursive),
+        )
+        .route(
+          "/r/children/:inscription_id/inscriptions/:page",
+          get(Self::child_inscriptions_recursive_paginated),
+        )
         .route("/r/metadata/:inscription_id", get(Self::metadata))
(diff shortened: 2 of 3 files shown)


elocremarc:
Nice! Saves some time. Should we not just add a new field in the current children endpoint responses? @casey Did we decide we did not want to add fields to recursive endpoint updates?

```diff
pub struct Children {
  pub ids: Vec<InscriptionId>,
+ pub inscriptions: Vec<InscriptionRecursive>,
  pub more: bool,
  pub page: usize,
}
```
> It also adds delegate to the inscription details.

If we are not adding fields then should we update this? It would update the current `/r/inscription` endpoint right?

casey:
We need to figure out a way to evaluate the performance impact of PRs like this. Probably the way to do it is to simulate a production workload with tons of requests, and see how it holds up. The worst-case scenario would be we add an endpoint like this, it gets widely used, and we can't ultimately support it, because it overwhelms the ordinals.com servers.

DrJingLee:
> We need to figure out a way to evaluate the performance impact of PRs like this. Probably the way to do it is to simulate a production workload with tons of requests, and see how it holds up. The worst-case scenario would be we add an endpoint like this, it gets widely used, and we can't ultimately support it, because it overwhelms the ordinals.com servers.

Then we buy more powerful servers haha

owenstrevor:
> We need to figure out a way to evaluate the performance impact of PRs like this. Probably the way to do it is to simulate a production workload with tons of requests, and see how it holds up. The worst-case scenario would be we add an endpoint like this, it gets widely used, and we can't ultimately support it, because it overwhelms the ordinals.com servers.

We can simulate a production workload and test with 10,000, 25,000, 50,000, and 100,000+ requests. Is this what you had in mind? Or do you just mean like 1,000 requests?

gmart7t2:
> The worst-case scenario would be we add an endpoint like this, it gets widely used, and we can't ultimately support it, because it overwhelms the ordinals.com servers.

Couldn't you simply reduce the page size until the servers can cope? Don't commit to the page size staying at 100 so inscriptions don't start depending on any specific page size.

owenstrevor:
image

Adding some more context here on how it works from @gmart7t2. There's some interesting implications to draw from it about how ord works for those who aren't familiar.

For our purposes, honestly all we really need is part B to get the blockheight. It would be super nice to get the delegate ID, but I see that is the same as getting the content (assuming you put some kind of limit on the size of the content that would be returned). If we mine for IDs via an inscription service, then we don't need anything from A (albeit that's not a great UX, it's definitely workable).

I'm not sure what others will want. But the crux is the blockheight and the content/delegate can be simulated by mining IDs.

An interesting bit of info for those who aren't familiar with ord, is that we are starting with the existing list of inscriptions here. We can't override the function that gets all the child inscriptions to include the delegate ID or blockheight because that would require to change how ord is indexed. Ultimately, without re-indexing, it seems the most performant thing we can do to get what we need is just simplify the PR to only include call B.

@casey if we do that should we still do some performance testing? Are you confident that just part B would be a sufficiently low burden?

gmart7t2:
I was wrong that "D" would be heaviest because it's fetching the raw tx, because "A" also fetches the raw tx, and so the 2nd (redundant?) fetch is quick due to caching.

And the timings for the Pizza Ninja children are for the first page of 100, not for the whole set

gmart7t2:
> Should we not just add a new field in the current children endpoint responses?

No. The current endpoint is fast, only using the ord database. This new endpoint runs `getrawtransaction` twice per child.

> Did we decide we did not want to add fields to recursive endpoint updates?

No. We decided the opposite. See https://docs.ordinals.com/inscriptions/recursion.html :

"additional object fields may be added or reordered, so inscriptions must handle additional, unexpected fields"

raphjaph:
Closing this in favor of #3771

PR #3762, opened 2024-05-16
Update index.rs to fix ord list command

### LINKED ISSUE #3493: `ord list` doesn't work in 0.17.1
The `is_output_in_active_chain()` function in index.rs is looking at the `in_active_chain` field of the raw transaction output, but it is always None.

That field is only set if `getrawtransaction` is passed a block hash, and it isn't.

This fixes it:

```
diff --git a/src/index.rs b/src/index.rs
index bec84d0a..42047131 100644
--- a/src/index.rs
+++ b/src/index.rs
@@ -1630,7 +1630,7 @@ impl Index {
       return Ok(false);
     };
 
-    if !info.in_active_chain.unwrap_or_default() {
+    if info.blockhash.is_none() {
       return Ok(false);
     }
 
```

cryptoni9n:
this is the code from @gmart7t2 to fix #3493 - I replicated the issue (in 0.18.5) and verified that this fixed it on my local.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -1660,7 +1660,7 @@ impl Index {
       return Ok(false);
     };
 
-    if !info.in_active_chain.unwrap_or_default() {
+    if info.blockhash.is_none() {
       return Ok(false);
     }
 


raphjaph (approved):
Thanks, LGTM!

PR #3766, opened 2024-05-18
Update list.rs to include inscription and rune info

### LINKED ISSUE #3701: `ord list <UTXO>` won't list inscriptions and runes
The command `ord list <utxo>` lists all sats that belong to that UTXO, but it (or any other command) won't list the inscriptions and runes.

cryptoni9n:
fixes #3701 - I've added the `/output/<OUTPOINT>` fields to the `ord list` command into a section called "list'.  I also reordered the fields in the pre-existing "ranges" section to make more logical sense.  There is a small amount of redundancy now with the list.sat_ranges vs ranges.start & end, and list.value vs ranges.size.  I figure that this is acceptable and better than changing the pre-existing ranges section, as some users may already be reliant upon those fields - but let me know if not and if further formatting or changes would be better.

### DIFF
--- a/src/subcommand/list.rs
+++ b/src/subcommand/list.rs
@@ -2,14 +2,21 @@ use super::*;
 
 #[derive(Debug, Parser)]
 pub(crate) struct List {
-  #[arg(help = "List sats in <OUTPOINT>.")]
+  #[arg(help = "List information for <OUTPOINT>.")]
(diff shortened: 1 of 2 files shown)


cryptoni9n (commented):
all suggestions committed; comments removed

raphjaph (approved):
LGTM

PR #3771, opened 2024-05-22
Fetch child inscriptions without using `getrawtransaction`

gmart7t2:
Add `/r/children/:inscription_id/inscriptions` endpoint.

In case #3761 is too heavy, this version avoids doing any `getrawtransaction` calls, at the cost of not returning the following fields: 
* content_type
* content_length
* delegate
* value

On my system this fetches the first 100 pizza ninja children in about 50 ms, compared to 7250 ms using the previous PR (#3761). That's about 150 times faster. Turns out fetching the raw transactions is kind of slow.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -79,6 +79,13 @@ pub struct Children {
   pub page: usize,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct ChildInscriptions {
+  pub inscriptions: Vec<ChildInscriptionRecursive>,
+  pub more: bool,
+  pub page: usize,
+}
+
 #[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
 pub struct Inscription {
   pub address: Option<String>,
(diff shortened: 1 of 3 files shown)


raphjaph (changes requested):
If you could add some docs and also an example call that would be great!

gmart7t2:
> If you could add some docs and also an example call that would be great!

Done.

src/api.rs
```diff
@@ -80,6 +80,13 @@ pub struct Children {
   pub page: usize,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct ChildInscriptions {
+  pub inscriptions: Vec<ChildInscriptionRecursive>,
```
raphjaph, line 85:
```suggestion
  pub children: Vec<ChildInscriptionRecursive>,
```
I don't have strong feelings about this, but from an end user (javascript dev) experience it could make the code clearer if it's called `children` instead of `inscriptions`. I'll refer to you and @owenstrevor

gmart7t2: It makes no difference to me what the field is called. I checked with Trevor and he's fine with it being changed like you suggested. Here's an out of context screenshot as fake proof of that.

Screenshot_20240527_074131_Discord

See? He's fine with it! I'll make a new commit ASAP.

raphjaph (approved):
LGTM

ZedZeroth:
@gmart7t2 I think it also makes sense to add this bulk/batch functionality to reinscriptions on a given sat? In my mind, children are useful for complex tree-like relationships, whereas reinscriptions are linear. Also, as @rabbigains-eth pointed out to me a while back, children can end up on a mess of many different sats, whereas reinscriptions are neater in the sense of being attached to one sat. In short, I think there are strong use cases for both, so both should have this kind of batch endpoint.

PR #3778, opened 2024-05-23
Link to `/address` across explorer

elocremarc:


elocremarc:
@casey how do we make updates to the front end easier without having to get into refactoring obscure tests like `batch_inscribe_works_with_some_destinations_set_and_others_not`?

Issue #3789, opened 2024-05-27
Explorer needs links to new `/address` route


elocremarc:
I started this in https://github.com/ordinals/ord/pull/3778 but gave up because of all the test refactoring.

raphjaph:
yes, you could also pull some changes from this PR #3372

cryptoni9n:
merged

PR #3799, opened 2024-06-04
link address on output & tx pages 

cryptoni9n:
partially addresses #3789 - will link up inscriptions.html separately since it triggers many more tests to fail.


raphjaph (approved):
Thanks!

PR #3156, opened 2024-02-22
Allow configuring chain with config file

casey:

PR #3800, opened 2024-06-04
Make Settings public

felipelincoln:
PR #3156 made the change
```diff
# src/index.rs:224
  pub fn open_with_event_sender(
-    options: &Options,
+    settings: &Settings,
    event_sender: Option<tokio::sync::mpsc::Sender<Event>>,
  ) -> Result<Self> {
```

And now we need the `settings` module to be public so we can use it on a separate rust app
example:
```rust
// copy of @mi-yu's code
use ord::{
    index::{event::Event, Index},
    options::Options, settings::Settings,
};

async fn main() {
    println!("Starting ord indexer...");

    let index_options = Options::parse();
    let index_settings = Settings::load(index_options).unwrap();
    let (sender, mut receiver) = tokio::sync::mpsc::channel::<Event>(128);
    let index = Index::open_with_event_sender(&index_settings, Some(sender)).unwrap();

    ...
}
``` 

This change was requested before, for the `options` module: #3138


raphjaph (approved):
LGTM

PR #3803, opened 2024-06-05
Update install.sh to add install location to system path.

so7ow:
The current [readme](https://github.com/ordinals/ord?tab=readme-ov-file#installation) has users run `ord --version` after executing the install.sh script, but that command won't usually work as expected. This PR updates install.sh to add the installation location to the system path, so that the recommended command will work.


casey:
I think this is a bit invasive, I don't really like things to modify my configure scripts automatically. I think a better approach is checking if the destination path is in `$PATH`, and if not, print a warning that the user should add it to their path.

PR #3805, opened 2024-06-06
Add txdata to block json response to sync easier outside rust

thewrlck:
_**I am very dumb on rust sorry if that's wrong.**_

The html returns the txs when you point to /block/840000 but not the json api.

Should be very cool that also the api returns the txs ids, in this way you have two options for syncing on top of ordinals.

1 - use events emitted by rust mounting rust on rust
2 - use whatever language you want to query block x block / tx x tx.

Can you add Vec Transaction also to the response of the json at /block endpoint?

I know that you can do directly using bitcoind but should be useful to have it on that endpoint.

https://github.com/ordinals/ord/issues/3770

@raphjaph @casey


thewrlck:
@raphjaph i realizes that was failing i updated

raphjaph (approved):
LGTM

thewrlck:
<3

PR #3811, opened 2024-06-07
Create public shutdown function

felipelincoln:
This enables apps to gracefully shutdown the `index.update()`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -220,7 +220,7 @@ pub fn parse_ord_server_args(args: &str) -> (Settings, subcommand::server::Serve
   }
 }
 
-pub fn shutdown(){
+pub fn shutdown() {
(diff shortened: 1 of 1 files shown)


felipelincoln:
> What's the purpose of this refactor? If it's just a refactor, I prefer the original version, but if the goal is to be able to shut down the index, couldn't you just call `gracefully_shutdown_indexer`? Maybe that should be pub instead?

The `gracefully_shutdown_indexer` requires that we register the thread running the `index.update()` onto the static `INDEXER` . E.g. https://github.com/ordinals/ord/blob/master/src/subcommand/server.rs#L166
I am not sure if that's ideal

casey:
Gotcha, that makes sense. I renamed "shutdown" to "shut_down" since it's a verb.

PR #3812, opened 2024-06-08
Add -dev suffix to version

### LINKED ISSUE #3791: Versioning is confusing when building master
It seems to reflect the most recently released version of `ord`, which is very confusing to users, especially when master has a schema update and isn't compatible with the most recent release. 

Example: at some point after 0.18.5 was released there was a schema update in the code. Now, if you build from master your binary will say it is version 0.18.5, but it won't read an index file created by the 0.18.5 release. ➡️ confusion 

Any good way to change the version after a release, to reflect the unreleased nature of building from master?

casey:
Fixes #3791.


raphjaph (approved):
Cool, LGTM!

Issue #3792, opened 2024-05-28
`ord wallet sats` should return all sat ranges not just rare sats


raphjaph:

cryptoni9n:
Hi Raph - I'm looking at working on this one, but want to ask some questions first.

1) I previously patched the `ord wallet outputs` command on my fork to show the sat ranges per output.  It worked well but I never created a PR for it and think I've since blown away that code.  My question is would it be better to keep `ord wallet sats` as is and just display the sat ranges with the outputs?  That way sats would still only show special sats but outputs would show all?

2) If it is more intuitive to show all the sat ranges in the sats command (either instead of, or in addition to also showing them in outputs), how do you think it should be best displayed?  Today, the special sats get shown like this:

```[
  {
    "sat": 315000000000,
    "output": "227ca0b1b54aac62b31cdb899fcc0103e572cefa27e68d92362f4b1b3ce2b904:0",
    "offset": 0,
    "rarity": "uncommon"
  },
  {
    "sat": 155000000000,
    "output": "1561a009ce7db4199cb77c652796c51dcba22273cecec13356dd9ba4d936d806:0",
    "offset": 0,
    "rarity": "uncommon"
  }
]
````
I suggest something to the effect of:

```[
  {
    "sat": 315000000000,
    "output": "227ca0b1b54aac62b31cdb899fcc0103e572cefa27e68d92362f4b1b3ce2b904:0",
    "range": 315000000000–320000000000,
    "offset": 0,
    "rarity": "uncommon"
  },
  {
    "sat": 155000000000,
    "output": "1561a009ce7db4199cb77c652796c51dcba22273cecec13356dd9ba4d936d806:0",
    "range": 155000000000–160000000000,
    "offset": 0,
    "rarity": "uncommon"
  },
  {
    "output": "b9fa5b9443459832c70e21a67612983bdfbc3e5f3e7c1902e5e896e62a43b351:25",
    "range": 747873459178359–747873459226914, 1457725397711488–1457725398115345, 379295356533988–379295356867957, 106824711233364–106824711413209
   }
]
````

What do you think?

mmikrut:
Hi, perhaps adding  output modifier parameter to `ord wallet sats`  command ( eg: `--uncommon` | `--all`)   would be a good idea?

mmikrut:
I also think that your original patch to add lists of sats to `ord wallet output` is a much desired option, I would love to test it !

[...]

PR #3817, opened 2024-06-11
Add sat ranges to output

cryptoni9n:
partially addresses #3792 by adding the sat ranges to the `ord wallet output` command.  My feeling is that the sat ranges are information about the outputs, so this is a good place for this information.  

Next, I'll add the --all flag  so that `ord wallet sats --all` will return all sat ranges, and `ord wallet sats` without a flag will continue to only show rare sats (as it does today).  

The description for the command says `List wallet satoshis` - but what do we want this output to look like?  Should it simply be a list of sat ranges with no reference to the outputs, or anything else?  Would something like the mock-up below be acceptable? or do we want to display more info?

> `ord wallet sats --all`
[1111111111111111-1111111111111112, 1111111111111113-1111111111111114, 1111111111111115-1111111111111116, 1111111111111117-1111111111111118, 1111111111111119-1111111111111120, 1111111111111121-1111111111111122]

### DIFF
--- a/src/subcommand/wallet.rs
+++ b/src/subcommand/wallet.rs
@@ -62,7 +62,7 @@ pub(crate) enum Subcommand {
   #[command(about = "Mint a rune")]
   Mint(mint::Mint),
   #[command(about = "List all unspent outputs in wallet")]
-  Outputs,
+  Outputs(outputs::Outputs),
   #[command(about = "List pending etchings")]
   Pending(pending::Pending),
   #[command(about = "Generate receive address")]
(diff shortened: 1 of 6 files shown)


raphjaph (changes requested):
I've simplified the code a bit and added a test. @cryptoni9n could you also add in this PR the `--ranges` flag. So you only see that ranges when it is specified. Let me know if you have any other questions.

cryptoni9n:
> I've simplified the code a bit and added a test. @cryptoni9n could you also add in this PR the `--ranges` flag. So you only see that ranges when it is specified. Let me know if you have any other questions.

Thanks Raph, I appreciate your help with this.  I'm out on vacation for about the next 10 days.  Sorry for the crap timing but I'm not sure I'll be able to get to this or the updates needed for my other PR's before then.  Will be sure to tackle this and the snafu updates when I get back.

raphjaph:
No worries, have fun on your vacation!

raphjaph (approved):
Just added the flag quickly myself.

LGTM

PR #3820, opened 2024-06-12
reset `SHUTTING_DOWN` when starting `update()`

felipelincoln:
calling the recently introduced `shut_down` function #3811 toggles the `SHUTTING_DOWN` state and calling `index.update()` after it wouldn't work as expected since the ord is "shutting down".
This change enables the update to be restarted.


raphjaph:
This would create a race condition. So probably best to create a dedicated function to cancel shutdown. Something like below:

```rust
pub fn cancel_shutdown() {
  ...
}
```

felipelincoln:
yeah that sounds better @raphjaph :+1:
I have updated the pr

raphjaph (approved):
LGTM

PR #3826, opened 2024-06-20
adding sat name to inscription page

### LINKED ISSUE #3199: Display sat name on inscription page
Maybe also the degree.

cryptoni9n:
fixes #3199 

I didn't add the degree, though.  Please let me know if that is still desired and if you think it should be in a separate PR or not.

### DIFF
--- a/templates/inscription.html
+++ b/templates/inscription.html
@@ -73,7 +73,7 @@
   <dt>sat</dt>
   <dd><a href=/sat/{{sat}}>{{sat}}</a></dd>
   <dt>sat name</dt>
-  <dd>{{ sat.name() }}</dd>
+  <dd><a href=/sat/{{sat.name()}}</a></dd>
 %% }
 %% if let Some(metaprotocol) = self.inscription.metaprotocol() {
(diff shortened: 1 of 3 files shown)


casey:
I think degree is probably not worth it. I don't think anyone aside from me has ever even mentioned degree notation 😅 Otherwise this looks good to me. @raphjaph what do you think?

templates/inscription.html
```diff
@@ -72,6 +72,8 @@ <h1>Inscription {{ self.number }}</h1>
 %% if let Some(sat) = self.sat {
   <dt>sat</dt>
   <dd><a href=/sat/{{sat}}>{{sat}}</a></dd>
+  <dt>sat name</dt>
+  <dd>{{ sat.name() }}</dd>
```
raphjaph, line 76: I think it would be cool to also link to `/sat/{sat_name}`. So just add an href and then this can be merged :)

raphjaph (changes requested):

raphjaph (approved):
LGTM

Issue #3809, opened 2024-06-07
Add more information to `/address`


raphjaph:
We recently merged https://github.com/ordinals/ord/pull/3757, which adds a more comprehensive address to output tracker. If you look at it you'll see the html page is just a list of outputs. If you want you could add more descriptive and aggregated information like inscriptions, btc, sats and runes.  Best to chunk these up into different PRs.

raphjaph:
Here's the PR #3810 that implements this for btc

cryptoquick:
Looking at the current HTML page, it looks like we have the following:

 - [x] inscriptions
 - [x] btc (sat balance)
 - [ ] sats
 - [x] runes

Also, the JSON response needs to be updated with this information as well.

We're also missing individual UTXO value, which would be really helpful for us as an alternative to running a separate electrs index, which would be a shame, since we're using just ord for just about everything else we need.

If I added this to the HTML page and JSON endpoint, would you be open to a PR?

raphjaph:
Yes, I'd be more than happy to review a PR!

I'm a bit hesitant about adding the value since you could just make a second call to the `/output/<OUTPOINT>` endpoint to get that information + the script-pubkey (which is actually important for transaction construction). I would be interested to know how the electrs endpoints are structured? What exactly do they return?

You could also add the rare sats and then it would show all the different assets and address can hold.

Another thing would be pagination such that the top-level address page shows 10 entries of each list and the we have pages such as `/address/<ADDRESS>/inscriptions/<PAGE>` or `/address/<ADDRESS>/runes/<PAGE>`.

cryptoquick:
> I'm a bit hesitant about adding the value since you could just make a second call to the /output/<OUTPOINT> endpoint to get that information + the script-pubkey (which is actually important for transaction construction).

I'll try this and see if that produces a satisfactory result, thank you.

> I would be interested to know how the electrs endpoints are structured? What exactly do they return?

Example electrs request:
`http://localhost:3002/address/bcrt1pgt8hl5dwsy7caxakqqtvrjkuu9dzpdyjxag0znjfsnwdses6nvasellazx/utxo`

Example response to that request:

```json
[
    {
        "txid": "ff3e3940fbfe5ce2152c99bf9ef7810df87634d8547414ad677bad9ae4334d01",
        "vout": 0,
        "status": {
            "confirmed": true,
            "block_height": 415,
[...]

PR #3831, opened 2024-06-24
Display rune balances in address page

yoitsyoung:
Hi! First ever public pull request 🙏  Please feedback is welcome

Referencing issue #3809 to add more information to address page. I also refactored the tests on templates/address.rs so that different categories of address information are tested separately. 

Unfortunately, I'm not able to get the integration at tests/server.rs to work; it seems to be that rune commands are not compatible with the mockcore::spawn() handle, hence it is necessary to enforce the transaction on regtest; however I get a 404 error when requesting pages with a regtest address format; so the test currently always fails.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -2233,24 +2233,38 @@ impl Index {
       .collect()
   }
 
-  pub(crate) fn get_rune_balances_for_outputs(&self, outputs: &Vec<OutPoint>) -> Result<Vec<(SpacedRune, u128)>> {
-    let outpoint_to_rune_balances = self.database.begin_read()?.open_table(OUTPOINT_TO_RUNE_BALANCES)?;
-    let id_to_rune_entry = self.database.begin_read()?.open_table(RUNE_ID_TO_RUNE_ENTRY)?;
-    let mut acc = Vec::new();
+  pub(crate) fn get_aggregated_rune_balances_for_outputs(
+    &self,
+    outputs: &Vec<OutPoint>,
+  ) -> Result<Vec<(SpacedRune, Decimal, Option<char>)>> {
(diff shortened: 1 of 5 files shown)


yoitsyoung:
cc @cryptoni9n

raphjaph (changes requested):
Thanks for the PR, this is great!

I've made some changes to help you along a bit. Since we are aggregating potentially the same runes in different outputs I added this rather complicated looking addition logic. Basically I create a Map with SpacedRunes to Pile (Amount) and if we have already encountered this rune in a previous output instead of inserting we add the balance on to the existing. 

The reason the tests weren't working because you forgot to add the `--index-addresses` flag to `TestServer::spawn`.

I also added the symbol next to the amount.

Some tasks for you if you'd like to refine this PR:

1. Send multiple different runes to the same address
2. Send the same rune twice to the same address (to check that different outputs are aggregated correctly)
3. We could sort the the runes lexicographically (optional)
4. Add a link (href) to the Rune name in the html so a user can click on it to go to that runes etching/parameter page.

If you have any other question, don't hesitate to ask!

yoitsyoung:
Thanks @raphjaph! Updated as per your comments. Definitely should have thought about aggregating the Rune balances, stupid me; I didn't sort by alphabet as I thought it we'd want to keep the client as un-opinionated as possible, but feel free to do so of course.

raphjaph (approved):
Perfect PR!

LGTM

yoitsyoung:
> Does this also apply to the JSON api?

I believe the address JSON-RPC endpoint only returns a list of outputs currently, @raphjaph I was thinking we can serialize the AddressHtml struct and pass it into the response?

PR #3832, opened 2024-06-25
Add typed errors with `snafu`

casey:
This migrates a couple error sites over to `snafu`, a library for typed errors. `snafu` has a nice system of context selectors, which makes it easy to attach context to an underlying error.

This only migrates two error sites, and leaves migration of other error sites to future PRs.

@raphjaph @cryptoni9n

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2305,6 +2305,7 @@ dependencies = [
  "serde_with",
  "serde_yaml",
  "sha3",
+ "snafu",
  "sysinfo",
  "tempfile",
  "tokio",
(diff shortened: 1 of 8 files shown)


cryptoni9n:
I've replicated these changes on my local and get the same 3 regex test failures.  I've been trying to fix them but am not making any progress.  :/

casey:
These were just due to a change in how the source error chain was formatted. Fixed in https://github.com/ordinals/ord/pull/3832/commits/cfe494d72fc9b1f74e7f9000fa821a00c0f63458

PR #3833, opened 2024-06-26
Require Bitcoin Core 27.0.0 or newer

### LINKED ISSUE #3781: Upgrade to Bitcoin Core version 27.0
Right now we require version 24.0. We should try upgrading and see if all the rpc calls still work.

ldiego08:
Closes #3781


raphjaph:
Thanks for the PR! 

I think I was wrong on that issue (or at least imprecise). We don't need any new features from Bitcoin Core and it would be best to keep that backwards compatibility with 24.0.0. 

When I wrote that issue I was just thinking: What would happen if you run `ord` with a newer version. Do RPC calls break? Stuff like that. I'll close the issue.

Issue #3846, opened 2024-07-10
Add pagination to /address page


cryptoni9n:
I see a couple of issues in v0.19.0 on the address page:

**FIXED** 1) from #3831 - some runes are not being displayed/linked correctly.  [Example 1](http://ordinals.com/address/bc1pvjx6xqa4rkvxc5suw6q9zrj4hz7aqz6r3a06kp5dpqcht6jq8jnqekq3kr),  [Example 2](http://ordinals.com/address/bc1pjjzkh79z7v3c0phxcykcfuz774n84ef77gqpmv36a2vmsmd0nzrqqa46k6)

2) from #3843 - some addresses have a very large number of inscriptions and outputs.  these pages take forever to load and could be detrimental to the site.  I think pagination is needed: http://ordinals.com/address/bc1pafhpvjgj5x7era4cv55zdhpl57qvj0c60z084zsl7cwlmn3gq9tq3hqdmn

edit: reopening to keep track of pagination issue

raphjaph:
Just updated ordinals.com. Yes, we definitely need pagination. Just wanted to get the release out so we skipped that. You could try your hand at that if you want :)

cryptoni9n:
adding some more info here for the first issue - the rune display issue only seems to occur with runes that have the `¤` as their currency symbol.

cryptoni9n:
reopening to keep track of pagination issue

raphjaph:
Jup jup, we should add this

alpeshvas:
How would pagination work though considering there are different type of assets in one page: UTXOs, inscriptions, runes etc…! Ideally all of these have their own unique identifier (URI) like 

/address/{addy}/inscriptions, /address/{addy}/inscriptions etc…

twosatsmaxi:
> Jup jup, we should add this

How would pagination work though considering there are different type of assets in one page: UTXOs, inscriptions, runes etc…! Ideally all of these have their own unique identifier (URI) like

/address/{addy}/inscriptions, /address/{addy}/inscriptions etc…

raphjaph:
> How would pagination work though considering there are different type of assets in one page: UTXOs, inscriptions, runes etc…! Ideally all of these have their own unique identifier (URI) like
> 
> 
> 
> /address/{addy}/inscriptions, /address/{addy}/inscriptions etc…

[...]

PR #3849, opened 2024-07-10
fix rune links for runes with default currency symbol on /address page

cryptoni9n:
partially addresses #3846 by fixing the linking/display for runes that have the default currency symbol.  runes with default currency symbol will now be linked and show the symbol on the /address page.


raphjaph (approved):
LGTM!

Issue #3192, opened 2024-02-28
Use proper error types


casey:
Currently, we use `anyhow` for all errors. I like anyhow, but it makes it too easy to bubble up an error without context, which leads to a less-than-useful error message. For example, I/O errors don't include the offending file, making them quite inscrutable. We should probably switch to using something like snafu or thiserror.

cryptoni9n:
Hi @casey & @raphjaph  - I've started working on converting the code base over to `thiserror` and wanted to check-in with you to make sure that this is still desired and won't all be for nought.  

I've created an errors.rs under /src and am stuffing most of the custom errors into it. 

I started out converting the files underneath /src/subcommands/wallet and it just kind of snowballed from there.  I've been working on the src/settings.rs and that one is requiring many, many changes.

I know that this is a ton of changes, but I hope to get through it all eventually.  I'm not sure exactly how I should submit the changes once I can get everything ready.  

Any thoughts/feedback?

casey:
Thanks for tackling this! I think it will make error messages much more useful, and I wish we did it from the beginning.

I actually think that prefer `snafu` over `thiserror`. The context selector is a very nice way to select an error type. For example:

```just
#[derive(Snafu)]
enum Error {
  #[snafu(display("I/O error at `{}`", path.display()))]
  Io { path: PathBuf, source: io::Error}
}
```

Then, from another module:

```rust
do_something(&path).context(error::Io { path: &path })?;
```

Is there any way to do this piecemeal? It will be an absolutely massive PR, and will conflict with basically every other PR.

Like converting the top-level error type to:

```rust
#[derive(Snafu)]
enum Error {
  #[snafu(display("{0}"))]
  Any(Box<dyn std::Error::Error),
  #[snafu(display("I/O error at `{}`", path.display()))]
  Io { path: PathBuf, source: io::Error}
}
```

And then adding error cases one by one.

cryptoni9n:
Thanks for the quick response, Casey.  If I understand you correctly (and I may not), the context selector that you've mentioned seems very similar to what I've been doing with `thiserror`.  The errors.rs creates the MyError enum:

```
use std::env::VarError;
use std::io;
use std::num::ParseIntError;
use std::num::TryFromIntError;
use thiserror::Error;
use redb::TransactionError;

#[derive(Error, Debug)]

pub enum MyError {
  #[error("Invalid environment variable `{var}`: `{value}`")]
  InvalidEnvVar { var: String, value: String },
  },
```

[...]

PR #3854, opened 2024-07-12
migrating outgoing.rs to snafu error handling

cryptoni9n:
part of fix for #3192 
converted src/outgoing.rs to use snafu instead of anyhow errors


cryptoni9n:
Hi @casey - I revised this PR and tried to apply your advice from #3858 to it.  Please review and let me know what you think - thanks!

src/error.rs
```diff
@@ -45,6 +46,39 @@ pub enum SnafuError {
   },
   #[snafu(display("Unrecognized representation `{}`", input))]
   UnrecognizedRepresentation { source: error::Error, input: String },
+  #[snafu(display("Unrecognized outgoing amount: `{}`", input))]
+  UnrecognizedAmount {
```
casey, line 50: I think I would call these `*Parse` instead of `Unrecognized*`.

src/error.rs
```diff
@@ -45,6 +46,39 @@ pub enum SnafuError {
...
+  },
+  #[snafu(display("Unrecognized outgoing inscription ID: `{}`", input))]
+  UnrecognizedInscriptionId {
+    source: inscriptions::inscription_id::ParseError,
+    input: String,
+  },
+  #[snafu(display("Unrecognized outgoing: `{}`", input))]
+  UnrecognizedOutgoing { input: String },
+  #[snafu(display("Failed to parse decimal: {}", source))]
+  DecimalParse { source: error::Error, input: String },
```
casey, line 77: If this is only used in parsing rune amounts, then maybe `RuneAmountParse`.

casey (changes requested):
Looks good! Left a couple comments.

src/error.rs
cryptoni9n: Thanks Casey - I'm pushing changes now that rename all of the `Unrecognized*` errors, except for `UnrecognizedRepresentation`, because that was already merged as part of #3192.

src/error.rs
cryptoni9n: sure thing, pushing changes now.  Are you ok with having the following errors be so similar, or would you prefer to merge them?

OutgoingRuneParse and RuneParse
OutgoingSatParse and SatParse
OutgoingInscriptionIdParse and InscriptionIdParse

casey: Whoops, sorry, I wasn't looking at those other errors. Those errors can all be combined. In general, existing errors should be reused.

cryptoni9n: ok, updates pushed.

casey (approved):
LGTM! I made a couple of small changes before merging.

- If a name conflicts, I usually don't import and rename it, I just refer to it by its full name, so I removed the `use bitcoin::address::Error as AddressError` and just referred to it as `bitcoin::address::Error`.
- I made `Representation::from_str` return a SnafuError, which avoids needing to wrap an anyhow error in a snafu error.

PR #3856, opened 2024-07-15
Commit twice to work around redb bug

casey:


cryptoni9n:
closed/reopened to restart checks.

casey:
With this commit, index is 221 GB, instead of 289 GB, so -68 GB.

raphjaph (approved):

PR #3858, opened 2024-07-16
migrate object.rs to snafu error handling

cryptoni9n:
part of fix for #3192 
makes SnafuError public within error.rs;
adds parse errors to error.rs;
updates test in parse.rs;
migrates object.rs to snafu error handling from anyhow


casey (changes requested):
Nice! A few comments!

- Error enum variants shouldn't have `Error` in their name, since it's redundant, i.e., they're variants in an enum which already has `Error` in the name.
- snafu errors which add context to an existing error should include the source error 
- To construct a snafu error, use `snafu_context`, and then provide an error selector, which in the case of the `error::Error` enum, will be `error::EnumVariantName`.

I just pushed a commit which does these for `AddressParseError`:

- It's now just called `AddressParse`
- It includes the underlying error, which is a `bitcoin::address::Error`
- It uses `.snafu_context(error::AddressParse { … })` to construct the error

cryptoni9n (commented):
thanks, for making these changes.  they'll be very useful for me as a reference for the requested updates.

cryptoni9n:
Hi Casey - Please see the latest changes.  I did my best in figuring out what the underlying / source error should be for each of these parse errors, but had to make a few things public from crates/ordinals/src/lib.rs in order to reference them - not sure if that's ok or not.  Please take a look and let me know what you think.

casey (approved):
Looks good to me! I changed the name of the argument in `from_str` to `input`, which allows using the field shorthand syntax (`{ input }` instead of `{ input: s }`) since it looks a little nicer.

PR #3884, opened 2024-08-05
Re-enter beta

casey:


raphjaph (approved):
Runes are dead but now in beta!

cbspears:
What is dead may never die

But it can enter beta

raphjaph:
> What is dead may never die But it can enter beta

This is must be how it feels to work on some dead bullshit web3 chain with no users. The death march continues

HubertusVIE:
> This is must be how it feels to work on some dead bullshit web3 chain with no users. The death march continues

Resurrecting the "dead"

May I remind all that the real world actually would need "Serious Runes" for economically useful applications. This is prevented by the current "degenerate minting" (vulgo misnomer "fair mint").

Meanwhile we had to move to BRC-20 5-Bytes. Happy to move to Runes if they get a nifty "self-mint" feature. 

Can we re-open the issue? There are several other people who signalled support in the thread. 

Flag to restrict minting rights #3677

PR #3889, opened 2024-08-07
Refactor from instructions envelope

SchnobiTobi:


casey:
Thanks for the PR! I don't really mind the verbosity here, since it's pretty critical code. We also don't touch it much, so I don't think it's worth refactoring.

PR #3891, opened 2024-08-10
add address feild to `/r/inscription/:id`

elocremarc:
This was removed in the inital PR in this commit https://github.com/ordinals/ord/pull/2628/commits/cc832bbf8964c5b208e803759fb5f79f8aecc94a
We need this field for `/r/address/:address` to work properly.


raphjaph:
I think we can add this back. The address is generated from the script pubkey so there is no dependency on the address index

elocremarc:
@raphjaph that was my thoughts too. Also I put address at the end of the response object because if someone used an array index instead of the key we won't break their inscriptions adding a new field.

mutabletrees:
Would love to have this one soon (as well as /r/address).

raphjaph (changes requested):
@elocremarc could you add this field to the recursive endpoint docs?

raphjaph (approved):
LGTM

PR #3894, opened 2024-08-10
Don't unnecessarily insert into utxo cache when indexing addresses

raphjaph:


casey (approved):
LGTM, let our sins be washed away as if in a cleansing rain.

PR #3899, opened 2024-08-14
Draft: Liquid Etchings

onchainguy-btc:
_This is a draft PR. Although the code builds, it is not yet functional. Tests are failing and some stuff was just commented out to make the build work. More fancy variable names need to be brainstormed. The purpose of this PR is to provide an idea of how a lightweight AMM and liquidity bootstrapping implementation could look for runes._

### The Idea

We cannot swap on-chain against native BTC (yet). Therefore, this PR only allows swapping runes against other runes. However, it is possible to create some WBTC rune with a centralized treasury. This PR also assumes that the most "entertaining" aspect for degens is the ability to etch runes in a mode where liquidity is bootstrapped, and the etched rune can be traded on-chain immediately. To achieve this, the etcher of that rune must specify a base rune to mint the new rune. The base rune can be any rune, such as UNCOMMON•GOODS or DOG•GO•TO•THE•MOON (random examples), assuming these have good off-chain liquidity in the long term. Each mint then requires a certain amount of base runes, all of which are locked in a liquidity pool. Once the rune is fully minted, a `seed` (a certain amount of the etched rune) is added to the pool along with the base runes.

### Problem: MEV

This PR does not yet include MEV prevention, but here are some ideas that might eliminate MEV:

- The output of a swap may not be used to swap again in the same block to prevent sandwiching.
- Commit/reveal-based swaps for large trades to prevent frontrunning. Miners could still censor reveals but setting a narrow slippage can make that unprofitable.

### Why?

It would be beneficial if runes could contribute more to miner fees as the subsidy decreases. Rune mints have slowed down since launch. Bootstrapped on-chain liquidity could bring additional transaction volume to the network and could also make us more independent from CEX-listings for liquidity.

@casey @raphjaph If you are open to integrating this feature, my friends and I are happy to fully implement it.


casey:
Sorry to close this, but anything that introduces MEV is a non-starter.

casey:
#

casey:
Also, I would prefer to see swaps implemented with existing bitcoin primitives, which should eventually be sufficient, not rune-specific primitives.

onchainguy-btc:
> Sorry to close this, but anything that introduces MEV is a non-starter.

@casey I opened the PR because we can make swaps work without introducing MEV. Without generic programmability prevention of MEV will be much easier than on Ethereum.

> Also, I would prefer to see swaps implemented with existing bitcoin primitives, which should eventually be sufficient, not rune-specific primitives.

I also don't view this PR as a competition to concepts like light pools but more as a fun addition to rune etchings. If runes have a proper AMM, the chances are actually higher that no other meta protocol will introduce an AMM with less MEV protection.

casey:
What MEV were you referring to when you wrote "This PR does not yet include MEV prevention"?

casey:
Can you also describe what this PR does so I can understand it without needing to refer to the code?

onchainguy-btc:
> What MEV were you referring to when you wrote "This PR does not yet include MEV prevention"?

Biggest MEV potential would be sandwiching. User submits a swap with a 3% slippage, miner puts a buy right before the swap and a sell right after to extract the max potential slippage. No risk for the miner because they can buy and sell in the same block. To avoid that (not included in the code yet), we can design it so that runes on a swap output can not be swapped within the same pool again in the same block (maybe even 2 blocks). This way, if a miner wants to extract the slippage, they would need to hold different amounts of runes on different outputs. However, this comes with the risk of holding certain runes and is not really MEV but market making. 

> Can you also describe what this PR does so I can understand it without needing to refer to the code?

The PR adds a new etch mode to runes where you can mint runes by spending existing runes. Once the mint cap is reached, both the spent runes plus a configured amount of the newly etched rune (seed) is put into a pool which people can trade against. Therefore a new swap flag is introduced. 

Some stuff you would find in an Ethereum style AMM has been left out to keep the upgrade simple. For example, you can’t add or withdraw liquidity (both runes) from pools. The liquidity is not owned by anyone.

casey:
I have a few thoughts:

- Most wrapped BTC implementations are fully centralized. I don't want to encourage their development or use.
- I don't think using runes on both sides of a trading pair is useful. BTC should be on one side of every trading pair to avoid fragmentation.
- There is MEV introduced by this implementation, and avoiding it requires a bunch of weird restrictions.

I think the proper direction for experimenting with improving trading is with base-layer bitcoin, which, although limited, is programmable.

Issue #3901, opened 2024-08-15
`ord wallets` 


raphjaph:
- list all Bitcoin Core ordinals wallets

cryptoni9n:
would this just be the same output that is produced from `bitcoin-cli listwallets`?

cryptoni9n:
> @cryptoni9n If you want you could implement the `--ordinals-only` flag. This would mean loading all wallets (which can take a while) and then checking that the descriptor matches the expected. We have helper function for that somewhere. That PR should also make sure to handle the waiting gracefully and not error while the wallets are loading.

This sounds a little above my head, not sure I'll be ready for this one anytime soon.

raphjaph:
It's a bit complex, that's why I was too lazy to do it myself. The `ord wallets` command gets you most of the way but would be extra nice to only show actual ordinals wallets.

PR #3902, opened 2024-08-16
List all Bitcoin Core wallets

raphjaph:
This basically just wraps `bitcoin-cli listwalletdir`. In order to filter by if it is an ordinals wallet we would have to load all wallets and then check the descriptor. This can take a while so I haven't implemented it here. I think we could add `--ordinals-only` that does this this.


raphjaph:
#3901

Issue #3900, opened 2024-08-15
Output page should link directly to first sat in output


raphjaph:

cryptoni9n:
Hi @raphjaph - I'm assuming this is done with the merge of #3903.  Please reopen if not.

PR #3903, opened 2024-08-16
Make first first and last sat in range clickable

raphjaph:
#3900


raphjaph:
I don't understand why these tests are failing, i hate the html tests.

casey:
The only additional information that the range page has is the total number of sats in the range. Instead of adding a `(range)` link, we could instead directly add the value to the ranges:

```
0–10 (10 sats)
```

And remove the range page entirely. The range page is one of the most useless pages anyways.

raphjaph:
I like that, will remove the ranges page in this PR

raphjaph:
It's done but I can't figure out what's wrong with the html tests. Could you maybe glance at them for 2mins? Maybe it'll jump out to you

casey:
Is the the parens? `(…)` is the regex grouping operator, so it doesn't match a literal `(…)`. I would just remove that `(range)` note anyways.

casey (approved):

cryptoni9n:
>  we could instead directly add the value to the ranges:
> 
> ```
> 0–10 (10 sats)
> ```

Wouldn't 0-10 be (11 sats) since the range should be inclusive on both sides?  For example, [this output](https://ordinals.com/output/8b332e313c0a55168291861805240840b0a9ae48be410fbdb14237792b32af76:273) shows that its range is 600 sats, but isn't it actually 601?

raphjaph:
I think we have to change how we display this. It's actually a half open range like so [0, 10), which means the last is not included.

raphjaph:
#4075

PR #3912, opened 2024-08-25
Update Bitcoin Core deploy to 26.2

casey:


casey:
This appears to work. In order to trigger the update on deploy, bitcoind must be deleted. Also, it seems like running `bitcoind --version` now creates a `settings.json` file, which will have the wrong permissions since it's run as root, so we chown it to the right user.

raphjaph (approved):
LGTM

PR #3926, opened 2024-09-03
Transaction data, inputs & outputs #3709

bitmapstr:
Added endpoint /r/txs/:query to retrieve tx ins & outs recursively.
Updated docs.


raphjaph:
This does too many things at once. Also if we add support for transactions we'd probably just add `/r/rawtx` and let clients parse out all things they need from a tx.

PR #3932, opened 2024-09-04
Delegator content `/r/undelegated-content/<INSCRIPTION_ID>`

elocremarc:
Allows a delegated inscription to acess its own content with `/r/delegator/:id` closes https://github.com/ordinals/ord/issues/3700

An inscription with delegated content "foo" and original content "bar" will return the following:

| Endpoint           | Response |
|--------------------|----------|
| `/content/:id`     | foo      |
| `/r/delegator/:id` | bar      |


elocremarc:
@raphjaph @casey thoughts? We discussed this in an issue and on stream a few months back.

lifofifoX:
Love this! Wonder if something like `/content/:id/delegation` is a better endpoint. Seems a bit odd to have this be the only endpoint under `/r/` that won't return JSON.

elocremarc:
@lifofifoX I been thinking the same thing. Also yeah `/r/` is required to be JSON this breaks that convention. Plus they already cache `/content/` on cloudflare this would be a easy add. 

I just went with the route Casey and I discussed in the [initial issue.](https://github.com/ordinals/ord/issues/3700#issuecomment-2109271932)

casey:
Endpoints are under `r/` not because they return JSON, but because they are recursive, i.e., accessible to inscriptions.

src/subcommand/server.rs
```diff
@@ -1493,6 +1494,36 @@ impl Server {
...
+    task::block_in_place(|| {
+      if settings.is_hidden(inscription_id) {
+        return Ok(PreviewUnknownHtml.into_response());
+      }
+
+      let inscription = index
+        .get_inscription_by_id(inscription_id)?
+        .ok_or_not_found(|| format!("inscription {inscription_id}"))?;
+
+      if inscription.delegate().is_some() {
```
casey, line 1513: Why check if there is a delegate? We could remove this check and unconditionally return the content of the inscription instead.

docs/src/inscriptions/recursion.md
```diff
@@ -42,6 +42,7 @@ The recursive endpoints are:
 - `/r/children/<INSCRIPTION_ID>/<PAGE>`: the set of 100 child inscription ids on `<PAGE>`.
 - `/r/children/<INSCRIPTION_ID>/inscriptions`: details of the first 100 child inscriptions.
 - `/r/children/<INSCRIPTION_ID>/inscriptions/<PAGE>`: details of the set of 100 child inscriptions on `<PAGE>`.
+- `/r/delegator/<INSCRIPTION_ID>`: inscription content of the delegator.
```
casey, line 45: I'm not super happy about the name. The "delegator" is the whole inscription, whereas this just returns its content.  I'm having trouble coming up with a better name though.`r/original-content`, `r/own-content`, `r/undelegated`, and `r/undelegated-content` are all ideas.

src/subcommand/server.rs
elocremarc: So you mean return envelope regardless?

My thought was if you guys cache this on the CDN like `/content/` then it would be caching it twice? Why would we return redundant information if `/content/` already has it if no delegate is present?

casey: The CDN is free, so it's probably better to make the code and behavior simpler than try to save CDN resources.

docs/src/inscriptions/recursion.md
elocremarc: Yeah good point delegator would be the entire envelope which will include the delegate tag. 
We could maybe further define what the content is called after ` OP_PUSH 0` Maybe call it the `body` or `payload`

`/r/body/:id`
`/r/payload/:id`

elocremarc:
@casey @raphjaph made the changes. Other than what we call the route it should be gtg.

casey (approved):
LGTM

raphjaph (approved):

kingwingsss:
i come from X

PR #3936, opened 2024-09-05
Feature add indexed recursive inscriptions endpoint

### LINKED ISSUE #3923: Recursion: Get latest child from parent inscription
Currently you need to loop over all pages to get the latest child. 

To save multiple calls, would it be possible to adopt something similar to sat number, using `at/-1` to get latest...

`/r/children/<INSCRIPTION_ID>/inscriptions/at/-1`

twosatsmaxi:
Fixes #3923 

- [x] Added tests
- [x] Lint


twosatsmaxi:
@casey @raphjaph can someone review it?

src/subcommand/server.rs
```diff
@@ -232,6 +232,10 @@ impl Server {
           "/r/children/:inscription_id/inscriptions",
           get(Self::child_inscriptions_recursive),
         )
+        .route(
+          "/r/children/:inscription_id/inscriptions/at/:index",
```
elocremarc, line 236: We dont need `/inscriptions/` in the route it is too verbose and is plural. We are only returning a single Id here.
We should follow the format of the first child endpoint not the second one. I never got around to doing this feature on the original child endpoint I was going to but ran out of time.
```diff
-        "/r/children/:inscription_id/inscriptions/at/:index",
+        "/r/children/:inscription_id//at/:index",
```

src/subcommand/server.rs
```diff
@@ -6206,6 +6245,56 @@ next
...
+    });
+
+    server.mine_blocks(1);
+    let latest_child_inscription_id = InscriptionId { txid, index: 69 };
+    let child_json = server.get_json::<api::ChildInscriptionRecursive>(format!(
+      "/r/children/{parent_inscription_id}/inscriptions/at/-1"
+    ));
+    assert_eq!(child_json.id, latest_child_inscription_id);
+  }
+
```
elocremarc, line 6297: Need test for `/at/0` which would be the first child inscription. As well as `/at/1` which is the second child. This checks the off by one error we had in the original sat endpoint PR see discussion about that [here](https://github.com/ordinals/ord/pull/2452#issuecomment-1791815670)

Look at the sat endpoint for reference of `/at/n` tests:

elocremarc (changes requested):

elocremarc:
Upon further review I don't think there is any benefit of limiting ourself to using the Struct `ChildInscriptionRecursive` this is from the newer child endpoint which is a more optmized child endpoint that uses less onchain data. This struct was made for tons of pagination at the cost of data for speed. Considering we are only returning the info of a single inscription we can afford to return everything about that inscription and use the `InscriptionRecursive` struct from `/r/inscription`

elocremarc:
> Upon further review I don't think there is any benefit of limiting ourself to using the Struct ChildInscriptionRecursive

On second thought I think we should only return the ID. 
@casey @raphjaph do you agree? 
This is redundant data from `/r/inscription`.  Its useful on `/r/children/:inscription_id/inscriptions` because it returns 100 inscriptions with data. However this is only returning info about one inscription.

twosatsmaxi:
@raphjaph can you have a look once?

twosatsmaxi:
@casey @raphjaph: still waiting for the review.

src/subcommand/server.rs
twosatsmaxi: sounds good.

src/subcommand/server.rs
twosatsmaxi: will do.

wagedu:
> > Upon further review I don't think there is any benefit of limiting ourself to using the Struct ChildInscriptionRecursive
> 
> On second thought I think we should only return the ID. @casey @raphjaph do you agree? This is redundant data from `/r/inscription`. Its useful on `/r/children/:inscription_id/inscriptions` because it returns 100 inscriptions with data. However this is only returning info about one inscription.

I'm a bit confused with the lack of hype for `address` info in general. Considering it's the 1st thing listed right after the `ID`, on any inscription page...
I'd vote for the whole data in the 100 children. 
My current case as example:  4 inscriptions with 100 children each. But instead of 4 queries, it necessitates 400 queries, to compare addresses.

Deviltep (commented):
```

twosatsmaxi:
@raphjaph @casey can you have a look once and also what do you think about comments by @elocremarc? would like your inputs too before going ahead make those changes, makes sense too me.
The PR has been open since more than a month, would appreicate your reviews.

src/subcommand/server.rs
twosatsmaxi: done.

src/subcommand/server.rs
twosatsmaxi: done.

PR #3956, opened 2024-09-20
Start indexing at the appropriate block

partialord:
When using --index-sats or --index-addresses, we need to index all blocks, but --index-inscriptions only needs blocks starting from 767430, and --index-runes can start at 840000.  The existing code for this is tricky and also wrong -- it incorrectly skips early blocks when using --index-addresses but not --index-sats, and it unnecessarily processes blocks from 767430 to 840000 when using --index-runes but not --index-inscriptions.

Replace the old code with a single, correct calculation of which block to start at.  This speeds up building a runes-only index by about 4x.


casey:
LGTM!

PR #3959, opened 2024-09-21
Remove dependency on `ord-bitcoincore-rpc` crate

raphjaph:
We need to upgrade `rust-bitcoin`. In order to do that we need to get rid of the dependency on our fork of `bitcoincore_rpc`. This works because most of the reason we forked it was for the now separate `mockcore` crate.


casey (approved):
sure why not

PR #3963, opened 2024-09-23
More efficient sat-range handling

partialord:
Keep our arrays of sat ranges in the low-level `[u8; 11]` format, rather than converting them to `(u64, u64)` and back.  Also, don't mutate the input sat ranges in `index_transaction_sats()`; this avoids the need for a copy in `index_utxo_entries()` and also makes that code simpler.

Add a fast path for `Sat::common()`, which avoids a divide and some branch mispredictions.

These changes save almost 2.5 hours of CPU time.  The overall speedup is less, though, because indexing now runs fast enough that bitcoind often can't keep up -- we spend a significant amount of time just waiting for it to send us blocks.

### DIFF
--- a/crates/ordinals/src/sat.rs
+++ b/crates/ordinals/src/sat.rs
@@ -57,10 +57,16 @@ impl Sat {
     self.into()
   }
 
-  /// `Sat::rarity` is expensive and is called frequently when indexing.
-  /// Sat::is_common only checks if self is `Rarity::Common` but is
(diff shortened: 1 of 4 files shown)


casey:
LGTM! @raphjaph Want to take a look at this? This drops building a sat-only index from 10 hours to 8.5 hours.

raphjaph (approved):
LGTM!

PR #3968, opened 2024-09-30
Fix non-existant output lookup

### LINKED ISSUE #3919: Non-existant output lookups return 500 error
For example:

https://ordinals.com/output/0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef:123

This is because the JSON RPC status code / message for output not found was changed, so we don't correctly transform it into a 404 not found HTTP error.

raphjaph:
Fixes #3919

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -1577,7 +1577,7 @@ impl Index {
       }
     }
 
-    self.client.get_raw_transaction(&txid, None).into_option()
+    Ok(self.client.get_raw_transaction(&txid, None).ok())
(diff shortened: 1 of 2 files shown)


casey (changes requested):
This is incorrect, since it's just swallowing any error that bitcoind returns and turning it into a 404. We need to fix the `into_option` implementation. Either the error code or message changed between versions, so we need to check for the new error code / message.

casey (approved):

PR #3969, opened 2024-09-30
Deploy script: chown at correct spot

raphjaph:


casey:
Is this the right spot? I think when I ran it, `bitcoind --version` was what created the `settings.json` file with the wrong permissions, since it's running as the root user.

I think what we should actually do is instead is run bitcoind as the correct user, so the permissions don't get fucked up in the first place:

```
sudo -u bitcoin bitcoind -conf=/etc/bitcoin/bitcoin.conf --version
```

PR #3974, opened 2024-10-02
fix batch test error

rabbitprincess:
fix some compile error

### DIFF
--- a/src/wallet/batch/file.rs
+++ b/src/wallet/batch/file.rs
@@ -377,8 +377,8 @@ inscriptions:
         reinscribe: true,
         sat: None,
         satpoint: None,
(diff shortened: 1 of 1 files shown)


raphjaph:
What is the exact error you are encountering? 

The `master` branch compiles without any problems on my machine, a windows and a linux box.

rabbitprincess:
It seems to be an error with rust-analyzer, not rustc. Sorry for the misunderstanding.

> What is the exact error you are encountering?
> 
> The `master` branch compiles without any problems on my machine, a windows and a linux box.

It seems to be an error with rust-analyzer, not rustc. Sorry for the misunderstanding.

PR #3979, opened 2024-10-04
Add /address/:address/cardinals API

lifofifoX:
This would make it easier for apps to rely on the address index for retrieving cardinals UTXOs.


src/subcommand/server.rs
```diff
@@ -887,6 +888,56 @@ impl Server {
...
+  ) -> ServerResult {
+    task::block_in_place(|| {
+      Ok(if accept_json {
+        let address = address
+          .require_network(server_config.chain.network())
+          .map_err(|err| ServerError::BadRequest(err.to_string()))?;
+
+        let mut outputs = index.get_address_info(&address)?;
+
+        outputs.sort();
```
raphjaph, line 905: Do people rely on lexicographical ordering? Or should it be ordered by size?

tests/json_api.rs
```diff
@@ -362,6 +362,54 @@ fn get_output() {
...
+          .unwrap()
+          .assume_checked()
+      ),
+      spent: false,
+      transaction: send.txid,
+      value: 2 * COIN_VALUE,
+    }]
+  );
+}
+
```
raphjaph, line 412: Would be good to add a test that asserts that runic and inscription utxos are not returned.

raphjaph (changes requested):
Looks good so far! Just missing a test

src/subcommand/server.rs
lifofifoX: I wasn't sure either! But followed the precedence of `address`

https://github.com/ordinals/ord/blob/f1c1ee5adb39c54e99c79e65791b6339186c35be/src/subcommand/server.rs#L860

tests/json_api.rs
lifofifoX: Makes sense, will add them!

Do you think we should have `/address/:address/ordinals` endpoint for fetching just inscriptions and runes? I'm not sure about naming it `/ordinals` if it includes Runes, but it goes well with `/cardinals`.

raphjaph:
Let me know when this PR is ready for review

tests/json_api.rs
lifofifoX: Alright, I'll work on separate PRs.

lifofifoX:
@raphjaph Added an additional test to check inscriptions/runes are excluded in cardinals response. Should be ready now!

lifofifoX:
@raphjaph Added other two endpoints in #4008.

raphjaph:
Let me have a look at this

raphjaph:
I think we should go with something like this: https://github.com/ordinals/ord/pull/4056

What do you think?

raphjaph:
Close by #4056

PR #3980, opened 2024-10-04
Remove /runes/balances API endpoint

lifofifoX:
This endpoint returns every output containing any kind of a Rune. That doesn't seem very useful, and it's potentially a DDoS vector.


raphjaph (approved):
It was only active if the JSON API is enabled, but you're probably right. Makes sense to remove it. LGTM

@casey what do you think?

casey:
I'm fine with removing it, it's kind of a nutty endpoint. If someone complains that it's gone, we can always add it back.

PR #3984, opened 2024-10-06
Update JSON-API & Recursive documentation

### LINKED ISSUE #3742: Update docs to show new JSON API endpoints
Steps:

1. Enumerate all endpoints that return JSON (basically update this list here: https://docs.ordinals.com/guides/explorer.html#json-api)
2. Add short description 
3. Add an example `curl` command with JSON response.

We should make a new section under Guides and call it API and pull out the JSON API stuff from the Explorer section.

cryptoni9n:
fixes #3742

draft mode for now while I look for typos and test the links and menus

I'm not confident I have all of the endpoints captured here yet.  Please comment if you see any missing.


cryptoni9n:
I think this may be ready for a first review now.

raphjaph:
This looks great so far! 

I made some changes so that the examples can be expanded underneath. It's based on [this](https://gist.github.com/azagniotov/a4b16faf0febd12efbc6c3d7370383a6). If you could transform the other things into that format, that would be great!

raphjaph:
Yes, this looks great!

I've made some tweaks to the first two endpoints, if you could apply those to the rest as well. 

Also could you apply the same section to the Recursion section of the docs and update the examples as well? I think we're missing some. Thanks!

cryptoni9n:
> Yes, this looks great!
> 
> I've made some tweaks to the first two endpoints, if you could apply those to the rest as well.

sure, I've applied those changes throughout.

> Also could you apply the same template to the Recursion section of the docs and update the examples as well? I think we're missing some. Thanks!

ok - should I make a new PR or would you like it in this one?

cryptoni9n:
close/reopen to restart faulty tests

raphjaph:
> ok - should I make a new PR or would you like it in this one?

You can put it in this one

docs/src/SUMMARY.md
```diff
@@ -29,6 +29,7 @@ Summary
   - [Settings](guides/settings.md)
   - [Teleburning](guides/teleburning.md)
   - [Testing](guides/testing.md)
+  - [API](guides/api.md)
```
raphjaph, line 32: Can we put this as the first entry to the guides section?

docs/src/inscriptions/recursion.md
```diff
@@ -30,80 +30,124 @@ Recursion has a number of interesting use-cases:
...
+    <code>GET</code>
+    <code><b>/content/&lt;INSCRIPTION_ID&gt;</b></code>
+ </summary>
 
-- `/blockheight`: latest block height.
-- `/blockhash`: latest block hash.
-- `/blockhash/<HEIGHT>`: block hash at given block height.
-- `/blocktime`: UNIX time stamp of latest block.
+### Description
+the content of the inscription with `<INSCRIPTION_ID>`
```
raphjaph, line 43:
```suggestion
The content of the inscription with `<INSCRIPTION_ID>`
```

docs/src/inscriptions/recursion.md
```diff
@@ -30,80 +30,124 @@ Recursion has a number of interesting use-cases:
...
+<details>
+ <summary>
+    <code>GET</code>
+    <code><b>/r/blockhash/&lt;HEIGHT&gt;</b></code>
+ </summary>
 
-```json
-777000
+### Description
+block hash at given block height in text.
```
raphjaph, line 79:
```suggestion
Block hash at given block height in text.
```

raphjaph (changes requested):
Looking sweet! 

Could you capitalize the the first word in all of the descriptions in `recursion.md`.

While we're at it maybe also add the required flags for endpoint to be active just somewhere in the description. If it doesn't require any specific flags then it doesn't have to say anything. The basic config flags I can think about are `--index-addresses`, `--index-sats` and `--index-runes`.

docs/src/SUMMARY.md
cryptoni9n: sure thing

docs/src/inscriptions/recursion.md
cryptoni9n: will do

docs/src/inscriptions/recursion.md
cryptoni9n: will do

raphjaph (approved):
This looks sweet!

Made some small tweaks so that it include the recursive endpoints on the API page as well. Let me know if it looks good and then I'll merge.

cryptoni9n:
thanks Raph! I think it looks good, but when I do the rich diff display, the include doesn't look like it is rendering correctly.  Perhaps the viewer isn't capable?  Not sure...
image

Also, I saw one other thing that needed to be removed so I took it out.

raphjaph:
The include statement is for the `mdbook` builder. You can run it and display the docs with `just serve-docs`

PR #3986, opened 2024-10-07
Charms info recursive `/r/charms`

elocremarc:
Returns the currently supported charms and their associated names. 
Allows inscriptions to have forwards compatibility with any new charms added.

### DIFF
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -228,6 +228,40 @@ curl -s  \
 <details>
   <summary>
     <code>GET</code>
+    <code><b>/r/charms</b></code>
+  </summary>
+
+### Description
+Latest charms.
+
+### Example
+```bash
+curl -s  \
+  http://0.0.0.0:80/r/charms
+```
+```json
+{
+  "coin": "🪙",
+  "uncommon": "🌱",
+  "rare": "🧿",
+  "epic": "🪻",
+  "legendary": "🌝",
+  "mythic": "🎃",
+  "nineball": "9️⃣",
+  "reinscription": "♻️",
+  "cursed": "👹",
+  "unbound": "🔓",
+  "lost": "🤔",
+  "vindicated": "❤️‍🔥",
+  "burned": "🔥"
+}
+```
+</details>
+
+
+<details>
+  <summary>
+    <code>GET</code>
     <code><b>/r/children/&lt;INSCRIPTION_ID&gt;</b></code>
   </summary>
 
@@ -3332,6 +3366,7 @@ plain-text responses.
 - `/blockhash`: latest block hash.
 - `/blockhash/<HEIGHT>`: block hash at given block height.
 - `/blocktime`: UNIX time stamp of latest block.
+
 
 See
 [examples](examples.md#recursion) for on-chain examples of inscriptions that feature this functionality.
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -1,7 +1,8 @@
-# Recursion
+Recursion
+=========
 
 An important exception to [sandboxing](../inscriptions.md#sandboxing) is
-recursion. Recursive endpoints are whitelisted endpoints that allow access to1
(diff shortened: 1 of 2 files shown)


casey:
I think I understand the purpose of this PR, if we add charms, old inscriptions can still associate an emoji with new charms. However, I'm not sure I can actually think of way in which this would be useful. Can you give an example?

elocremarc:
@casey This is for a light ordinals explorer/viewer to traverse the family tree of an inscription. When you display the /r/inscription information you get the emojis but have to hardcode the updates. 

Another option what if ord did some light onchain libraries in leu of recursive endpoints?
This endpoint will have to be static anyway why not just do this onchain?

If this list changes ord could just reinscribe the new library sat. Then everyone can just subscribe to that sat get the new list. Then we don't need another endpoint.

Other useful stuff in a library you guys could inscribe. Could be converting between sat notation or other sat tools. We made the sat endpoint not take notation for simplicity but it would be nice to be able have a little library to use these notations onchain. People might not implement it correct etc... 

TBH I am kinda on the fence on if this PR is worth doing a recursive endpoint for.
I see both sides haha.

I will just inscribe this list myself and maintain it if I need to.

casey:
I think this is kind of reasonable, I don't hate it. The maps seems backwards to me though, it should map charm name to current emoji.

elocremarc:
@casey ok made the changes and added to the new docs.

src/subcommand/server.rs
```diff
@@ -1742,6 +1743,17 @@ impl Server {
     })
   }
 
+  async fn charms() -> ServerResult {
```
casey, line 1746: This can be:

```
  async fn charms() -> ServerResult<Json<BTreeMap<Charm, &'static str>>> {
```

`charm` implements serialize by delegating to `Display`, which is what is used in `to_string`.

docs/src/inscriptions/recursion.md
```diff
@@ -225,6 +225,40 @@ curl -s  \
...
 </details>
 
+<details>
+  <summary>
+    <code>GET</code>
+    <code><b>/r/charms</b></code>
+  </summary>
+
+### Description
+Latest charms.
```
casey, line 235:
```suggestion

Latest charms.
```

docs/src/inscriptions/recursion.md
```diff
@@ -225,6 +225,40 @@ curl -s  \
...
+  "cursed": "👹",
+  "unbound": "🔓",
+  "lost": "🤔",
+  "vindicated": "❤️‍🔥",
+  "burned": "🔥"
+}
+```
+</details>
+
+
```
casey, line 261: I'm not sure why we use `-s` and `-H "Accept: application/json"` for our curl examples. I don't think any of the recursive endpoints require the `Accept` header, and the `-s` will only suppress errors, which isn't useful.

```suggestion
### Example

```bash
curl http://0.0.0.0:80/r/charms
```

```json
{
  "coin": "🪙",
  "uncommon": "🌱",
  "rare": "🧿",
  "epic": "🪻",
  "legendary": "🌝",
  "mythic": "🎃",
  "nineball": "9️⃣",
  "reinscription": "♻️",
  "cursed": "👹",
  "unbound": "🔓",
  "lost": "🤔",
  "vindicated": "❤️‍🔥",
  "burned": "🔥"
}

```

</details>
```

File: src/subcommand/server.rs [hunk 4]
```diff
@@ -6237,6 +6249,33 @@ next
       .is_none());
   }
 
+  #[test]
+  fn charms_recursive() {
+    let server = TestServer::new();
+
+    let charms_response =
+      server.get_json::<serde_json::Map<String, serde_json::Value>>("/r/charms");
```
casey, line 6257:
```suggestion
    let charms_response =
      server.get_json::<BTreeMap<String, String>>("/r/charms");
```

casey: And then assert_eq against a BTreeMap built from the actual charms.

casey (changes requested):
Nice, see comments!

raphjaph:
This seems stale, please reopen if not

PR #3996, opened 2024-10-11
Add `deprecatedrpc=warnings` for ord to work with Bitcoin Core v28.0

gmart7t2:


raphjaph:
After updating `rust-bitcoin`, `ord` should work with v28 now. Could you check if it works for your setup?

raphjaph:
Now that we officially support v28 I will close this

PR #4002, opened 2024-10-18
Utxo signals

raphjaph:
- use CBOR instead of JSON
- [COSE](https://datatracker.ietf.org/doc/rfc9052/)
- [counter signing](https://datatracker.ietf.org/doc/html/rfc9338) with OTS
- use  cbor `bstr` for content (efficient nesting)
- base layer for protocols on top


raphjaph:
This is me just playing around with an idea. 

Basically you can send a piece of data into a network if you can sign for it with an existing UTXO. I think this could be an interesting primitive for bunch of different use cases. UTXOs could signal approval/disapproval for things, you could signal that you're willing to trade the UTXO, etc.

Vardominator:
This can be used for on-chain governance

lifofifoX:
Love this idea!!

If the signal stores current block height, it could be an indicator to know when something was signaled + also be utilized for rate limiting. i.e only allow signaling once every X blocks.

raphjaph:
> Love this idea!!
> 
> If the signal stores current block height, it could be an indicator to know when something was signaled + also be utilized for rate limiting. i.e only allow signaling once every X blocks.

Yes, that's one of the ways to rate limit. Since it's UTXO based you can set a plethora of options for rate limiting.

casey:

PR #4005, opened 2024-10-20
Add `wallet addresses`

raphjaph:


src/subcommand/wallet/addresses.rs
```diff
@@ -0,0 +1,75 @@
...
+  #[serde(skip_serializing_if = "Option::is_none")]
+  pub inscriptions: Option<Vec<InscriptionId>>,
+  #[serde(skip_serializing_if = "Option::is_none")]
+  pub runes: Option<BTreeMap<SpacedRune, Decimal>>,
+}
+
+#[derive(Debug, Parser)]
+pub(crate) struct Addresses {
+  #[arg(short, long, help = "Only show a list of used addresses.")]
+  compact: bool,
```
casey, line 16: I think we can remove this. We have a `--compact` flag for `ord decode`, but that's because the output of decode is crazy long by default, but the output of this command is pretty reasonable.

casey (approved):
LGTM

PR #4008, opened 2024-10-21
Add /address/:address/ JSON API endpoints

lifofifoX:
This is branched off #3979 

In addition to the `/address/:address/cardinals` endpoint, this PR adds `/address/:address/runes` and `/address/:address/inscriptions` endpoints.

This PR also groups all the `/address/:address/*` tests. That way, we don't need a dedicated test to check that inscriptions/runes are excluded from `/address/:address/cardinals`.


lifofifoX:
@raphjaph Any thoughts on this?

raphjaph:
I think we should go with something like this: https://github.com/ordinals/ord/pull/4056

What do you think?

raphjaph:
Closed by #4056

PR #4009, opened 2024-10-22
Show total child count

arik-so:
Add a child totalizer to the inscription page.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -93,6 +93,7 @@ pub struct ChildInscriptions {
 pub struct Inscription {
   pub address: Option<String>,
   pub charms: Vec<Charm>,
+  pub child_count: u64,
   pub children: Vec<InscriptionId>,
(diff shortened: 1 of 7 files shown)


casey:
We're discussing how to display the number of children on the /inscription page, and not coming to an obvious conclusion. Do you want this PR so that the number of children is displayed, or so that the number of children is returned from the API? If you don't care that the number of children is displayed on the /inscription page, we could remove that from this PR and worry about it later.

casey:
Gotcha. We don't really like modifying the "children" `<dt>` So maybe it could go in the `all` link instead? I.e. `all 4` or `all 4 children`.

raphjaph:
> Come to think of it though, it might be prudent for the /children page to also show the total somewhere.

That makes sense. Let's put it in a separate PR though

raphjaph (approved):
LGTM

PR #4020, opened 2024-10-25
Group indices in status response object

arik-so:
Minor, but admittedly subjective, legibility improvement.


raphjaph:
Thanks for the PR but I think we like it the way it is. Doesn't seem like a huge advantage to us.

Issue #4021, opened 2024-10-25
Satoshi Endpoint Proxy


arik-so:
The `server --proxy` option is helpful for testing mainnet inscriptions' behavior locally, but a lot of inscriptions that require such testing rely on a specific dynamically inscribed ordinal, which is usually uninscribed locally, and breaks testability.

I wonder whether allowing the `--proxy` option to also apply as a fallback for the `/r/sat/:sat/at/:index` endpoint might be worth pursuing? That should help with resolving dynamic content.

raphjaph:
Yes, this is a good idea. I even have an issue for this https://github.com/ordinals/ord/issues/3806

PR #4022, opened 2024-10-25
Proxy recursive satoshi endpoint

arik-so:
WIP: Builds on #4204.

In case people find any use in #4021, this provides a draft implementation.


src/subcommand/server.rs
```diff
@@ -534,6 +520,45 @@ impl Server {
...
+    let status = response.status();
+
+    if let Some(proxy) = server_config.proxy.as_ref() {
+      if status == StatusCode::NOT_FOUND {
+        return task::block_in_place(|| Server::proxy(proxy, &path));
+      }
+
+      // This is a workaround for the fact the the /r/sat/<SAT_NUMBER>/at/<INDEX>
+      // does not return a 404 when no inscription is present on the sat.
+      if path.starts_with("r/sat") && path.contains("/at/") {
```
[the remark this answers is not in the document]
raphjaph: i pushed something but no idea if it's better. A local regtest index proxying stuff should be able to also have a sat index

src/subcommand/server.rs
```diff
@@ -536,6 +536,27 @@ impl Server {
       if status == StatusCode::NOT_FOUND {
         return task::block_in_place(|| Server::proxy(proxy, &path));
       }
+
+      // This is a workaround for the fact the the /r/sat/<SAT_NUMBER>/at/<INDEX>
+      // does not return a 404 when no inscription is present on the sat.
+      if matches!(
```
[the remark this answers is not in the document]
raphjaph: I agree it's hacky but I don't think we can use axum stuff inside here.

src/subcommand/server.rs
```diff
@@ -283,6 +282,7 @@ impl Server {
         )
         .route("/r/inscription/{inscription_id}", get(r::inscription))
         .route("/r/metadata/{inscription_id}", get(r::metadata))
+        .route("/r/sat/{sat_number}/at/{index}", get(r::sat_at_index))
```
casey, line 285: I think this should be moved back, so it's sorted.

casey: We try to keep these sorted, so `r/sat/{sat_number}/at/{index}/content` should be moved up.

src/subcommand/server.rs
```diff
@@ -537,6 +537,27 @@ impl Server {
       if status == StatusCode::NOT_FOUND {
         return task::block_in_place(|| Server::proxy(proxy, &path));
       }
+
+      // This is a workaround for the fact the the /r/sat/<SAT_NUMBER>/at/<INDEX>
+      // does not return a 404 when no inscription is present on the sat.
+      if matches!(
+        path.split('/').collect::<Vec<&str>>().as_slice(),
```
casey, line 544: Let's do a regex match instead, this is pretty heinous.

casey (changes requested):
Couple of comments, otherwise LGTM.

src/subcommand/server.rs
casey: I always just do `:sort` in vim and use whatever order that uses.

raphjaph (approved):
LGTM! @casey you wanna have a look at it one more time as well?

src/subcommand/server.rs
```diff
@@ -545,6 +545,25 @@ impl Server {
       if status == StatusCode::NOT_FOUND {
         return task::block_in_place(|| Server::proxy(proxy, &path));
       }
+
+      // This is a workaround for the fact the the /r/sat/<SAT_NUMBER>/at/<INDEX>
+      // does not return a 404 when no inscription is present on the sat.
+      let regex = Regex::new(r"(?m)r\/sat\/\d+\/at\/-?\d+$").unwrap();
```
casey, line 551: This regex should be compiled once and reused using `lazy_static`

src/subcommand/server.rs
casey: Ah, sorry, I wasn't looking at the whole diff and didn't see that proxiable_routes had been moved into a separate section.

casey (changes requested):
Minor comment, just about reusing the regex.

casey (approved):
LGTM!

Made some small changes:

- Forward slashes don't need to be escaped in regexes. (They do in Vim, but that's because `/` is often used at the pattern delimiter.)
- I think multiline mode `(?m)` was unnecessary.
- I pass the path through without removing the leading forward slash, and instead remove it in the `proxy` function, since I think that's cleaner.

PR #4026, opened 2024-10-26
BIP322 sign file

raphjaph:

PR #4027, opened 2024-10-26
Sign for inscription and output

raphjaph:
Depends on #4026


src/subcommand/wallet/sign.rs
```diff
@@ -6,19 +6,33 @@ use {
...
+    .args(&["message", "file"])),
+group(
+  ArgGroup::new("signer")
+    .required(true)
+    .args(&["address", "inscription", "output"]))
 )]
 pub(crate) struct Sign {
   #[arg(long, help = "Sign for <ADDRESS>.")]
-  address: Address<NetworkUnchecked>,
+  address: Option<Address<NetworkUnchecked>>,
```
casey, line 31: Let's combine these into a `Signer` enum. We could either have this as an argument, or maybe a `--sign-with` flag.

src/subcommand/wallet/sign.rs
```diff
@@ -6,19 +6,33 @@ use {
...
+  #[arg(long, help = "Sign for <INSCRIPTION>.")]
+  inscription: Option<InscriptionId>,
+  #[arg(long, help = "Sign for <UTXO>.")]
+  output: Option<OutPoint>,
   #[arg(long, help = "Sign <MESSAGE>.")]
   message: Option<String>,
```
casey, line 37: Maybe call this `--text` since there's always a message, whether it comes on the command line or from a file.

src/subcommand/wallet/sign.rs
```diff
@@ -64,6 +98,8 @@ impl Sign {
 
     Ok(Some(Box::new(Output {
       address: address.as_unchecked().clone(),
+      inscription: self.inscription,
+      output: self.output,
       message: self.message.clone(),
```
casey, line 103: Let's remove this, since there's always a message that's being signed, but it would be unreasonable to include a very large file.

casey (changes requested):

casey (approved):
LGTM

PR #4032, opened 2024-10-27
Un-pin redb dependency and update to 2.2.0

casey:
Redb issues possibly fixed! See https://github.com/cberner/redb/issues/879

Testing on charlie now.


casey:
Full sync successful, and took 20 hours instead of 21 hours on our test server.

raphjaph (approved):

PR #4041, opened 2024-11-03
Rescan wallet when restoring from mnemonic 

### LINKED ISSUE #1589: Trigger wallet rescan on wallet import/restore
One user [reported](https://gist.github.com/windsok/5b53a1ced6ef3eddbde260337de28980?permalink_comment_id=4463824#gistcomment-4463824) that after importing a wallet they originally created with Sparrow, `ord` was not showing any of the inscriptions they had received to the wallet. To fix the issue they initiated a manual rescan with `rescanblockchain` using `bitcoin-cli`.

Not sure if they did something weird when restoring the wallet, or if we need to add some logic to automatically trigger a rescan in bitcoin core when wallets are imported.

casey:
Finally fixes #1589.

### DIFF
--- a/src/subcommand/wallet/restore.rs
+++ b/src/subcommand/wallet/restore.rs
@@ -45,6 +45,7 @@ impl Restore {
           name,
           settings,
           mnemonic.to_seed(self.passphrase.unwrap_or_default()),
+          bitcoincore_rpc::json::Timestamp::Time(0),
         )?;
       }
     }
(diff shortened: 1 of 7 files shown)


src/subcommand/wallet/restore.rs
```diff
@@ -45,6 +45,7 @@ impl Restore {
           name,
           settings,
           mnemonic.to_seed(self.passphrase.unwrap_or_default()),
+          bitcoincore_rpc::json::Timestamp::Time(0),
```
raphjaph, line 48: Should we allow setting the timestamp for mnemonic restoration?

casey: I think we should, but this PR is an improvement, so I'd like to merge it right away and create a new issue. There are potentially considerations like should the user be able to set a date instead of a unix timestamp, etc.

Issue #3990, opened 2024-10-09
ordinals.com is now drawing a scrollbar next to each inscription


gmart7t2:
![Screenshot_2024-10-09_04-03-33](https://github.com/user-attachments/assets/40fbd85e-46c6-4c10-a75d-cb4788bbcf59)

gmart7t2:
On Android, too:
![Screenshot_20241009_040555_Presearch](https://github.com/user-attachments/assets/a29d18ec-78cb-49a7-9300-a293ec4c92b9)

gmart7t2:
It looks like this was introduced by #3947 removing the `scrolling=no`.

gmart7t2:
Size by side comparison of 0.20.0 and 0.20.1:
![Screenshot_2024-10-09_04-23-28](https://github.com/user-attachments/assets/1ad4827b-640c-466b-bf62-bf3c82a695ef)

casey:
I landed a partial fix for this in #4042. The image preview is now fixed, so that the images don't overflow. This should remove all scrollbars on image inscriptions. I also removed scrollbars entirely from inscription thumbnails. This means that the only scrollbars remaining are on non-image inscriptions on /inscription which overflow their bounds. I'm not entirely satisfied with that, since I'm not sure we should have scrollbars at all, especially because they're so ugly on Windows, so I'm going to leave this image open so people have a place to complain about it.

gmart7t2:
I still see scrollbars on some thumbnails even with scrolling=no on the iframe:

![Screenshot_2024-11-04_09-47-31](https://github.com/user-attachments/assets/304274e9-6977-41d3-b182-6f1df8d14090)

casey:
Are those on the iframe element itself, or on an interior element of an HTML inscription? I suppose an HTML inscription could have interior content which overflowed the size it's rendered at. But if that's the case, I'm not sure there's much we can do about it, since we don't want to start monkeying with the contents of HTML inscriptions.

casey:
I wrote up the situation in #4048.

PR #4042, opened 2024-11-03
Hide image preview and thumbnail scrollbars

casey:
In #3947, we allowed scrolling in iframes. I recently visited ordinals.com on windows, and there were hideous scrollbars on nearly every inscription.

This isn't a windows-specific issue, i.e., the scrollbars are there on macOS, they just look worse on Windows, which is why we didn't notice.

This PR does a couple things:

- Add `float: left` to images in the image preview. This, for some reason, makes the images the correct size, so they don't overflow, due to some mysterious CSS behavior that I don't fully understand.
- Add back `scrolling=no` to inscription thumbnails. This prevents scrollbars for appearing everywhere except the actual /inscription page, which I think is better. There's no reason to allow scrolling a thumbnail, especially if the user can just click it to scroll.

I'm still not sure that we shouldn't just re-add `scrolling=no` to inscriptions. Users can click the content link if they want to scroll, and the bars on windows are absolutely disgusting looking. If you made an HTML inscription, previewed it on an old version of ord, and then inscribed it on mainnet, you're now getting scrollbars where before you didn't see any.

However, this PR is a pareto improvement, unlikely removing scrollbars entirely, so I think we should merge this and if the remaining scroll bars are an issue, deal with them later.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -3947,7 +3947,7 @@ mod tests {
 <dl>
   <dt>inscriptions</dt>
   <dd class=thumbnails>
-    <a href=/inscription/.*><iframe sandbox=allow-scripts loading=lazy src=/preview/.*></iframe></a>
+    <a href=/inscription/.*><iframe sandbox=allow-scripts scrolling=no loading=lazy src=/preview/.*></iframe></a>
   </dd>.*",
     );
   }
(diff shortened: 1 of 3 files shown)


raphjaph (approved):
LGTM

raphjaph:
Fixes #3990

gmart7t2:
@casey 

> This prevents scrollbars for appearing everywhere except the actual /inscription page

It doesn't. See this example: it has scrolling=no but the thumbnail still displays the scrollbar: https://without21.ordstuff.info/block/868826

Screenshot_2024-11-04_09-47-31

PR #4045, opened 2024-11-04
Allow including metadata when burning inscriptions

casey:
Allow including CBOR metadata, from a CBOR or JSON source, when burning inscriptions.

If the OP_RETURN burning an inscription includes a data push immediately after the OP_RETURN, deserialize it as CBOR and display it on the inscription page below the inscription.

Because OP_RETURNs are not executed, there is no need to split metadata into 520 chunks, so we only consider the first data push after the OP_RETURN. This leaves additional data pushes free for some future use case.

I think we should consider whether or not we want to do this. I like the idea of being able to include a message when burning an inscription. But, it does modify the way an inscription appears on /inscription. However, you can already effectively add images by adding a child inscription, so I think this doesn't really make a difference.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,7 +273,7 @@ checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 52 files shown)


raphjaph (approved):
LGTM

PR #4047, opened 2024-11-04
Add a metaprotocol field to the api::Inscription struct to display in the server’s JSON output

kbehouse:
Add a way to fetch metaprotocol using server api JSON output


raphjaph (approved):
LGTM

switch-900:
This would be great as an endpoint.

PR #4050, opened 2024-11-05
Output recursive

1stBitcoinSent:
Add `/r/output/ID` recursive endpoint.

😁😅 many 1sts for me here so please go easy on me.


1stBitcoinSent:
😀 Hi @casey Hello @raphjaph I'm not sure how to run these checks 😅

1stBitcoinSent:
test example using output sat_ranges
1stBitcoinSent test example

thinking 🤔 it would also be interesting to be able to represent the history of sats artistically with recursive endpoint access to transactions

raphjaph:
How would inscriptions get the outpoint (other than hardcoding it)?

1stBitcoinSent:
> How would inscriptions get the outpoint (other than hardcoding it)?

Hello 😁 @raphjaph 
This is my current workflow:
`const getInscriptionId = () => window.location.pathname.split("/").pop();`
`let response = await fetch(`/r/inscription/${getInscriptionId()}`);`
`let inscriptionData = await response.json();`
`inscriptionData.output`

raphjaph:
This definitely needs some tests though as well.

1stBitcoinSent:
> This definitely needs some tests though as well.

Hey @raphjaph 😁 Thank you for getting back to me on this 😀 is this ok?

raphjaph:
I'm looking at this today, just added some structs and gonna figure out what exactly to return

1stBitcoinSent:
> I'm looking at this today, just added some structs and gonna figure out what exactly to return

Thank you so much raph 🧡😆

docs/src/inscriptions/recursion.md
```diff
@@ -3202,6 +3202,37 @@ curl -s -H "Accept: application/json" \
...
+Information about an output.
+
+### Example
+```bash
+curl -s -H "Accept: application/json" \
+  http://0.0.0.0:80/r/output/4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b:0
+```
+```json
+{
+  "address": "04678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5f",
```
raphjaph, line 3222: This is not an address.

docs/src/inscriptions/recursion.md
```diff
@@ -3202,6 +3202,37 @@ curl -s -H "Accept: application/json" \
...
+  "indexed": true,
+  "value": 5000000000,
+  "inscriptions": [],
+  "runes": {},
```
raphjaph, line 3226: we should find an example that includes some runes

docs/src/inscriptions/recursion.md
1stBitcoinSent: yea it's strange 🤔
[https://mempool.space/tx/4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b#flow=&vout=0](https://mempool.space/tx/4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b#flow=&vout=0)

docs/src/inscriptions/recursion.md
1stBitcoinSent: do you know of an output that should never change like the genesis output?
maybe a burnt rune output? would that still show a rune amount if it was burnt?

raphjaph:
Hey, thanks for this PR but we decided to go with a slightly different approach in #4148. It was just too different to continue work here so I made a new PR.

PR #4055, opened 2024-11-07
Identify collapsible nodes with class=collapse

casey:
Instead of identifying collapsible with `class=collapse`, instead of with a regular expression.

This is faster, simpler, and less error prone. Going forward, any elements which should be collapsable should get this class. Things like addresses, any kind of hash, outpoints, and satpoints, should be collapsable.


raphjaph (approved):
LGTM

PR #4058, opened 2024-11-09
Add assert_html function

casey:
This PR adds an `assert_html` test method which takes a path and a value implementing `PageContent`, and asserts that the page returned from the server is the same as that from rendering the value as an HTML template.

I think this is a dramatic improvement over our HTML / regex-based tests, which have terrible diffs and are super verbose, and basically force the test to recreate all the formatting of templates.

I only converted three instances of the old test type to use `assert_html`. I think going forward, we should change to `assert_html` whenever a test breaks.


raphjaph (approved):
banger

PR #4063, opened 2024-11-10
Only burn one sat

### LINKED ISSUE #3977: Burn commmand should always burn one sat
Since the burn command is for burning inscriptions, it should default to burning the minimum amount of BTC possible to burn an inscription, i.e., one sat. (I think OP_RETURN outputs are not required to be above the dust limit.)

Also, I think `--postage` option should be removed, since I can't think of a good reason to specify some other amount of postage, especially because we forbid large values for postage, and the only reason to have customizable postage would be to burn large amounts of BTC along with the inscription.

onchainguy-btc:
Resolves #3977 

Tbh not sure if it's actually that simple. Tested with [this inscription](https://ordinals.com/inscription/fb2de94a4b1c30dbf9c228dec2971ae1bc88ffb3479ae59ecb981f1f65add1eci0). And this was the [burn tx](https://mempool.space/tx/97ad0dae72ab33fa0e64cc573d3b54d19a282c2e1af6bb0ac326f41d6c67c503).

### DIFF
--- a/src/subcommand/wallet/burn.rs
+++ b/src/subcommand/wallet/burn.rs
@@ -28,12 +28,6 @@ pub struct Burn {
     you understand the implications."
   )]
   no_limit: bool,
-  #[arg(
-    long,
-    help = "Target <AMOUNT> postage with sent inscriptions. [default: 10000 sat]",
-    value_name = "AMOUNT"
-  )]
-  postage: Option<Amount>,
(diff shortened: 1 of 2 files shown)


casey (approved):
LGTM

raphjaph (approved):
LGTM

PR #4065, opened 2024-11-11
Add a timestamp to restore

### LINKED ISSUE #4043: Allow specifying wallet rescan timestamp when restoring
We recently fixed #1589, which triggers a wallet rescan when a wallet is restored from a descriptor. However, rescans are expensive. We should allow the user to optinally provide a timestamp or date when restoring, in order to avoid rescanning the entire chain.

Maybe something like:

```
ord wallet restore --timestamp N
```

Where N is either a unix timestamp, or a YYYY-MM-DD date. If the user provides a date, we should round it to midnight the previous day, to avoid issues with time zones.

raphjaph:
Fixes #4043

### DIFF
--- a/src/subcommand/wallet/restore.rs
+++ b/src/subcommand/wallet/restore.rs
@@ -33,10 +33,17 @@ impl Restore {
     match self.from {
       Source::Descriptor => {
         io::stdin().read_to_string(&mut buffer)?;
+
         ensure!(
           self.passphrase.is_none(),
           "descriptor does not take a passphrase"
         );
+
+        ensure!(
+          self.timestamp.is_none(),
+          "descriptor does not take a timestamp"
+        );
+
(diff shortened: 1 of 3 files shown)


raphjaph:
@gmart7t2 I think this should fix #4043. Let me know if I'm missing something.

We now import both descriptors in one RPC call, so timeouts shouldn't brick the wallet. We had to use the manual call to `importdescriptors` since the `bitcoincore_rpc` crate doesn't support setting multiple descriptors at once for some reason. @casey has a [PR](https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/378) open there to fix that though.

gmart7t2:
@raphjaph it used to be that if there was a timeout while rescanning, the seed phrase wouldn't be displayed (I'm think more about "wallet create" I guess). Is that still the case?

raphjaph:
#4095

PR #4075, opened 2024-11-14
Fix off-by-one-error in sat range template

casey:
Sat ranges are no longer intervals, and instead link to the first and last sat in the range. However, the last sat is off by one. Fix that.

### DIFF
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -3936,7 +3936,7 @@ mod tests {
 </dl>
 <h2>1 Sat Range</h2>
 <ul class=monospace>
-  <li><a href=/sat/0 class=mythic>0</a>-<a href=/sat/5000000000 class=uncommon>5000000000</a> \\(5000000000 sats\\)</li>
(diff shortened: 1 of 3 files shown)


casey:
@arik-so I think that's a separate issue. The HTML is straightforwardly wrong, since we link directly to the sat which is the end of the interval, and this fixes that.

I think the JSON API is kind of a separate issue, since I think there there's more of an argument that it's a programming interface, and so you want half-open ranges.

raphjaph (approved):
LGTM

gmart7t2:
This is pretty confusing. We have had the sat ranges being shown as half open intervals since launch and now the meaning of the sat range has been changed to be a closed interval with no visual indication that the meaning has changed.

This seems dangerous. At this point, people are used to the ranges being half open.

I would suggest not using the same syntax for half open ranges and closed ranges.

casey:
@gmart7t2 I agree this sucks. However, I think we were in an unfortunately situation where people who are used to half-open ranges knew what the template meant, and people who weren't were confused, so I'm not sure this change is a negative. I'm also not sure how we could change the notation without making it worse.

PR #4083, opened 2024-11-16
Add replicate and swap recipes

casey:
`just replicate` copies indices from `charlie` to `alpha` and `bravo`

`just swap HOST` swaps in the new index on the given host.


raphjaph (approved):
LGTM

PR #4084, opened 2024-11-17
Enable redb's new quick-repair mode

partialord:
This enables redb's new quick-repair mode so you can try it out.  It's using redb master, since this version hasn't been released yet.

Recovery after a crash always means rolling back to the last complete commit; if that commit was made with quick-repair enabled, then recovery will be instant.  So to get the new behavior, you'll need to let it make one successful commit with the new code before you start making it crash.  After that, no matter what you do, you should never have to wait for repair again.  You can tell it's working because it won't print the "Index file needs recovery" message, even after a crash; it'll just look like a normal clean startup.

Quick-repair mode needs to save the database allocator state with each commit, which costs about 200 ms per commit for a 190 GB database (it scales linearly with file size).  My guess is that's totally fine, so for now quick-repair is unconditionally enabled for all writes to the database.

If you try this out, let me know how it goes!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2904,8 +2904,9 @@ dependencies = [
 
 [[package]]
 name = "redb"
-version = "2.2.0"
-source = "git+https://github.com/cberner/redb?rev=c66447a6530d62615c5b377b3dca05799ab03efd#c66447a6530d62615c5b377b3dca05799ab03efd"
+version = "2.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7c2a94325f9c5826b17c42af11067230f503747f870117a28180e85696e21ba"
 dependencies = [
  "libc",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2905,8 +2905,7 @@ dependencies = [
 [[package]]
 name = "redb"
 version = "2.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


casey:
Very nice! I just deployed it to charlie for testing.

casey:
Just tried it out by `kill -9`'ing ord a bunch. Works like a charm!

Luigi-1Combo:
It is best to add a parameter "enable-quick-repair" to ord, which can be turned off when synchronizing from the beginning to increase the speed, and can be turned on after synchronizing to the latest height.

raphjaph (approved):

casey:
> It is best to add a parameter "enable-quick-repair" to ord, which can be turned off when synchronizing from the beginning to increase the speed, and can be turned on after synchronizing to the latest height.

We could add that, but only if enabling quick repair is significantly slower, which it might not be, if it's a per-commit overhead, and our commits are large and slow. So we'd need to benchmark to know the speed difference.

casey:
@partialord reports that the overhead is pretty small: 4ms per commit for a database under 4 GB, and 200ms per commit for a 190 GB database. Building with the default options, total overhead should be about 1.5 seconds, and for a full index (sats, runes, etc) it would be a about a minute, which doesn't seem like enough to justify adding a new flag.

twosatsmaxi:
this is good, nice. Thanks @partialord. Gone are the days where it takes hours to repair.

cryptoni9n:
love to see major improvements like this - big thanks to @partialord and the rest of the team!

PR #4085, opened 2024-11-17
Make build script public

arronzhang:
I need this method when build a custom wallet. Thanks.


raphjaph (approved):
Seems fine to me!

raphjaph:
Have you tried out the `--index-address` flag? I'm hoping that `ord` could replace the electrum dependency for many people's infra. We're still missing mempool support but other than that I'd be interested to hear if it would meet your requirements.

We have the `/address/<ADDRESS>` and `/outputs/<ADDRESS>` endpoints.

arronzhang:
Yes, I want to mint more than one in a block, so I need the outputs by address in the mempool. Include all outputs spent via CPFP, in order to calculate the best fee rate for my next mint. Additionally a transaction broadcast api is needed. That's all for now.

PR #4088, opened 2024-11-20
Add `/r/parents/:inscription_id/inscriptions` endpoint.

gmart7t2:
Analogous to #3771 but for parents instead of children.


gmart7t2:
The examples in the docs for /r/children/.../inscription* have the full output, so I did the same here. 

Maybe it makes sense to shorten those examples, too?

src/subcommand/server.rs
```diff
@@ -2225,6 +2234,63 @@ impl Server {
...
+        .ok_or_not_found(|| format!("inscription {child}"))?;
+
+      let (ids, more) = index.get_parents_by_sequence_number_paginated(entry.parents, 100, page)?;
+
+      let parents = ids
+        .into_iter()
+        .map(|inscription_id| {
+          let entry = index
+            .get_inscription_entry(inscription_id)
+            .unwrap()
```
casey, line 2260: These errors should be handled. If the result type of a map is `Result<T>`, then you can collect into a `Result<Vec<T>>`, which will return an error if any of the results is `Result::Err`.

src/subcommand/server.rs
```diff
@@ -2225,6 +2234,63 @@ impl Server {
...
+            .unwrap();
+
+          let satpoint = index
+            .get_inscription_satpoint_by_id(inscription_id)
+            .ok()
```
casey, line 2265: Same here, this error should be handled.

casey (changes requested):
Looks good! The errors in `parent_inscriptions_recursive` should be handled, see my comment. In general, errors from the index shouldn't be ignored. `unwraps` on `Option`s are okay though, when we expect the index to contain the inscription.

casey:
Can you rebase or merge master? This has some conflicts.

cryptoni9n:
> Can you rebase or merge master? This has some conflicts.

rebased and ready for review

docs/src/inscriptions/recursion.md
```diff
@@ -542,2579 +542,45 @@ curl -s \
       "satpoint": "0648b2a2d6dc6c882a4c61f86f1e2e4354eeafebad96e319dc98510f1bc4f2cd:1:0",
       "timestamp": 1726282054
     },
```
casey, line 544: Can you put these deletions into a follow up PR? The change here which combines a very large deletion with some additions makes this harder to review.

src/api.rs
```diff
@@ -144,6 +151,19 @@ pub struct ChildInscriptionRecursive {
   pub timestamp: i64,
 }
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub struct RelativeInscriptionRecursive {
```
casey, line 155: Should `ChildInscriptionRecursive` be deleted in favor of this struct?

src/subcommand/server/r.rs
```diff
@@ -430,6 +430,62 @@ pub(super) async fn parents(
   parents_paginated(Extension(index), Path((inscription_id, 0))).await
 }
 
+pub async fn parent_inscriptions_recursive(
```
casey, line 433: Since we now have a dedicated module for recursive andpoints, `_recursive` isn't needed anymore:

```suggestion
pub async fn parent_inscription(
```

src/subcommand/server/r.rs
```diff
@@ -430,6 +430,62 @@ pub(super) async fn parents(
...
+  Extension(index): Extension<Arc<Index>>,
+  Path(inscription_id): Path<InscriptionId>,
+) -> ServerResult {
+  parent_inscriptions_recursive_paginated(Extension(index), Path((inscription_id, 0))).await
+}
+
+pub async fn parent_inscriptions_recursive_paginated(
```
casey, line 440: Since we now have a dedicated module for recursive andpoints, `_recursive` isn't needed anymore:

```suggestion
pub async fn parent_inscriptions_paginated(
```

src/subcommand/server/r.rs
```diff
@@ -430,6 +430,62 @@ pub(super) async fn parents(
...
+      .get_inscription_entry(child)?
+      .ok_or_not_found(|| format!("inscription {child}"))?;
+
+    let (ids, more) = index.get_parents_by_sequence_number_paginated(entry.parents, 100, page)?;
+
+    let parents = ids
+      .into_iter()
+      .map(
+        |inscription_id| -> ServerResult<api::RelativeInscriptionRecursive> {
+          let entry = index
```
casey, line 455: This code looks identical to that in `children_inscriptions_paginated`, you can define a function which contains the code which takes the index and goes from an inscription ID to a `api::RelativeInscriptionRecursive`, and call it in both places.

casey (changes requested):
Looks mostly good to me! See comments.

docs/src/inscriptions/recursion.md
cryptoni9n: I've backed up the deletion commit to a local branch and rebased; will submit a separate PR for those changes.

src/subcommand/server/r.rs
cryptoni9n: updated

src/subcommand/server/r.rs
cryptoni9n: updated

src/subcommand/server/r.rs
cryptoni9n: added `fn to_relative_inscription` to reduce duplication

src/api.rs
cryptoni9n: I couldn't see any reason why not.  It's not being used anywhere else in the code as far as I can see.  I've taken it out in the latest commit.

raphjaph (approved):
LGTM

PR #4093, opened 2024-11-21
Do not display teleburn address

raphjaph:
Nobody uses this and it confuses people. This PR just doesn't display the address on the frontend, all other teleburn logic remains.


casey (changes requested):
Looks good but tests need fixing. Any changed tests should use the new style.

rot13maxi:
end of an era. 

good change tho.

casey:
Yah. I think the main thing is that there's too much stuff on /inscription. The ethereum teleburn address is only the first victim.

Issue #4091, opened 2024-11-20
Display mnemonic on `ord wallet create` even if import descriptors times out


casey:
Reported by @gmart7t2 here:

> @raphjaph it used to be that if there was a timeout while rescanning, the seed phrase wouldn't be displayed (I'm think more about "wallet create" I guess). Is that still the case?
>
> In case this isn't clear:
>
> When I create a new ord wallet, ord imports new descriptors, bitcoind rescans the last 2 hours of blocks, which might cause an RPC timeout, which aborts the ord command and prevents the JSON result (which contains the new seed phrase) from being displayed.

If `bitcoind` times out, the descriptors can be imported, but we'll never have printed out the mnemonic. We should fix this somehow. I think we should print the same output to standard out, but *also* return an error code and print an error message to stderr.

PR #4095, opened 2024-11-21
Wait for wallet to load

raphjaph:
Addresses #4091 and #4111.

@gmart7t2 This catches a couple of errors that usually occur when the wallet takes a long time to load. It should now wait until a wallet or descriptor is finished loading before proceeding. 

Do you know by chance what the Bitcoin Core error code or message is if a descriptor for a wallet is still loading?

### DIFF
--- a/src/wallet.rs
+++ b/src/wallet.rs
@@ -555,36 +555,29 @@ impl Wallet {
       });
     }
 
-    loop {
-      match settings
-        .bitcoin_rpc_client(Some(name.clone()))?
-        .call::<serde_json::Value>(
-          "importdescriptors",
-          &[serde_json::to_value(descriptors.clone())?],
-        ) {
-        Ok(_) => {
-          break;
-        }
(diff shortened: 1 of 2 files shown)


1stBitcoinSent:
> Do you know by chance what the Bitcoin Core error code or message is if a descriptor for a wallet is still loading?

I think it's:
`error code: -18
error message:
Requested wallet does not exist or is not loaded
`

src/wallet.rs
```diff
@@ -555,9 +555,34 @@ impl Wallet {
       });
     }
 
-    settings
-      .bitcoin_rpc_client(Some(name.clone()))?
-      .call::<serde_json::Value>("importdescriptors", &[serde_json::to_value(descriptors)?])?;
+    loop {
+      match settings
```
casey, line 559: This seems wrong, shouldn't we only ever call import descriptors once, as opposed to in a loop?

casey (approved):

Issue #3, opened 2021-12-12
Bug: Handle multiple atoms per coin


casey:
My code assumes that there's at most one atom per coin, in particular:

```rust
let mut atoms: BTreeMap<OutPoint, u64> = BTreeMap::new();
```

It should be:

```rust
let mut atoms: BTreeMap<OutPoint, Vec<u64>> = BTreeMap::new();
```

casey:
N/A

PR #4097, opened 2024-11-22
Add function to calculate rune unlock height

casey:
See `Rune::unlock_height`. This works but makes me nervous. I noticed that my height calculation was one too high whenever we were 0/4, 1/4, 2/4, or 3/4ths of the way through an interval of runes with a given length. So I just fudged it and subtracted one in those cases.

I've exhaustively tested lengths 1 through 7, and they're all correct. (Checked by cross checking with asserts that the result of `Rune::unlock_height` matches `Rune::minimum_at_height`.) However, I'm worried that this pattern is actually the result of some deep number theory harmonic integer sequence or something, and is somehow wrong for larger intervals.

### DIFF
--- a/crates/ordinals/src/rune.rs
+++ b/crates/ordinals/src/rune.rs
@@ -41,6 +41,8 @@ impl Rune {
 
   const UNLOCK_INTERVAL: u32 = SUBSIDY_HALVING_INTERVAL / 12;
 
+  const UNLOCKED: usize = 12;
+
   pub fn n(self) -> u128 {
     self.0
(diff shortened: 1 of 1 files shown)


t4t5:
I think I know what the issue is. PR opened here: https://github.com/casey/ord/pull/3 👍

casey:
@t4t5 Nice, that helps! I tweaked it a little bit to avoid floating point math. I think this version is now much less mysterious.

raphjaph (approved):

PR #4100, opened 2024-11-25
Add /r/rawtx/:transaction_id and /r/rawtxinfo/:transaction_id endpoints

alpergundogdu:
These commands bypass the index and act as a proxy to get raw transaction data from bitcoin cli

### DIFF
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -3370,6 +3370,292 @@ curl -s -H "Accept: application/json" \
 ```
 </details>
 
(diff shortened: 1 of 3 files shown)


casey:
We cannot expose JSON responses from Bitcoin Core in recursive endpoints because recursive endpoints have much stronger backwards compatibility guarantees. If Bitcoin Core changed the responses, inscriptions might break.

If you want access to the bytes of a transaction, that might be reasonable. Maybe better to open an issue first though, to discuss your use case.

PR #4101, opened 2024-11-25
Add sat values to address response (#3809)

cryptoquick:
Returns the sat value of each individual output in the address response. See issue: #3809

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -228,8 +228,15 @@ pub struct SatInscriptions {
 
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct AddressInfo {
-  pub outputs: Vec<OutPoint>,
+  pub outputs: Vec<OutpointInfo>,
   pub inscriptions: Vec<InscriptionId>,
   pub sat_balance: u64,
   pub runes_balances: Vec<(SpacedRune, Decimal, Option<char>)>,
+}
+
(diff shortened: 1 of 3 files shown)


cryptoquick:
Not sure how this test failure is related...
https://github.com/ordinals/ord/actions/runs/12012245784/job/33483021419?pr=4101#step:5:721

raphjaph:
I think that was just an intermittent failure.

raphjaph:
We already have the `sat_balance` so I'm unsure what this would gives us. I think by sats I meant rare sats (sats tracked by ord).

cryptoquick:
Sure, this just returns the sat value of each individual UTXO, since we already have that information, it's just all added together currently. Does that makes sense?

zk-bits:
Getting individual utxo values would be super helpful and mimics esplora more closely @raphjaph

raphjaph:
This is what the `/outputs/<ADDRESS>` endpoint is for. The `/address/<ADDRESS>` endpoint should only show aggregated information.

cryptoquick:
We're currently running ord 0.21.0 in production, and that endpoint doesn't seem to be working. Locally against regtest I see it working on 0.21.3. Is that API new, or do you think our server might be misconfigured?

cryptoquick:
Ah, I see in the release notes this new outputs API just came out last month in 0.21.3. Thanks, that should do just fine!

raphjaph:
Perfect!

PR #4105, opened 2024-11-26
Improve `ord wallet send` help message

casey:
The help text is definitely an improvement. I also think that changing the argument name from `outgoing` to `asset` is also an improvement. I'm not sure about changing `outgoing` to `asset` in the output. I think it's better, but it breaks any commands relying on it.


raphjaph (approved):
LGTM

Issue #4108, opened 2024-11-27
Recursion: Raw transaction endpoint for inscriptions


alpergundogdu:
We've talked about this endpoint a few months ago with @raphjaph on "Best in Slot <> ord" telegram channel and he mentioned there are other teams looking for a recursive raw transaction endpoint and invited us to open a PR for it.

In our use-case, we want to be able to track the inscriptions' old pkscripts and change the artwork based on some calculations over them. We can get the current txid via `/r/inscription/<INSCRIPTION_ID>`, we can get the genesis txid via `inscription_id` so we can apply sat movement rules to current sat location and track the inscription back to the genesis txid.

In order to do that, we need vout (pkscript and value), vin (txid and vout), and also blockhash to get the blocktime of the target transaction (we can use `/r/blockinfo/<QUERY>` to get the blocktime from blockhash).

I'll attach a PR to this issue in a bit, we updated the code per casey@'s initial suggestion on [PR 4100.](https://github.com/ordinals/ord/pull/4100)

Adding just the txhex and parsing client side could be error-prone if a new transaction version is introduced, the parser may stop working and the inscription may lose all history after a single new version tx.

For now, we chose a limited, stable set needed for base ordinals sat tracking, so we can assume these will always be in bitcoin as long as inscriptions work.

Verbose endpoint can keep some optional future fields in case another team needs more information about a tx and this information is also part of bitcoin and/or ordinals rules so it will always be available, new fields can always be added to the response but as casey@ mentioned we can never remove a field after it is introduced so we tried to limit the response fields as much as possible.

alpergundogdu:
Opened [PR 4109](https://github.com/ordinals/ord/pull/4109) for this.

raphjaph:
> Adding just the txhex and parsing client side could be error-prone if a new transaction version is introduced, the parser may stop working and the inscription may lose all history after a single new version tx.

Transactions have version fields for exactly that reason. Whatever you use client side to parse transactions from hex should handle new versions and be able to parse the subset of fields you're interested in. 

I understand the issue but now the onus flips on us to figure out what to return if a future transaction type doesn't have those "stable" fields you suggested. You could build in an upgrade mechanism such that if the sat is reinsribed with a new parsing library that will be used instead.

samedcildir:
We don't want to use an upgradable script since it defeats one of the main purposes of inscriptions which is to be immutable and to be completely in control of the art that you own.

[...]

PR #4109, opened 2024-11-27
Add /r/rawtxinfo/:transaction_id endpoint

alpergundogdu:
This endpoint bypasses the index and acts as a proxy to get raw transaction
data from bitcoin cli

Response from bitcoincore_rpc is converted into a defined structure in this
codebase to help with compatibility

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ rustls-acme = { version = "0.8.1", features = ["axum"] }
 serde = { version = "1.0.137", features = ["derive"] }
 serde-hex = "0.1.0"
 serde_json = { version = "1.0.132", features = ["preserve_order"] }
-serde_with = "3.7.0"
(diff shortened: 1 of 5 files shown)


alpergundogdu:
#4108 for more context

raphjaph:
Same issue as #4100. We should only return the raw hex encoded transaction similiar to how the `/r/metadata` only return the hex encoded CBOR.

samedcildir:
Hello @raphjaph, we've explained in the issue #4108 why a raw txhex might be problematic if a new transaction type is introduced. The inscriptions will not be able to decode those new transactions and the calculations will start to fail after that. A raw cbor is fine since cbor decoding will not change but txhex decoding may change with bitcoin upgrades.

For this reason we thought a minimal endpoint with fields that are also used in sat tracking and must be used in a utxo based system can be included in the response since they'll very likely stay the same.

raphjaph:
Merged #4147

PR #4117, opened 2024-11-30
Enable rune burning in wallet

### LINKED ISSUE #3640: Feature request: command to burn runes

raphjaph:
Closes #3640

Tested on mainnet:

https://mempool.space/tx/80516dbf06b3f9052ae0fb05b2e15db5fabbe5226e02019529571f00620c9740

https://ordinals.com/rune/RUNESREFRIGERATOR

### DIFF
--- a/src/subcommand/wallet/burn.rs
+++ b/src/subcommand/wallet/burn.rs
@@ -110,7 +110,7 @@ impl Burn {
           Amount::ZERO,
         )
       }
-      _ => unreachable!(),
+      _ => bail!("burning is only implemented for inscriptions and runes"),
     };
 
     let base_size = unsigned_transaction.base_size();
--- a/src/subcommand/wallet/burn.rs
+++ b/src/subcommand/wallet/burn.rs
@@ -97,7 +97,7 @@ impl Burn {
       Outgoing::Rune { decimal, rune } => {
         ensure!(
           self.cbor_metadata.is_none() && self.json_metadata.is_none(),
-          "metadata not supported for burning runes"
+          "metadata not supported when burning runes"
         );
 
         (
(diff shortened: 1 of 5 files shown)


src/subcommand/wallet/burn.rs
```diff
@@ -28,68 +28,88 @@ pub struct Burn {
...
+          self.cbor_metadata.is_none() && self.json_metadata.is_none(),
+          "metadata not supported for burning runes"
+        );
+
+        (
+          Self::create_unsigned_burn_runes_transaction(&wallet, rune, decimal, self.fee_rate)?,
+          Amount::ZERO,
+        )
+      }
+      _ => bail!("burning is only implemented for inscriptions and runes"),
```
casey, line 108: Let's make this exhaustive instead of a wildcard, and make the messages reflect that, i.e. "can't burn bitcoin", "can't burn satpoints", "can't burn sats", etc.

casey: We could easily support burning sats and satpoints, but we can do that later.

src/subcommand/wallet/burn.rs
```diff
@@ -141,4 +161,143 @@ impl Burn {
       .build_transaction()?,
     )
   }
+
+  fn create_unsigned_burn_runes_transaction(
```
casey, line 165: I definitely think we should deduplicate this with send, it's a long function, and they seem really close to each other.

casey (changes requested):

casey (approved):
LGTM

PR #4139, opened 2024-12-19
Show video controls on click

casey:
This PR modifies inscription video previews so controls (play, pause, etc) are only visible after clicking the preview.

Since thumbnail iframes are wrapped with an `<a>` tag, the `<a>` tag will intecept any clicks, so it entirely disables video controls on thumbnails, which I think is correct.

On `/inscription` the preview iframe isn't wrapped with an `<a>`, so you can enable controls by clicking on it.

I think this is nice, since otherwise video inscriptions load with visible controls, which is not very nice looking.

### DIFF
--- a/static/preview-video.js
+++ b/static/preview-video.js
@@ -1,4 +1,3 @@
-let video = document.getElementsByTagName('video')[0];
-video.addEventListener('click', () => {
-  video.controls = true;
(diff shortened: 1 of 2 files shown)


raphjaph (approved):
LGTM

PR #4142, opened 2024-12-21
Show overflow on <ol> in /blocks list

casey:
Show overflow on `<ol>` in /blocks list. The only thing that overflows is the block number on each `<li>`. I'm not sure why we hid it in the first place, but it looks pretty bad.

I also simplified the template slightly by calculating the height up front for each block, and indented the rust code to match the indentation level it would have in actual code.


raphjaph (approved):
LGTM

PR #4147, opened 2024-12-22
Add `/r/txhex/<txid>` for tx-hex recursion

samedcildir:
This endpoint bypasses the index and acts as a proxy to get raw transaction hex from bitcoin cli. It returns hex encoded raw transaction data as a JSON string.

PR also includes recursion docs update to reflect the new endpoint and tests for the new endpoint.

### DIFF
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -247,6 +247,28 @@ curl -s  \
 <details>
   <summary>
     <code>GET</code>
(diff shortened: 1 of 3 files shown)


samedcildir:
@casey @raphjaph is there any progress about this PR? I guess we've gone through all scenarios about this endpoint in the Issue #4108 and we can use the endpoint with an immutable transaction parsing library. Thank you for sharing your insights and time.

raphjaph (approved):
LGTM

PR #4152, opened 2024-12-26
Correct total sat supply figures & update current sat supply figures

ZedZeroth:
It might be worth @gmart7t2 double-checking these figures before they're committed. Original discussion is here: https://github.com/ordinals/ord/issues/3587 Thanks


casey:
LGTM! Double checked in #4153.

PR #4153, opened 2024-12-26
Add method returning rare sat supply

casey:
I wanted to double check these before merging #4152, so I I added a method to `Rarity` which returns the supply, along with a test to ensure that the figures are correct.


raphjaph (approved):
LGTM

PR #4156, opened 2024-12-27
Offers

raphjaph:
Tested in this tx: https://mempool.space/tx/78ccdf1ee2bf14f36e83653a99586329b2931a17f786ffa8e6e0b79650d2bc40


src/subcommand/wallet/offer/create.rs
```diff
@@ -0,0 +1,98 @@
...
+      .bitcoin_client()
+      .wallet_process_psbt(&result, Some(true), None, None)?;
+
+    ensure! {
+      !result.complete,
+      "PSBT unexpectedly complete after processing with wallet",
+    }
+
+    Ok(Some(Box::new(Output {
+      psbt: base64_encode(result.psbt),
```
lifofifoX, line 93: `result.psbt` is already base64 here.

src/subcommand/wallet/offer/create.rs
```diff
@@ -0,0 +1,98 @@
...
+          script_pubkey: wallet.get_change_address()?.into(),
+        },
+        TxOut {
+          value: self.amount + postage,
+          script_pubkey: seller_address.clone().into(),
+        },
+      ],
+    };
+
+    let tx = fund_raw_transaction(wallet.bitcoin_client(), self.fee_rate, &tx)?;
```
lifofifoX, line 73: Am I tripping or is this is not protecting inscribed/runic UTXOs?

raphjaph: your not tripping, good catch. Fixed

src/subcommand/wallet/offer/create.rs
casey: Nice, fixed!

casey:
@lifofifoX This is finally done! Have a look over it, especially at tests and edge cases.

lifofifoX:
@casey Looks great to me! Only potential issue I see is that accepting offer might error if the sat is reinscribed.

Other than that, it's nice that accepting offer isn't concerned with where the inscription is going. This way, it'd be compatible with offers created with one of the web wallets.

casey:
> @casey Looks great to me!

Dope.

> Only potential issue I see is that accepting offer might error if the sat is reinscribed.

We'll definitely relax that in a follow up PR. Either by allowing any inscriptions on the same sat to be transferred, as long as one of them was passed to `--inscription`, or requiring `--inscription` multiple times, and requiring you pass it for all reinscriptions.

Probably the former. These days, `ord` is really a power-user tool, and I think we can expect users to understand that when inscriptions are on the same sat, they will always be transferred together.

In any case, we will also eventually allow `--inscription` to be passed multiple times, so you can accept an offer for any number of inscriptions, and provide an `ord wallet offer view` command, for viewing a PSBT offer and seeing what it would do if accepted, which inscriptions it's moving, where they go, and what the balance change would be.

PR #4158, opened 2024-12-31
replace unwrap with safer ?

twosatsmaxi:


casey:
We use unwraps where it would be a bug if the value was `None` or `Err(_)`.

For example, in this case:

```
          let output = usize::try_from(output).unwrap();
```

`output` is a u32, and we expect that we're on a 32-bit or better system, so we expect `usize::try_from` to always succeed. In one of these cases, we actually want to panic on error, instead of returning a standard error, which would make it seem like it was a user error.

We actually have a trait, `IntoUsize`, which provides an infallible `into_usize` method, implemented for `u32`, to express this invariant.

A nice refactor would be to find those instances where we're doing `usize::try_from(x).unwrap()` and `x.try_into().unwrap()` where `x` is a `u32`, and replace them with `.into_usize()`.

PR #4176, opened 2025-01-12
Allow viewing contents of satscard

casey:


raphjaph (approved):
needs some docs but LGTM

PR #4180, opened 2025-01-14
update minimum required bitcoin core version to v25

cryptoni9n:
#4106 updated the required version of Bitcoin Core to 25 - this updates the documentation.

thanks to @so7ow for pointing this out


cryptoni9n:
> For this to work with testnet4 isn't v28 required?

hi @emilcondrea - yes, I think you're right, but this is PR is to just bring the documentation up to date with the MIN_VERSION specified in the src/wallet.rs.

raphjaph:
Thanks! I'll merge this for now but will update the version or write something in the docs about using testnet4 in a follow-up PR.

PR #4187, opened 2025-01-17
Remove double commit for redb bug

raphjaph:
@casey can we get rid of this now?


casey:
I'm not actually sure. @cberner do you know if we can remove this?

cberner:
No, I'm pretty sure that's https://github.com/cberner/redb/issues/829 which is still open

casey:
@cberner Thanks, we'll watch that issue!

PR #4189, opened 2025-01-17
fix Japanese translation (introduction, overview)

SatoshiRoppongi:
In order to increase the number of Japanese Ordinals fans, I have revised the Japanese translation of the handbook to be more natural(I'm Japanese).
Due to the large volume, only the introduction, overview has been modified at this time, but I plan to expand the translation in the future.


raphjaph (approved):
Thanks, looks good to me!

PR #4191, opened 2025-01-20
Add configurable max-savepoints and savepoint-interval

emilcondrea:
Testnet4 has deep reorgs, causing ord to end-up in unrecoverable state. It would be good to have configurable reorg environment so it can be more aggressive on testnet4 but also not risk too much on mainnet. @raphjaph, this is related to https://github.com/ordinals/ord/pull/4169


raphjaph (approved):
LGTM

PR #4204, opened 2025-01-29
Use middleware for proxied endpoint

arik-so:
This avoids the overhead of having to implement the NotFound proxy for each endpoint individually.


raphjaph (approved):
LGTM

PR #4211, opened 2025-02-07
Add blockhashes endpoint at /blockhashes/{start}/{interval}/{page}

### LINKED ISSUE #3879: Please can we have a batch blockhash endpoint to provide a source of global/permanent pseudorandom values for onchain games?
Games need random values that are permanently accessible to all players / the whole system so that "everyone" knows what the outcome was.

Examples could be (1) whether one player hits or misses another, (2) what the weather is/was at any given time, (3) whether any kind of random event happens or didn't happen.

Blockhashes are an ideal source of such (psuedo) random data for onchain games. Players cannot easily manipulate blockhashes so their value is unpredictable. Once they have been produced they remain static and provide the same value for all ordinals forever.

I have implemented such randomness in my "chaindrop lottery" that determines whether certain ordinals in the collection receive upgrades or not: https://palindromes.io/flowers/

I am pulling the historic fortnightly difficulty adjustment retarget blockhashes, but these sequential requests are currently the primary bottleneck slowing down the loading of the ordinals:

`async harvest(bgn,inc,end){
if(!end)end=parseInt(await this.rec.text('/blockheight'),10);
let hsh=[];
for(let i=bgn;i<=end;i+=inc){
const h=await this.rec.text('/blockhash/'+i);
hsh.push(h)}
return hsh}
}`

Ideally such an endpoint would include parameters for "start block", "increment" and "end block", and then return a JSON array of all relevant blockhashes. In my case, I start on the first DA blockhash after the halving (840672) then increment by 2016 until the current block height is reached. I think this is sustainable as it only requires an additional 26 blockhashes per year. If performance is a concern, then the parameters of the endpoint could be limited to daily/weekly/fortnightly increments.

Unless there is an alternative source of **continually generated** and **permanently stored** random values, then I think such an endpoint is essential for the creation of fully functional onchain games.

Thanks

cryptoni9n:
This endpoint is intended to fix #3879 by providing an array of 100 blockhashes starting at block `/{start}`.  If this endpoint is called with `/{start}/{interval}`, the array will contain 100 blockhashes starting at block `/{start}` and then skip `/{interval}` before the 2nd blockhash, and so on.  Finally, if the endpoint is called with `/{start}/{interval}/{page}`, the array will contain 100 blockhashes starting at block `/{start}` and then skip `/{interval}` before returning the `/{page}` of results for `/{start}/{interval}`.  If there are not 100 blockhashes for the provided parameters, a smaller dataset will be returned.  If there are no blockhashes for the provided parameters, an empty set will be produced.

For example, calling `curl -s -H "Accept: application/json"   http://localhost/blockhashes/0 | jq` will return blockhashes for blocks 0-99.  Calling `curl -s -H "Accept: application/json"   http://localhost/blockhashes/0/2 | jq` will return every other blockhash, starting with block 0 so that the set contains blockhashes for block 0, 2, 4, 6, etc-198.  Finally, the pagination works by calling `curl -s -H "Accept: application/json"   http://localhost/blockhashes/0/1/1 | jq` to get the second page of blockhashes starting with block 0 without an interval, so the array would contain blockhashes 100-199.  `curl -s -H "Accept: application/json"   http://localhost/blockhashes/0/100/0 | jq ` would give you block 0, 100, 200, 300, etc.

***Reviewer note**: this is definitely my most involved PR to date, and while I have tested this code and confirmed the results and that all tests pass, I'd be surprised if everything here is done 100% correctly.  I am also really new to tests, so I'm hoping the coverage included makes sense but am also open to adding more tests if needed.  Please provide feedback as to what needs improvement and I'll work to make it happen.  Additionally, I will add documentation once this has passed review, either in this or a new PR.  Thanks!

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -42,6 +42,30 @@ impl Block {
   }
 }
 
+#[derive(Debug, Deserialize)]
+pub struct BlockHashesParams {
+  pub start: u64,
+  pub interval: Option<u64>,
+  pub page: Option<u64>,
+}
+
+impl BlockHashesParams {
+  pub fn new(start: u64, interval: Option<u64>, page: Option<u64>) -> Self {
+    Self {
+      start,
+      interval,
+      page,
+    }
+  }
+}
+
+#[derive(Serialize)]
+pub struct BlockHashesJson {
+  pub block_hashes: Vec<String>,
+  pub prev_page: Option<u64>,
+  pub next_page: Option<u64>,
+}
+
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct BlockInfo {
   pub average_fee: u64,
--- a/src/index.rs
+++ b/src/index.rs
@@ -37,13 +37,14 @@ use {
 };
 
 pub use self::entry::RuneEntry;
+pub use rtx::Rtx;
 
 pub(crate) mod entry;
 pub mod event;
 mod fetcher;
 mod lot;
 mod reorg;
-mod rtx;
+pub mod rtx;
 mod updater;
 mod utxo_entry;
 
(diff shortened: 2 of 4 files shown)


cryptoni9n:
@raphjaph Hi Raph - I'm seeing some conflicts in src/subcommand/server.rs due to the changes in #4200.  Please provide guidance on how best to resolve.

cryptoni9n:
Closing this request for now; need to additionally implement this for recursion.

raphjaph:
Ok let me know when there's a new one

PR #4212, opened 2025-02-07
Implement galleries

### LINKED ISSUE #4140: Galleries and legacy collections
Parent child has slowly been getting more adoption, but it hasn't been universal, and many important collections exist which cannot adopt it retroactively.

If a collection does not use parent-child, there is no on-chain reference for which inscriptions are part of the collection and which aren't, which is extremely important for collectors, forcing them to rely on centralized sources like the collection project's socials or exchanges.

I'd like to create some way for these legacy collections, as well as new collections which can't use parent-child (for example because they mint on demand) to have an on-chain indication of which inscriptions are in the collection.

I'm a bit hesitant to introduce something which overlaps in functionality with parent-child, but since adoption of parent-child has been relatively slow, I think it makes sense to start at least talking about it and seeing what it would look like.

My current idea is to add something called a "gallery", which is a list of inscription IDs that are included in an inscription, either under a new tag, or in the existing CBOR metadata. The inscriptions in a gallery would be displayed below the inscription on `/inscription`, separately from any children.

For example, if the skulls under inscription 0 used a gallery instead of parent/child, they would be displayed as:

Screenshot 2024-12-19 at 3 22 23 PM

Galleries have the following pros and cons:

- Unlike parent-child, it is not possible to add additional inscriptions to a gallery, it is one and done.
- Like parent-child, it is not possible to remove an inscription from a gallery.
- Unlike parent-child, which requires the inscriber to have custody of the parent and be the creator of the child, *anyone* can create a gallery with any inscriptions.

Because of the last point, that anyone can create a gallery with any inscriptions, we cannot display backlinks from a gallery item to the gallery inscription itself. You could navigate to the gallery, then click an item and go to that item, but you could not go to an item and see a backlink up to the galleries that contain it. If we allowed this, anyone could spam inscriptions that they don't hold into galleries, and imply a relationship which does not exist.

There are a few ways that galleries could be encoded in inscriptions:

1. Add a new odd tag, and include gallery member inscription IDs under this tag
2. Add a new odd tag, for structured CBOR metadata, and include gallery member inscription IDs under this tag
3. Allow including gallery member inscriptions IDs in the existing CBOR metadata in tag 5

The difference between 1 and 2 is that in 1, we would have a new, dedicated tag for gallery members, and in 2, we would have a more general purpose tag for structured metadata.

For 3, we would reuse the existing metadata tag, and would need to have some way of avoiding conflicts. Let's say that we went with 3 and used the top-level key `gallery`. This could conflict with existing metadata which contains a "gallery" key. A way of avoiding that, which ties into some thoughts about more structured metadata that I've been having, would be to introduce a way of opting into a more structured metadata schema, perhaps with a new odd tag which gave the metadata schema version. If this additional tag was set, metadata would be interpreted according to a schema that we define, with all keys being reserved for our use, except for a top level `gallery` key for gallery item inscription IDs, a top level `unstructured` key, for the continued inclusion of unstructured metadata. New keys could be defined as time goes on, for example a more structured, standardized, and richer "attributes" tag.

I think 3 is appealing from the point of view of the ord codebase, since we already have functionality to allow users to provide JSON and CBOR metadata, and this would allow them to provide the gallery member entries via the same features.

For example, to inscribe with the new metadata schema, and provide the metadata, including gallery member entries in `metadata.json`:

```console
ord wallet inscribe --metadata-schema 1 --json-metadata metadata.json …
```

This would also allow inscriptions to use the recursive metadata endpoint to fetch metadata, and avoid having to define new endpoints to fetch gallery members.

Feedback is welcome!

casey:
Fixes #4140.


raphjaph:
Think about naming gallery something else. 

Also tests are failing.

stet:
> Think about naming gallery something else.
> 
> Also tests are failing.

They are actually curated onchain **exhibits**. Many exhibits can be in a "gallery" which would be an external presentation layer.

PR #4214, opened 2025-02-10
Don't unnecessarily bump pointer by postage for SameSat mode

SmarakNayak:
Noticed that the pointer gets bumped by the postage amount for every inscription even when in `SameSat` mode. Added a small fix to only bump the pointer when in other modes.

This doesn't change which sat inscriptions are attached to because they all overflow back to a pointer of 0 anyway. You do save some space on unnecessary pointers though.


raphjaph (approved):
Good find! 

This is actually a bug (it can't be triggered because of the way our wallet constructs the transactions). Thanks for the fix, LGTM

PR #4215, opened 2025-02-10
Upgrade redb to 2.4.0 to improve reorg performance

emilcondrea:
Reorg just happened today and I have a version of ord 22.1 running redb 2.3.0 that is still working to recover(and probably will take 1-2 hours) and the one with redb 2.4.0 recovered in 1 second.

https://github.com/cberner/redb/releases/tag/v2.4.0

https://github.com/ordinals/ord/issues/4207 @raphjaph


raphjaph (approved):
LGTM

PR #4235, opened 2025-02-18
fix-Update main.rs

romashka-btc:
fix 404 links


raphjaph:
Thanks, Fixed in #4242

PR #4238, opened 2025-02-22
Allow empty query parameters on /satscard route

casey:
Also threw in some fixes for clippy lints from rust 1.85.


raphjaph (approved):
Lgtm

PR #4243, opened 2025-02-26
Extend properties CDDL with attributes

### LINKED ISSUE #4222: Attributes
We recently merged #4212, which adds the ability to create "galleries". Any inscription can contain a gallery, which is a list of inscription IDs. These inscriptions appear in a dedicate section called "gallery" on `/inscription`, similar to how children appear.

There are many use cases for galleries, one of which is allowing off-chain collections to inscribe a gallery with the members of the collection, so that there exists an immutable on-chain record of the members of the collection, lessening ecosystem-wide dependence on centralized services.

However, off-chain collections, in addition to simply being a list of inscriptions, also include per-inscription metadata, such as title, background color, accessories, etcetera. Ideally there is some path for these collections to also migrate their metadata on chain.

Gallery items may currently only contain an inscription id, but they are encoded as CBOR maps, and so may be extended with additional attributes.

The goal of this issue is to define how attributes may be included with gallery items, to allow all off-chain collection data to be migrated to a single, immutable gallery inscription.

We could make this metadata free form, and simply make it the dual of inscription metadata, which is also free-form. But I think it may be desirable to define a strict schema, so that attributes are more easily searchable, and are in a fixed format. If we define such a schema, we would also allow it to be used for non-gallery inscriptions.

casey:
Fixes #4222. As a first stab, just extend the CDDL.

To do:
- write tests for all different value types
- make sure trait order is preserved
- how should gallery item metadata be displayed?
- hidden metadata
- look at existing collection attributes
- add a value for "none", since this seems to be common

### DIFF
--- a/docs/src/inscriptions/properties.md
+++ b/docs/src/inscriptions/properties.md
@@ -20,13 +20,40 @@ of the properties value is as follows:
 ```cddl
 Properties = {
   ? 0: [*GalleryItem],
+  ? 1: Attributes,
   * any => any,
 }
 
 GalleryItem = {
-  ? 0: bstr .size (32..36),
+  ? 0: bytes .size (32..36), ; inscription ID
+  ? 1: Attributes,
   * any => any,
 }
+
+Attributes = {
+  ? 0: text,   ; title
+  ? 1: text,   ; description
+  ? 2: text,   ; alt text
+  ? 3: text,   ; author
+  ? 4: Traits, ; traits
+}
+
+Traits = {
+  * text => Value
+}
+
+Value =
+  Proportion /
+  bool /
+  bytes /
+  float /
+  integer /
+  tdate /
+  text /
+  time /
+  uri
+
+Proportion = #6.6582895(float)
 ```
 
 The above CDDL schema is provided as a convenience. As always, the ordinals
--- a/docs/src/inscriptions/properties.md
+++ b/docs/src/inscriptions/properties.md
@@ -54,6 +54,8 @@ Value =
   uri ; <a> tag
 
 Proportion = #6.6582895(float) ; value may be between 0 and 1 inclusive
+InscriptionId = #6.6582896(bytes .size (32..36)) ; links to `/inscription/<id>`
+RuneId = #6.6582897([integer, integer]) ; links to `/rune/<id>`
 ```
 
 The above CDDL schema is provided as a convenience. As always, the ordinals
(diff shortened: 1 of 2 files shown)


lifofifoX:
Few more things came to mind.

- `alt text` is going to be same as `title` in majority, if not all, of the cases. Why not start without it?
- Most links will likely require some label. How about including labels in the `? 4: [+ uri], ; links` attribute? I understand people can just use a trait instead. But seems odd to have both. Maybe `? 4: [+ uri], ; links` can be removed altogether in favor of traits.
- Traits value type list seems extensive. Why not start small and expand as we go along, in order to avoid having types almost no one ever uses?

casey:
> * `alt text` is going to be same as `title` in majority, if not all, of the cases. Why not start without it?

`alt text` is a description, not a title. Although I think maybe it's mostly superfluous with title and description, even though alt text is supposed to be a literal description of the content, so I removed it.

> * Most links will likely require some label. How about including labels in the `? 4: [+ uri], ; links` attribute?

I'm wondering about this. I think most links are explanatory. For example, if it's a link to `discord.gg`, we know it's Discord, and can even give it a Discord logo.

> I understand people can just use a trait instead. But seems odd to have both. Maybe `? 4: [+ uri], ; links` can be removed altogether in favor of traits.

I think that they're presented differently. Traits should be grouped under together, but project links definitely don't belong with traits.

This seems better:

```
links:
- https://discord.gg
traits:
- hat: red
- background: black
```

Than this:

```
traits:
- discord: https://discord.gg
- hat: red
- background: black
```

> * Traits value type list seems extensive. Why not start small and expand as we go along, in order to avoid having types almost no one ever uses?

The all seem reasonably useful and reasonably easy to implement. But if you think one in particular isn't useful, then feel free to suggest that we remove it.

lifofifoX:
> But if you think one in particular isn't useful, then feel free to suggest that we remove it.

`Proportion` and `bytes` stood out as something that might not get much use.

`tdate` and `time` seem to achieve the same. Is there a scenario where you'd want to use one over the other?

casey:
> `Proportion` and `bytes` stood out as something that might not get much use.

Proportion is actually one of the very few data types which is supported by OpenSea, so I think we should have it for coverage of what people use for ERC-721s. I think it also gets some use for "power level" type stuff.

Bytes I agree is maybe niche, but raw bytes are elemental, so I definitely want people to be able to use them.

> `tdate` and `time` seem to achieve the same. Is there a scenario where you'd want to use one over the other?

`tdate` has a time zone, so it's a localized time. `time` is a time-zone-free integer timestamp. I definitely see how they overlap. Although I think they're both useful.

My thinking is that I definitely usually think changes should be minimal, but adding different value types just seems easy enough.

casey:
This has been implemented.

PR #4244, opened 2025-02-26
docs: redacted twitter nickname

Olexandr88:


raphjaph:
Thanks, but I think Erin wanted it to be like this.

Olexandr88:
Ok

PR #4248, opened 2025-02-28
Possible typo here: 3.md

oxbau:
Hi,
I noticed a -possible- typo in this text.

"Popularity is the font of value"

    "Font" should likely be "fount," as "fount of value" is the correct phrase.
    Corrected: "Popularity is the fount of value"

Thanks.


casey:
[There appears to be some controversy about this!](https://www.grammarphobia.com/blog/2017/04/font-fount.html)

It looks like people do use font, but it's not the traditional form.

PR #4252, opened 2025-03-03
Migrate to BDK wallet

raphjaph:
The [`ord/bdk`](https://github.com/ordinals/ord/tree/bdk) branch is where all the work is collected.

### TODO:

- [ ] make sure we can support multiple accounts without reentering key
- [ ] keep open the possibility of a wallet password
- [ ] make sure we can support multisigs in the future
- [ ] allow configuring lookahead (gap limit), setting?
- [ ] should we publish our persister to get review?
- [ ] should we pay a BDK dev to review/audit our code?
- [ ] think very hard about the database schema before final merge
- [ ] can we use the new descriptor format for combined external and internal (`<0;1>`)?
- [ ] add test for timestamp restore
- [ ] make sure that if we drop the wallet without persisting, we either panic, or we try to persist

### Notes

- what should the format for importing/restoring a wallet from a descriptor be? The same as before?
- how should we handle the `rawtr(...)` commit output backups/recovery keys?
- for the commit output  could we derive the untweaked key pair from the descriptor?

### Tests

- [ ] test that you get error when ord has wrong network

### TODO after merge into `master`

- [ ] rename `TestServer` to `Ord`
- [ ] rename `run_and_deserialize_outptut` to `run_and_deserialize_stdout`
- [ ] use strongly-typed Snafu errors for all wallet errors


casey:

PR #4254, opened 2025-03-04
Always store script pubkey in output table

raphjaph:


src/index/utxo_entry.rs
```diff
@@ -21,13 +21,12 @@ enum Sats<'a> {
...
 /// stored as a varint.
 ///
-/// If `--index-addresses`, the script pubkey stored as a varint followed by
-/// that many bytes of data.
-///
 /// If `--index-inscriptions`, the list of inscriptions stored as
 /// `(sequence_number, offset)`, with the sequence number stored as a u32 and
 /// the offset as a varint.
 ///
+/// The script pubkey stored as a varint followed by that many bytes of data.
```
casey, line 28: Why did this comment move? It looks like the order of how things are stored didn't change.

src/index/utxo_entry.rs
```diff
@@ -61,14 +59,12 @@ impl UtxoEntry {
...
-      offset += varint_len;
+    let (script_pubkey_len, varint_len) = varint::decode(&self.bytes[offset..]).unwrap();
+    offset += varint_len;
 
-      let script_pubkey_len: usize = script_pubkey_len.try_into().unwrap();
-      script_pubkey = Some(&self.bytes[offset..offset + script_pubkey_len]);
-      offset += script_pubkey_len;
-    }
+    let script_pubkey_len = usize::try_from(script_pubkey_len).unwrap();
+    let script_pubkey = Some(&self.bytes[offset..offset + script_pubkey_len]);
```
casey, line 66: The forward variable declaration of script_pubkey was no longer necessary.

casey (changes requested):
This looks reasonable to me, but I think it needs a test, and we need to profile index time and index size. I think this is probably fine but we need to understand the impact of doing it.

casey (approved):
LGTM

PR #4255, opened 2025-03-04
Runes 2.0 proposal

VanjaRo:
# Runes 2.0 proposal

Currently Runes have a specific Mint mechanism where anyone can mint a certain amount of runes for some specified period.

As Runes has gained huge popularity as the native token standard for BTC, there is a suggestion to improve the existing standard by allowing the specification of Minter Bitcoin address in Etching inscription.

```rs
struct Etching {
  divisibility: Option<u8>,
  premine: Option<u128>,
  rune: Option<Rune>,
  spacers: Option<u32>,
  symbol: Option<char>,
  terms: Option<Terms>,
  minter: Option<u32>,
  // ...other minting parameters, discussable...
}
```

To preserve the original compact structure of Runestone we can use `Runestone.Edict` for additional mintings. `Etching.minter` can make `Edict { id, Amount: minAmount, Output: outputIndex }` and place a transaction output at outputIndex with the output script set to the target address, to which minting should happen. To only allow this minting for `Etching.minter` we must add a condition to accept a transaction: at least one of the inputs in the transaction must use the previous output that belongs to `Etching.minter`. 

## Potential solution to allow Minter address change:

We can set `Runestone.Mint` and make `Edict { id, Amount: 0, Output: outputIndex, }` with 0 amount where `outputIndex` will correspond to output with a script set to the corresponding script of the new Minter address.

## Global adjustments

There can be other suggestions on how to efficiently add this targeted minting mechanism by Minter and how to perform minting transaction validation.

Small adjustments can be easily reflected at the official Ord indexer, libraries and wallets that support Runes.

## Main use case  

The main use case to support such minting is to enable Runes stablecoins where the total supply of coins cannot be known in advance.

### DIFF
--- a/crates/ordinals/src/etching.rs
+++ b/crates/ordinals/src/etching.rs
@@ -9,6 +9,7 @@ pub struct Etching {
   pub symbol: Option<char>,
   pub terms: Option<Terms>,
   pub turbo: bool,
+  pub minter: Option<u32>,
 }
 
 impl Etching {
--- a/crates/ordinals/src/runestone.rs
+++ b/crates/ordinals/src/runestone.rs
@@ -84,6 +84,11 @@ impl Runestone {
         ),
       }),
       turbo: Flag::Turbo.take(&mut flags),
+      minter: Tag::Minter.take(&mut fields, |[minter]| {
+        let minter = u32::try_from(minter).ok()?;
+        // Check if the minter is a valid output index
+        (u64::from(minter) < u64::try_from(transaction.output.len()).unwrap()).then_some(minter)
+      }),
     });
 
     let mint = Tag::Mint.take(&mut fields, |[block, tx]| {
@@ -148,6 +153,7 @@ impl Runestone {
       Tag::Spacers.encode_option(etching.spacers, &mut payload);
       Tag::Symbol.encode_option(etching.symbol, &mut payload);
       Tag::Premine.encode_option(etching.premine, &mut payload);
+      Tag::Minter.encode_option(etching.minter, &mut payload);
 
       if let Some(terms) = etching.terms {
         Tag::Amount.encode_option(terms.amount, &mut payload);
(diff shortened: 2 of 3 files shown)


Georgyskr:
Hey Ords Team!
Georgy speaking here

MIDL, as a foundation, deeply values the Runes community and would love to have our project’s token as a Rune. Additionally, we currently support issuing stablecoins as Runes.

The ability to mint Runes with authorization is crucial for many use cases that have been successfully established on other networks. Therefore, we’d like to start a discussion with the core Ords devs on possible ways to implement this—without requiring the minting of additional Runes to be tied to an initial batch.

This is just an initial proposal from our team, as a full implementation would require careful and thorough updates to Ord’s indexation. Before making significant changes at the repo level, we wanted to present the concept and align on the development direction.

Looking forward to your thoughts!

casey:
Thanks for the PR! I believe this is a dupe of this closed issue: #3677

Please respond in that issue, but my TL;DR on my view is: I think that either issuing a bunch up front and holding them in reserve, or making a new rune to represent new issuance, is probably a fine solution and less complicated. Especially if there is an easy, standard way that makes runes held in reserve shown as "in reserve" and not added to the circulation total.

afa7789:
@casey sorry, I don't get it, will there be any effort towards this in the future ? I did not understand the current situation for this, nothing was done in the closed issue #3677 and at this too. ( just PR reverted ).

but looks like most of the people who are asking for the future still preffers to have a stricted minter flag, is there a way to get this feature done ? I have enough time to work on something like this ( Would of course need to study the project structure to contribute to it ).

casey:
I don't think that this feature is a good idea. It introduces substantial complexity for unclear gain. See the discussion in #3677.

afa7789:
@casey  Yeah I have finished reading it.

looks like the proposed "workaround" still doesn't cover some certain use cases, as : 
- Pre-mining the total supply is legally problematic in many jurisdictions (regulators may see it as securities issuance).  
- Without mint restriction, anyone could mint and break trust, value, and utility of such tokens.

the status quo of the project is to not allow this type of solution//feature in it ? it's something ordinals team does not want ? would it mean the case for one more protocol ? so  the endgame is that if someone is looking for something like what was requested in the issues to use other technology?

thanks for the quick reply any way.

casey:
@afa7789 Have you considered the downsides of adding this feature? What would be the downsides of adding a "minter address" which was allowed to mint runes?

afa7789:
downside is maintaining more code, and complexity.

Issue #4256, opened 2025-03-04
Some resources and notes for locking scripts


raphjaph:
- https://github.com/SachinMeier/bitcoinex/blob/master/lib/taproot.ex
- https://github.com/Larkhell/redeem-script-generator/blob/main/src/index.js
- https://github.com/paulmillr/scure-btc-signer/blob/main/src/utils.ts#L100
- https://gist.github.com/lifofifoX/2fb7fe22f1cb17fd71bf8fc43fb6fe58
- https://github.com/paulmillr/scure-btc-signer/blob/main/src/utils.ts#L100

lifofifoX:
> secondly it would make more sense to lock bitcoin native assets using bitcoin-time aka block-height

You can use block height with `OP_CLTV` and absolute heights are conceptually simpler than relative ones. You can check out my implementation on https://tx.dev/utxo/timelock

samedcildir:
I didn't know that we could use block height with CLTV. However if we use absolute block height, the lock wallet address will change for every locked utxo and it would be much harder to implement it with current indexers and wallets. If we use relative block height with a few pre-selected standard heights, it would mean the indexer only needs to track a few extra wallet addresses which would be very easy to implement.

Also with absolute block height limit, since users cannot control which block height the transaction will be mined, they may not be able to lock it the exact amount of blocks needed and probably one of the most requested feature which is staking will require a certain amount of lock to give something in return.

lifofifoX:
@samedcildir Good point re: address discovery!

And `OP_CSV` definitely makes more sense from staking point of view. I'm thinking about this more from primary mint POV, where you'd want all assets to unlock at the same time.

PR #4259, opened 2025-03-05
Include blockhash in /r/utxo response

lifofifoX:
This enables recursive retrieval of a UTXO blockhash, allowing to determine how long an inscription has been held or when it was burned.

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -164,6 +164,7 @@ pub struct UtxoRecursive {
   pub runes: Option<BTreeMap<SpacedRune, Pile>>,
   pub sat_ranges: Option<Vec<(u64, u64)>>,
   pub value: u64,
+  pub blockhash: Option<BlockHash>,
 }
 
 #[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
--- a/src/index.rs
+++ b/src/index.rs
@@ -2393,11 +2393,20 @@ impl Index {
       return Ok(None);
     };
 
+    let Some(tx_info) = self
+      .client
+      .get_raw_transaction_info(&outpoint.txid, None)
+      .into_option()?
+    else {
+      panic!("can't get transaction info: {}", &outpoint.txid);
+    };
+
     Ok(Some(api::UtxoRecursive {
       inscriptions: self.get_inscriptions_for_output(outpoint)?,
       runes: self.get_rune_balances_for_output(outpoint)?,
       sat_ranges: self.list(outpoint)?,
       value: utxo_entry.value().parse(self).total_value(),
+      blockhash: tx_info.blockhash,
     }))
   }
 
--- a/src/subcommand/server.rs
+++ b/src/subcommand/server.rs
@@ -6072,7 +6072,7 @@ next
       ..default()
     });
 
-    server.mine_blocks(1);
+    let block = &server.mine_blocks(1)[0];
 
     let inscription_id = InscriptionId { txid, index: 0 };
     let second_inscription_id = InscriptionId { txid, index: 1 };
@@ -6102,6 +6102,7 @@ next
           (50 * COIN_VALUE, 2 * 50 * COIN_VALUE)
         ]),
         value: 150 * COIN_VALUE,
+        blockhash: Some(block.header.block_hash()),
       }
     );
   }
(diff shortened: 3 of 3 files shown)


samedcildir:
This would be a great addition since with this we can also determine the remaining locktime of the assets locked with the methods in #4256

src/index.rs
```diff
@@ -2393,11 +2393,20 @@ impl Index {
       return Ok(None);
     };
 
+    let Some(tx_info) = self
+      .client
+      .get_raw_transaction_info(&outpoint.txid, None)
+      .into_option()?
+    else {
+      panic!("can't get transaction info: {}", &outpoint.txid);
```
f7931a-hc, line 2401:
```suggestion
      return Ok(None);
```

src/api.rs
```diff
@@ -164,6 +164,7 @@ pub struct UtxoRecursive {
   pub runes: Option<BTreeMap<SpacedRune, Pile>>,
   pub sat_ranges: Option<Vec<(u64, u64)>>,
   pub value: u64,
+  pub blockhash: Option<BlockHash>,
```
raphjaph, line 167: This should not be an Option since we would only ever return confirmed UTXOs (i.e. no utxos in the mempool).

src/index.rs
```diff
@@ -2393,11 +2393,20 @@ impl Index {
...
+    };
+
```
raphjaph, line 2403: We shouldn't panic here, so probably just return Ok(None) like above.

raphjaph (changes requested):
I like it, just some small things

src/api.rs
lifofifoX: Fixed in c3d9d4d7ed2486c489afe8b81b325c5f1e0e9c0b

src/index.rs
lifofifoX: Fixed in c3d9d4d7ed2486c489afe8b81b325c5f1e0e9c0b

raphjaph:
Could you add a test for this?

docs/src/inscriptions/recursion.md
```diff
@@ -1010,7 +1010,8 @@ curl -s \
   "inscriptions": null,
   "runes": null,
   "sat_ranges": null,
-  "value": 5000000000
+  "value": 5000000000,
+  "blockhash": "00000000000000000000f238f617bd97d3e877108328102d9adcb14a2a92ef00"
```
casey, line 1014: These should be alphabetically sorted.

docs/src/inscriptions/recursion.md
```diff
@@ -1039,7 +1040,8 @@ curl -s \
       1905800627509443
     ]
   ],
-  "value": 330
+  "value": 330,
+  "blockhash": "00000000000000000000f238f617bd97d3e877108328102d9adcb14a2a92ef00"
```
casey, line 1044: sort

src/api.rs
```diff
@@ -164,6 +164,7 @@ pub struct UtxoRecursive {
   pub runes: Option<BTreeMap<SpacedRune, Pile>>,
   pub sat_ranges: Option<Vec<(u64, u64)>>,
   pub value: u64,
+  pub blockhash: BlockHash,
```
casey, line 167: sort

src/index.rs
```diff
@@ -2477,11 +2477,20 @@ impl Index {
...
+    else {
+      return Ok(None);
+    };
+
     Ok(Some(api::UtxoRecursive {
       inscriptions: self.get_inscriptions_for_output(outpoint)?,
       runes: self.get_rune_balances_for_output(outpoint)?,
       sat_ranges: self.list(outpoint)?,
       value: utxo_entry.value().parse(self).total_value(),
+      blockhash: tx_info.blockhash.unwrap(),
```
casey, line 2493: sort

casey, line 2493: Under what circumstances will this unwrap panic?

casey (changes requested):
See comments.

src/index.rs
lifofifoX: It should never panic because this only gets called if there's an entry present in `OUTPOINT_TO_UTXO_ENTRY`.

lifofifoX:
@raphjaph I updated the existing test to add coverage for this. Do you think this needs its own test?

PR #4272, opened 2025-03-10
Better preview titles

jtraub91:
Add better titles to preview pages (per https://github.com/ordinals/ord/issues/4264)

In this solution, `inscription_number` is fetched from the index entry where necessary, and passed to the preview structs and templates, to be used as part of the HTML title tags, respectively. Note also `inscription_number` is added to the `r::content_response` function signature, to be able to be passed to `PreviewIframeHtml`, and explicitly set to 0 in tests.


raphjaph (approved):
LGTM

I just got rid of some of the `unwraps` and added a small test.

PR #4288, opened 2025-03-22
Recognize inscriptions with content type `audio/ogg` as audio

### LINKED ISSUE #4285: Not all `audio/ogg` inscriptions have content type `audio/ogg;codecs=opus`
https://ordinals.com/inscription/cae866439378149018855e6538e406beb53129012e9e26dba8cfb68d821d866ai0

casey:
When making inscriptions, we prefer to use the most specific content type possible, including media type parameters. However, when deciding to use which preview to use, we should accept media types without parameters.

Fixes #4285.


raphjaph (approved):
LGTM

PR #4293, opened 2025-03-23
Index Runes explicitly spent to op_return

### LINKED ISSUE #4208: /r/utxo does not know about rune burns
I burned some Runes in [149f8cba424976a0e4196694b849184b96f27a5ccfb4ae171dbbabdcf26bb3e0](https://mempool.space/tx/149f8cba424976a0e4196694b849184b96f27a5ccfb4ae171dbbabdcf26bb3e0)

Decoded Runestone:

```json
{
  "inscriptions": [],
  "runestone": {
    "Runestone": {
      "edicts": [
        {
          "id": "864670:295",
          "amount": 500000000,
          "output": 0
        }
      ],
      "etching": null,
      "mint": null,
      "pointer": null
    }
  }
}
```

However, [/r/utxo](https://ordinals.com/r/utxo/149f8cba424976a0e4196694b849184b96f27a5ccfb4ae171dbbabdcf26bb3e0:0) does not include anything re: this burn.

```json
{"inscriptions":[],"runes":{},"sat_ranges":[],"value":0}
```

What's the best way to retrieve Rune burns through recursive endpoints?

lifofifoX:
Attempts to fix #4208.

It currently has failing tests that I don't know how best to handle.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -6729,13 +6729,22 @@ mod tests {
           ..Default::default()
         },
       )],
-      [(
-        OutPoint {
-          txid: txid3,
-          vout: 1,
-        },
-        vec![(id, 889)],
-      )],
+      [
+        (
+          OutPoint {
+            txid: txid3,
+            vout: 0,
+          },
+          vec![(id, 111)],
+        ),
+        (
+          OutPoint {
+            txid: txid3,
+            vout: 1,
+          },
+          vec![(id, 889)],
+        ),
+      ],
     );
 
     event_receiver.blocking_recv().unwrap();
--- a/src/index/updater/rune_updater.rs
+++ b/src/index/updater/rune_updater.rs
@@ -180,7 +180,6 @@ impl RuneUpdater<'_, '_, '_> {
         for (id, balance) in &balances {
           *burned.entry(*id).or_default() += *balance;
         }
-        continue;
       }
 
       buffer.clear();
@@ -198,14 +197,16 @@ impl RuneUpdater<'_, '_, '_> {
       for (id, balance) in balances {
         Index::encode_rune_balance(id, balance.n(), &mut buffer);
 
-        if let Some(sender) = self.event_sender {
-          sender.blocking_send(Event::RuneTransferred {
-            outpoint,
-            block_height: self.height,
-            txid,
-            rune_id: id,
-            amount: balance.0,
-          })?;
+        if !tx.output[vout].script_pubkey.is_op_return() {
+          if let Some(sender) = self.event_sender {
+            sender.blocking_send(Event::RuneTransferred {
+              outpoint,
+              block_height: self.height,
+              txid,
+              rune_id: id,
+              amount: balance.0,
+            })?;
+          }
         }
       }
 
--- a/src/runes.rs
+++ b/src/runes.rs
@@ -643,7 +643,10 @@ mod tests {
           ..default()
         },
       )],
-      [(OutPoint { txid, vout: 0 }, vec![(id, 100)])],
+      [
+        (OutPoint { txid, vout: 0 }, vec![(id, 100)]),
+        (OutPoint { txid, vout: 1 }, vec![(id, 100)]),
+      ],
     );
   }
 
@@ -1266,7 +1269,7 @@ mod tests {
       )],
     );
 
-    context.core.broadcast_tx(TransactionTemplate {
+    let txid1 = context.core.broadcast_tx(TransactionTemplate {
       inputs: &[(id.block.try_into().unwrap(), 1, 0, Witness::new())],
       outputs: 2,
       op_return: Some(
@@ -1297,7 +1300,7 @@ mod tests {
           ..default()
         },
       )],
-      [],
+      [(OutPoint { txid: txid1, vout: 2 }, vec![(id, u128::MAX)])],
     );
   }
 
@@ -2617,7 +2620,7 @@ mod tests {
           ..default()
         },
       )],
-      [],
+      [(OutPoint { txid, vout: 1 }, vec![(id, u128::MAX)])],
     );
   }
 


raphjaph:
In order to fix the tests you will have to filter out all outputs in the `balances` struct that are `OP_RETURN`. This function has access to `self.index` so there should be a way to easily get that information.

```rust
  #[track_caller]
  pub(crate) fn assert_runes(
    &self,
    mut runes: impl AsMut<[(RuneId, RuneEntry)]>,
    mut balances: impl AsMut<[(OutPoint, Vec<(RuneId, u128)>)]>,
  ) {
    let runes = runes.as_mut();
    runes.sort_by_key(|(id, _)| *id);

    let balances = balances.as_mut();
    balances.sort_by_key(|(outpoint, _)| *outpoint);

    for (_, balances) in balances.iter_mut() {
      balances.sort_by_key(|(id, _)| *id);
    }

    pretty_assert_eq!(runes, self.index.runes().unwrap());

    pretty_assert_eq!(balances, self.index.get_rune_balances().unwrap());

    let mut outstanding: HashMap<RuneId, u128> = HashMap::new();

    for (_, balances) in balances {
      for (id, balance) in balances {
        *outstanding.entry(*id).or_default() += *balance;
      }
    }

    for (id, entry) in runes {
      pretty_assert_eq!(
        outstanding.get(id).copied().unwrap_or_default(),
        entry.supply() - entry.burned
      );
    }
  }
```

src/index/updater/rune_updater.rs
```diff
@@ -198,14 +197,16 @@ impl RuneUpdater<'_, '_, '_> {
...
+        if !tx.output[vout].script_pubkey.is_op_return() {
+          if let Some(sender) = self.event_sender {
+            sender.blocking_send(Event::RuneTransferred {
+              outpoint,
+              block_height: self.height,
+              txid,
+              rune_id: id,
+              amount: balance.0,
+            })?;
+          }
```
raphjaph, line 209: I don't really consume this event_sender so I'm unsure if people would want sending to an OP_RETURN to be an event.

src/index/testing.rs
```diff
@@ -128,7 +128,20 @@ impl Context {
...
+        .index
+        .get_output_script(outpoint)
+        .ok()
+        .is_some_and(|script| script.is_op_return())
+      {
+        for (_, balance) in balances.iter() {
+          burned_balances += *balance;
+        }
+      }
+
```
raphjaph, line 144: To be extra sure we should not only calculate the burned amount here but also ask `self.index` for the burned amount.

raphjaph:
Apologies for letting this sit so long. You were on the right track but were just missing a call to make tests simpler. I've made the changes, please have a look and let me know if this looks good @lifofifoX

src/index/updater/rune_updater.rs
lifofifoX: Yeah, same. I thought it'd be a good idea to keep existing behavior there. But could argue both ways.

raphjaph: I'm leaning towards yes, I'll let @casey chime in if he sees this

src/index/testing.rs
```diff
@@ -128,7 +128,20 @@ impl Context {
...
     balances.sort_by_key(|(outpoint, _)| *outpoint);
 
-    for (_, balances) in balances.iter_mut() {
+    let mut burned_balances = 0;
+
+    for (outpoint, balances) in balances.iter_mut() {
+      if self
+        .index
+        .get_output_script(outpoint)
+        .ok()
```
casey, line 137: We should probably unwrap and panic here.

src/index/testing.rs
```diff
@@ -146,7 +159,7 @@ impl Context {
 
     for (id, entry) in runes {
       pretty_assert_eq!(
-        outstanding.get(id).copied().unwrap_or_default(),
+        outstanding.get(id).copied().unwrap_or_default() - burned_balances,
```
casey, line 162: Add assert above which asserts that burned_balances is correct, similar to the above asserts.

src/index/testing.rs
```diff
@@ -128,7 +128,20 @@ impl Context {
     let balances = balances.as_mut();
     balances.sort_by_key(|(outpoint, _)| *outpoint);
 
-    for (_, balances) in balances.iter_mut() {
+    let mut burned_balances = 0;
```
casey, line 131: I would just call this burned. It's not really "balances" because it's just a number. Balances so named because it's a map.

src/index.rs
```diff
@@ -499,6 +499,27 @@ impl Index {
     )
   }
 
+  #[cfg(test)]
+  pub fn get_output_script(&self, output: &OutPoint) -> Result<ScriptBuf> {
```
casey, line 503: Should this return an error? Test functions should just panic and not return errors, since an error can be ignored and a panic cannot.

src/index/testing.rs
```diff
@@ -128,7 +128,20 @@ impl Context {
...
+
+    for (outpoint, balances) in balances.iter_mut() {
+      if self
+        .index
+        .get_output_script(outpoint)
+        .ok()
+        .is_some_and(|script| script.is_op_return())
```
casey, line 138: This should also unwrap. If we expect the output script to be in the index, it is a fatal error for it not to be, and we want to panic so we notice.

src/index/updater/rune_updater.rs
```diff
@@ -198,14 +197,16 @@ impl RuneUpdater<'_, '_, '_> {
...
 
-        if let Some(sender) = self.event_sender {
-          sender.blocking_send(Event::RuneTransferred {
-            outpoint,
-            block_height: self.height,
-            txid,
-            rune_id: id,
-            amount: balance.0,
-          })?;
+        if !tx.output[vout].script_pubkey.is_op_return() {
```
casey, line 200: This reduces the amount of information a listener gets. This is a separate change and should be done in a separate PR.

src/index.rs
```diff
@@ -499,6 +499,27 @@ impl Index {
...
+    Ok(if self.index_addresses {
+      let rtx = self.database.begin_read()?;
+      let outpoint_to_utxo_entry = rtx.open_table(OUTPOINT_TO_UTXO_ENTRY)?;
+      let entry = outpoint_to_utxo_entry
+        .get(&output.store())?
+        .ok_or_else(|| anyhow!("UTXO entry not found for output: {:?}", output))?;
```
casey, line 509: Should these panic?

src/index.rs
```diff
@@ -499,6 +499,27 @@ impl Index {
...
+
+      let parsed = entry.value().parse(self);
+
+      ScriptBuf::from_bytes(parsed.script_pubkey().to_vec())
+    } else {
+      self
+        .get_output_info(*output)?
+        .ok_or_else(|| anyhow!("output {output} not found"))?
```
casey, line 517: Should these panic?

casey (changes requested):
Always panic on unexpected errors in tests, or when we're looking something up in the index that we expect to be there. This kind of defensive programming can catch a lot of bugs.

We should definitely not change the behavior of the event sender. This is a conceptually separate change, even though it's related to OP_RETURNs. If the client is relying on getting this information, than it could be a tricky bug if they don't.

raphjaph:
@lifofifoX any thoughts on the review comments?

PR #4306, opened 2025-03-27
Add search index

terror:


src/options.rs
```diff
@@ -73,6 +73,8 @@ pub struct Options {
     help = "Do not index inscriptions."
   )]
   pub(crate) no_index_inscriptions: bool,
+  #[arg(long, help = "Use search index at <SEARCH_INDEX>.")]
+  pub(crate) search_index: Option<PathBuf>,
```
casey, line 77: I think this doesn't even need to be configurable initially. We could just put it into the ord data dir.

src/lib.rs
```diff
@@ -279,6 +282,14 @@ fn gracefully_shut_down_indexer() {
       log::warn!("Index thread panicked; join failed");
     }
   }
+
+  if let Some(search_indexer) = SEARCH_INDEXER.lock().unwrap().take() {
+    shut_down();
+    log::info!("Waiting for search index thread to finish...");
+    if search_indexer.join().is_err() {
+      log::warn!("Search index thread panicked; join failed");
+    }
```
casey, line 291: Just good for debuggability:

```suggestion
    if let Err(err) = search_indexer.join() {
      log::warn!("Search index thread panicked: {err}");
    }
```

casey (changes requested):

PR #4308, opened 2025-03-29
Fill iframe when displaying images under /thumbnail

### LINKED ISSUE #4296: Fill iframe when displaying images under /thumbnail
We now have different routes for previews, `/preview/ID`, shown as the large preview on /inscription, and thumbnails, `/thumbnail/ID`, shown everywhere else.

We should change image previews (`preview-image.{html,js}`) to fill the document when served under /thumbnail, so there isn't any wasted space.

qvkare:
Modifies the preview-image.js file to detect when images are being displayed under the /thumbnail route and applies background-size: cover to fill the entire iframe space, eliminating wasted space.

Fixes #4296

- Added URL path detection to determine if we're in the /thumbnail route
- Applied background-size: cover when in thumbnail mode to fill the entire iframe
- Maintained original behavior for /preview route

Tested with different image sizes to ensure proper display in both routes.

### DIFF
--- a/static/preview-image.js
+++ b/static/preview-image.js
@@ -1,15 +1,29 @@
+function isThumbnailRoute() {
+  return window.location.pathname.includes('/thumbnail');
+}
+
 function resize() {
   if (
     body.clientWidth * window.devicePixelRatio < img.naturalWidth
     || body.clientHeight * window.devicePixelRatio < img.naturalHeight
   ) {
     body.style.imageRendering = 'auto';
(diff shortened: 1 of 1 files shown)


casey:
Thanks for the PR! I think JavaScript is the wrong approach here. When instantiating the template, the server already knows whether or not it's a thumbnail, so we can just do it server-side, which is faster and doesn't break if JS isn't available. See #4297.

PR #4323, opened 2025-04-07
Don't display `(expected)` when rune has already unlocked

casey:
We reuse the `Blocktime` struct which already does this, instead of only using the timestamp.


raphjaph (approved):
LGTM

PR #4324, opened 2025-04-08
Update sparrow-wallet.md

owenzimmew06:
Hello,

I identify a typo and made suggestions for improvement in this text.

1. "There is a few of important things to check here" (Sending your inscription section):

"There is" is incorrect because "few" is plural.

Suggested revision: "There is" should be "There are".

2. "UXTO" is a typo and should be "UTXO."

Suggested revision: "UXTO" should be "UTXO".

3. "there is a few things you need to check" (Sending your inscription section):

Same as the previous point, "there is" should be "there are."

Suggested revision: "there is a few things" should be "there are a few things".

Thanks.


raphjaph (approved):
LGTM

PR #4326, opened 2025-04-10
Allow inscribing JPEG XL

casey:
JPEG XL is [not widely supported](https://caniuse.com/?search=jpegxl).

However, you can take advantage of the format to create extremely small but visually compelling images, which is perfect for inscribing. [example](https://jxl-art.surma.technology/).

So, we could consider adding the `jxl` extension and allow `ord` to inscribe `jxl` without issue. This would be a different policy than usual, since if people don't know that JPEG XL isn't widely available, they might inscribe it and be disappointed when it only appears correctly in Safari.


raphjaph (approved):
LFG

PR #4329, opened 2025-04-20
Rpc client retries when "Resource temporarily unavailable"

SmarakNayak:
The bitcoin RPC client would sometimes fail to connect with a "Resource temporarily unavailable" error.

Retrying the connection on this error solved the issue for me.


raphjaph (approved):
Seems fine to me

PR #4351, opened 2025-05-19
Update Dockerfile rust image

actuallycabbage:
### Summary
Bumps the Dockerfile's builder image from `rust:1.79.0-bookworm` to `rust:1.81.0-bookworm`.

### Context
The project no longer builds at 1.79 and so this PR bumps it to the minimum `rustc` version at which the project builds again.

### Note
I am not very well-versed in rust dependency chains or module resolution, but `Cargo.toml` currently sets `rust-version = "1.80.0"` which appears to be too low as well. I'm leaving it untouched for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.79.0-bookworm as builder
+FROM rust:1.81.0-bookworm as builder
 
 WORKDIR /usr/src/ord
 


raphjaph:
Thanks! 

So it doesn't compile with Rust version 1.80.0?

actuallycabbage:
No, it does not.

```
$ rustc --version
rustc 1.80.0 (051478957 2024-07-21)
```

```
$ cargo build
error: rustc 1.80.0 is not supported by the following packages:
  half@2.5.0 requires rustc 1.81
  litemap@0.7.5 requires rustc 1.81
  redb@2.4.0 requires rustc 1.81
  redb@2.4.0 requires rustc 1.81
  redb@2.4.0 requires rustc 1.81
  zerofrom@0.1.6 requires rustc 1.81
```

raphjaph:
Ok good catch, we've updated the minimum rust version in the `Cargo.toml`.

raphjaph (approved):

PR #4356, opened 2025-05-20
Update minimum Rust version to 1.85.0

blieb:
rust version 1.85.0 minimum is required for this build. Upgrading to latest stable release (1.87.0). Fixing the error below

```
3.806 error: failed to parse manifest at `/usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/redb-2.5.0/Cargo.toml`
3.806 
3.806 Caused by:
3.806   feature `edition2024` is required
3.806 
3.806   The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.81.0 (2dbb1af80 2024-08-20)).
3.806   Consider trying a newer version of Cargo (this may require the nightly release).
3.806   See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ edition = "2021"
 homepage = "https://github.com/ordinals/ord"
 license = "CC0-1.0"
 repository = "https://github.com/ordinals/ord"
-rust-version = "1.81.0"
+rust-version = "1.85.0"
 
 [workspace.dependencies]
 base64 = "0.22.0"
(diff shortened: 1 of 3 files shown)


raphjaph (approved):
Thanks!

I've updated it to the minimum required version (1.85.0) and tested that it is in fact the minimum.

raphjaph:
I'm fixing the lint failures in #4373

PR #4366, opened 2025-06-06
fix spelling issues CHANGELOG.md

davidjsonn:
Hey team! Fixed error in CHANGELOG.md

`consistant` - `consistent`
`uneccessary` - `unnecessary`
`Vaccuum` - `Vacuum`
`intructions` - `instructions`


raphjaph (approved):

raphjaph:
Thanks for finding these but on second thought we should not change this since it's part of the CHANGELOG and those are generated from the commit messages. It should be consistent with the commit history.

PR #4370, opened 2025-06-21
[CHANGELOG.md]: fix misspelling of "receive"

kks-code:


raphjaph:
Thanks but we should not change this since it's part of the CHANGELOG and those are generated from the commit messages. It should be consistent with the commit history.

PR #4371, opened 2025-06-22
Allow inscribing with backup and Core v29

### LINKED ISSUE #4335: Can't inscribe with core v29.0 without --no-backup
A bug introduced in Bitcoin core v29.0 prevents ord from being able to import the recovery key into its core wallet:

```
$ ord inscribe --file test.txt --fee-rate 1
error: commit tx recovery key import failed
```

https://github.com/bitcoin/bitcoin/issues/32376

Possible fixes:
* don't set internal:false when importing the key (because that's the default anyway, and the bug only happens if you explicitly set it)
* don't set a label on the recovery key descriptor

gmart7t2:
Fix #4335

A bug in bitcoin core v29 means we can't add a descriptor for the backup key if we specify "internal:false" and also provide a label.

Using None for internal uses the default (false) which does work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ edition = "2021"
 homepage = "https://github.com/ordinals/ord"
 license = "CC0-1.0"
 repository = "https://github.com/ordinals/ord"
-rust-version = "1.81.0"
+rust-version = "1.85.0"
 
 [workspace.dependencies]
 base64 = "0.22.0"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.81.0-bookworm as builder
+FROM rust:1.85.0-bookworm AS builder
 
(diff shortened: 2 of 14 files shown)


raphjaph:
This seems fine to me since the default is false: https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/importdescriptors

I'm also currently syncing my node and building the index to check that this works.

@casey any thoughts?

casey:
This fix seems fine to me. I can confirm that [the docs](https://developer.bitcoin.org/reference/rpc/importdescriptors.html) show `false` as being the default if `internal` is omitted, so this shouldn't be a change in behavior.

gmart7t2:
It shouldn't be a change in behavior, but it is because of the Bitcoin core bug

raphjaph:
Ok I've tested it and it works!

PR #4388, opened 2025-07-30
docs: fix broken link

MozirDmitriy:
Replaced broken link http://storage.googleapis.com/books/ngrams/books/datasetsv2.html
with correct one https://storage.googleapis.com/books/ngrams/books/datasetsv3.html


casey:
It's unfortunate, but I think we should just leave the broken link. I checked, and there used to be a v2 dataset which is no longer available. We could just change the link to point to the v3 dataset, which is still available, but that might lead to a different frequency.tsv file.

PR #4399, opened 2025-08-26
[Work In Progress] Track custom satoshis

dannydeezy:
Introduces the `--track-sats` command line option which enables a user to add a file with a list of custom satoshi numbers. Ord will then track and index the locations for all sats in this file. This should be useful for rare satoshi enthusiasts who like to track locations of satoshi categories beyond the default uncommon, rare, epic, etc.

Note the intention is not to determine which satoshis all users should track, but instead to give the users flexibility to track whichever sats they care about. The expectation is that market participates will externally align on which satoshis are worth tracking, and perhaps develop and share standard tracking files, but all this will be independent of ord

Potentially further todos:
- We could allow the file to be CSV where the second column is "category" which would let the user assign each of the custom satoshis to a category (such as "2-digit palindrome", "silkroad", etc)
- We could add a way to track sat ranges rather than just individual sats


dannydeezy:
Bear with me - all this was written by AI and I haven't given it a proper human review yet 😆

PR #4412, opened 2025-09-24
Update to Rust 2024

casey:
This PR consists of three commits:

1. Update `Cargo.toml` to Rust 2024
2. Updating formatting with `cargo fmt`
3. Manually collapse if-let chains


raphjaph (approved):
Nice, LGTM!

PR #4421, opened 2025-10-19
64 digit numbers are txids not inscription numbers or block heights

gmart7t2:
I made an inscription with no hex letters in its id: `4273262611454246626680278280877079635139930168289368354696278617i0`

* If I go to `https://ordinals.com/4273262611454246626680278280877079635139930168289368354696278617` it redirects to `/inscription/` because it thinks digit strings are inscription numbers. It fails with `"Invalid URL: number too large to fit in target type"`.

  It should redirect to `/tx/` if it's a 64 digit number.

* And if I go to `https://ordinals.com/4273262611454246626680278280877079635139930168289368354696278617:0` it redirects to `/rune/` because it thinks pairs of digit strings with a colon between are rune identifiers. It fails with `"Invalid URL: invalid height: number too large to fit in target type"`.

  It should redirect to `/output/` if the first digit string is a 64 digit number.

### DIFF
--- a/docs/src/guides/api.md
+++ b/docs/src/guides/api.md
@@ -1354,10 +1354,24 @@ curl -s -H "Accept: application/json" /
   "fee": 322,
   "height": 767430,
   "id": "6fb976ab49dcec017f1e201e84395983204ae1a7c2abf7ced0a85d692e442799i0",
+  "metaprotocol": "LOL-420",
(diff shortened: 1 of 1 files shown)


gmart7t2:
> Nice catch! These should have tests, they're very simple to write. See the `fallback` test in `server.rs`.

I added a test for each case

raphjaph (approved):
LGTM

PR #4423, opened 2025-10-27
Use single properties struct

casey:
The original properties PR used two structs, a `Properties` struct, which is a nice struct which enforces some invariants, and a `RawProperties` struct which is deserialized from CBOR.

After working on attributes, now think that was a mistake. As we add more structs, for example one for attributes, we then need to duplicate each and have a raw version and a nice version. This PR removes the `Properties` struct, renames `RawProperties` to `Properties`, and uses it everywhere.

### DIFF
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -4,6 +4,7 @@ use {
 };
 
 #[derive(Debug, Decode, Default, Encode, PartialEq)]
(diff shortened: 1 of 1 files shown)


raphjaph (approved):
LGTM

PR #4430, opened 2025-11-01
3 fixes to runes specification:

gmart7t2:
1. pointer can't be greater than or equal to number of outputs (not strict inequality)
2. edict output number can't be greater than number of outputs (not inputs)
3. bad pointer is a possible cause of cenotaphs (for completeness)


casey:
Nice, these are great, thank you!

PR #4439, opened 2025-11-14
Don't error if block inscription page request is out of range

casey:
Previously, requesting `/inscriptions/block/{height}/{page}` with pages over the maximum would error with an `anyhow!` error.

This is incorrect. `anyhow!` errors are interpreted as internal errors, i.e., errors encountered when processing valid user input. Invalid user input such as requests for pages that don't exist, should generate `ServerError::NotFound`, and return a 404 to the user.

I noticed this because the server logs were getting spammed with `error serving request: page index X exceeds inscription count`. An important distinction between errors from invalid user input and internal errors is that invalid user input errors don't result in log messages.

I fixed the bug by removing the check for an out-of-bound page and just returning a template with no inscriptions. This did not cause any test failures, so generating this error was untested. If a test had been written, this probably wouldn't have happened, since the request would generate a 500 INTERNAL SERVER ERROR, which is obviously incorrect.

There were also a bunch of weird, extraneous conversions which used `usize::try_from` to convert from a u32 to a usize. `ord` does not support 16bit computers or 8bit microcontrollers, so you can always assume that a u32 can be converted to a usize. See the `IntoUsize` trait, which provides an unconditional conversion from a `u32` to a `usize`.

I added a test that requesting an out-of-bounds page returns an empty template.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -304,6 +304,8 @@ impl Index {
       Err(DatabaseError::Storage(StorageError::Io(error)))
         if error.kind() == io::ErrorKind::NotFound =>
       {
+        log::info!("Creating new index");
+
         let database = Database::builder()
           .set_cache_size(index_cache_size)
           .create(&path)?;


raphjaph (approved):
LGTM

Thanks for fixing!

PR #4440, opened 2025-11-16
Fix trait template description list

casey:


casey:
@raphjaph I just went ahead and merged this, I want to cut a release and sync that release on charlie. Do take a look though.

PR #4444, opened 2025-11-24
chore(ci): upgrade checkout to v6

rejected-l:
Updates `actions/checkout` to v6 in CI workflows. No code changes.

https://github.com/actions/checkout/releases/tag/v6.0.0


casey:
Unless there's a reason to upgrade, we can just stay on the old version.

PR #4446, opened 2025-11-27
Remove Runes

portlandhodl:
As a thanksgiving gift to all I have decided to release bag holders from their eternal prison of number go down. I have read at least 2^128 Twitter posts personally about this topic. As such I am what you would call "An expert" in this number go down technology that was introduced into Ord just over a year and half ago. 

As a moral obligation to use my powers of manipulating time an space, Runes must be sent to the Shadow realm not to be rescued to Yugi and his friends.

Best, 

- Maximillion Pegasus
- President Industrial Illusions


0xbrito:
Cringe

cbspears:
Not cringe 😎

PR #4448, opened 2025-12-03
Update to boilerplate 1.1.0

casey:


casey:
@raphjaph This is failing because the mdbook theme, which we manually override because of translations, is out of date with the current version used by mdbook. Could you take a look? I honestly don't understand any of the changes we made to the theme 😅 It'll prevent CI from passing on any PR until we fix.

PR #4449, opened 2025-12-11
Use title instead of inscription number when present

lifofifoX:
Titles look much nicer than larger inscription numbers.

image

### DIFF
--- a/templates/inscription.html
+++ b/templates/inscription.html
@@ -1,5 +1,5 @@
 <h1>Inscription {{ self.number }}</h1>
-%% if let Some(title) = self.properties.attributes.title.as_ref() {
+%% if let Some(title) = &self.properties.attributes.title {
 <p class=subtitle>{{ title }}</p>
 %% }
 <div class=inscription>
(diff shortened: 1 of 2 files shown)


cbspears:
Big fan of this idea. Maybe consider moving the inscription # to a subheading underneath & slightly darker grey just for clarity.

lifofifoX:
@cbspears Good call. Updated it.

image

casey (changes requested):
@lifofifoX I'm very much in favor of displaying the title, if present, with the inscription number.

However, I'm not sure about the inconsistency of some inscriptions having inscription numbers at the top, and others having titles at the top. For example, what if someone creates an inscription with the title `Inscription 0`. This could be deceptive.

How about we add the title under the inscription number, if present. So inscription number still always gets the `<h1>`, and title gets `<p>`, and later we can discuss swapping them, or otherwise changing the emphasis?

The reason the title should get `<p>` instead of `<hN>` is because it doesn't contribute to the structure of the document, it's just a sub-title. We can use CSS to style the `<p>` however though.

lifofifoX:
@casey That seems pretty reasonable. Made the changes and attached screenshots showing the changes.

Inscription
image

Gallery
image

PR #4451, opened 2026-01-04
fix: set internal=false on recovery key descriptor import

8144225309:
## Problem

Recovery key descriptors are imported with `internal: None`, which defaults to `true`. Bitcoin Core hides internal descriptors from `listunspent`, making stuck commit UTXOs invisible.

## Fix

Set `internal: Some(false)` so the descriptor is treated as a receive address and appears in `listunspent`.

## Testing

Tested on Bitcoin Core 28.0 regtest - recovery UTXOs now appear in `listunspent`.

## Reference

bitcoin/bitcoin#26813

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1386,12 +1386,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "executable-path"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ebc5a6d89e3c90b84e8f33c8737933dda8f1c106b5415900b38b9d433841478"
-
-[[package]]
 name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2675,7 +2669,6 @@ dependencies = [
  "derive_more",
  "dirs",
  "env_logger",
- "executable-path",
  "futures",
  "hex",
  "http-body-util",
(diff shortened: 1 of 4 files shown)


casey:
Thanks for the PR! This seems like a harmless change, but I don't understand Bitcoin Core's behavior at all. Why are internal descriptors hidden? Shouldn't `listunspent` display everything, including change?

8144225309:
I encountered this bug making non-standard txs: successful commits but failed reveals. This locked up funds for me
twice, taking a week each time to diagnose and recover. Thanks to the discord for the manual workaround (dump
descriptors → load into Sparrow). This fix might need supplemental code to handle pre-existing untracked UTXOs.

Re: Core behavior - achow101 explains in bitcoin/bitcoin#26813 that descriptor wallets don't create address book
entries for descriptors imported without labels. Without an entry, Core assumes it's change and hides it from
listunspent. This footgun got worse after Core 23.0 (April 2022) made descriptor wallets the default.

The recovery key is imported with active: false, which may interact with the missing internal flag in ways that
trigger change detection. Explicitly setting internal: false fixes it - the exact mechanism is murky, but the fix
works.

Tldr: Core behavior is weird (achow herself comments: https://github.com/bitcoin/bitcoin/issues/26813#issuecomment-1372456440), this fixes it for ord.

casey:
Thanks for the explanation. Unfortunate that we have to work around a bitcoin core issue, but it is what it is. Sorry you ran into this, and thanks for the fix.

PR #4452, opened 2026-01-04
Refresh Bitcoin Core RPC client on authentication failure

RobertClarke:
When Bitcoin Core restarts, it regenerates the `.cookie` file with new credentials. Previously, `ord` would fail with 401 errors until restarted.

This is particularly useful when running `ord server` long-term while occasionally restarting Bitcoin Core for configuration changes (peer adjustments, network tuning, etc.). With this change, `ord` automatically recovers without manual intervention.

Changes:
- Wrap `Index.client` in `RwLock` to allow runtime credential refresh
- Add `with_retry()` wrapper that detects 401 errors and re-reads the cookie file
- Apply retry logic to all Bitcoin Core RPC calls
- Add similar retry logic to the async `Fetcher` used during indexing

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -35,7 +35,7 @@ use {
   std::{
     collections::HashMap,
     io::{BufWriter, Write},
-    sync::Once,
+    sync::{Once, RwLock},
   },
 };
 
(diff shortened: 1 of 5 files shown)


casey:

PR #4459, opened 2026-02-02
Simplify minicbor property decoding

casey:
Remove raw properties structs using the new `skip_if` attribute.


casey:
@raphjaph This is pretty straightfoward, although there is some code movement, if you want to take a look.

PR #4494, opened 2026-03-08
Return error when exporting addresses without address index

### LINKED ISSUE #3917: Don't export addresses without address index
When doing `ord index export`, we currently export addresses even if the index wasn't built with `--index-addresses`. This is prohibitively slow. If the address index doesn't exist, we should just not export addresses.

Bortlesboat:
When `--include-addresses` is used but the index was not built with
`--index-addresses`, export falls back to fetching every transaction
via RPC, which is prohibitively slow.

Return a clear error instead and remove the RPC fallback path.

Fixes #3917

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -707,6 +707,10 @@ impl Index {
   }
 
   pub fn export(&self, filename: &String, include_addresses: bool) -> Result {
+    if include_addresses && !self.index_addresses {
+      bail!("--include-addresses requires index built with --index-addresses");
+    }
+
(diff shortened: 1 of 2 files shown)


Bortlesboat:
Rebased onto current master and added an integration regression for the missing address-index error. The focused test passes, as do rustfmt and full all-target/all-feature Clippy with warnings denied.

casey:
I'm not sure this should be forbidden. If it's slow, then perhaps just a warning is fine.

PR #4496, opened 2026-03-10
ADD /r/block/{query}/tx for block tx recursion

dave4506:
Adds support for `/r/block/{query}/tx` for recursive query of the transactions contained in a block.

Open to feedback/change on the endpoint and how the data should be returned.

Adds a requested endpoint discussed here:
https://github.com/ordinals/ord/discussions/4480

### DIFF
--- a/docs/src/inscriptions/recursion.md
+++ b/docs/src/inscriptions/recursion.md
@@ -225,6 +225,37 @@ curl -s \
 <details>
   <summary>
     <code>GET</code>
+    <code><b>/r/block/&lt;HEIGHT_OR_BLOCK_HASH&gt;/tx</b></code>
+  </summary>
+
+### Description
+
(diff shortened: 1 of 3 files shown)


Blockamoto:
Do you think it would be better to simply have /r/block/<height/hash> to return block hex? Then everything else can be resolved client-side.

Because every time you need to run a tx query, you'd need to pull the block hex if you did it this way. Is that correct?

And doing it this way would waste a whole load of unneccesary calls, if user then wanted to fetch the transaction data.

In my case, I want to know the sum of all outputs, so I would need to run /r/tx on each of the thousands of ID's returned, and a script on each, which is nuts considering we already pulled the entire block hex data. If I simply had /r/block to return the block hex, I could feasible do this with just that one call and a client-side script, saving thousands of pings.

PR #4537, opened 2026-06-18
offer: pass `input_weights` so `wallet offer create` funds on Bitcoin Core v24+

hans-crypto:
## Problem

`ord wallet offer create` fails on Bitcoin Core v24+ with:

```
error: JSON-RPC error: RPC error response: RpcError { code: -4,
  message: "Not solvable pre-selected input COutPoint(<seller>, 0)", data: None }
```

`offer create` builds a buyer-initiated PSBT whose input 0 is the seller's
inscription UTXO, pre-selected and left unsigned (the seller signs it on
`offer accept`). That input is at the seller's address, which the buyer's
wallet has no descriptor for. Core v24+ rejects pre-selected inputs it cannot
solve in `fundrawtransaction` (so fee estimation stays accurate) unless the
caller supplies their weight via `input_weights`; `offer create` passes `None`,
so funding aborts. This hits any offer for an inscription the wallet doesn't
hold — the normal buyer-initiated case.

## Regression

Offers (#4156) merged 2025-01-11, when the minimum Core was pre-v25 and
`fundrawtransaction` did not enforce pre-selected-input solvability. The v28
minimum (#4185) landed six days later; the cross-wallet offer path was never
re-tested against it.

## Fix

Supply `input_weights` for the seller's input, estimating its weight the same
way `wallet sweep` already does for its foreign pre-selected inputs.

## Test coverage

`mockcore` funded any pre-selected input regardless of solvability, so the bug
was invisible. This PR teaches it the Core v24+ rule: a pre-selected input that
is neither wallet-owned nor declared in `input_weights` is rejected with the
same error. With that, `wallet::offer::create::created_offer_is_correct` and
`offers_can_be_submitted` fail with the exact Core error without the
`offer/create.rs` change and pass with it.


hans-crypto:
I run this exact patch in a downstream fork ([cat21-ord](https://github.com/ordpool-space/cat21-ord)) under CI end-to-end against **real Bitcoin Core 28**, not just `mockcore`. The fork commit is the same 3-file change as this PR: [ordpool-space/cat21-ord@86e4ac5d](https://github.com/ordpool-space/cat21-ord/commit/86e4ac5d).

In the run below, a regtest stack (bitcoind 28 + electrs + the patched ord) drives `ord wallet offer create` for an inscription at an address the buyer's wallet has no descriptor for — the cross-wallet case that hits the v24+ solvability check. With the patch the call funds and returns a valid PSBT, and an independent PSBT builder produces a byte-identical offer over the same seller UTXO (`0xfffffffd` sequence, input unsigned, 546-sat postage output, `amount + postage` to the seller). Without `input_weights` it fails with `Not solvable pre-selected input COutPoint(<seller>, 0)`.

Green CI run: https://github.com/ordpool-space/ordpool-sdk/actions/runs/27783820403

Issue #4435, opened 2025-11-13
Bump database schema for 0.24.0


gmart7t2:
Redb v3 .redb files are incompatible with redb v2 .redb files, so ord 0.24.0 doesn't work with index.redb from older versions

casey:
Thanks for reporting this! Fixed in #4436. I'm going to cut a new release with the schema bump.

PR #4405, opened 2025-09-10
Add /galleries page

SatoshiRoppongi:
Implementation for this issue
https://github.com/ordinals/ord/issues/4320

Issue #3623, opened 2024-04-19
index apears to have been built with older, incompatible version of ord, consider deleting and rebuilding


Horianu:
index "C:\users\AppData\roaming\ord\index.redb" apears to have been built with older, incompatible version of ord, consider deleting and rebuilding the index: index schema 24, ord schema 25

This is what I get when updating from 0.17.1 to any version of 0.18 , have asked and other didn't have that problem at all.

Any idea why?

cryptoni9n:
Hi Horianu - this is by design.  Anytime ord has a major version update, it requires a new index schema and everyone must then reindex.

Additionally - questions like this are better asked on the official ord discord, called the Ordicord here https://discord.gg/ordinals

raphjaph:
https://docs.ordinals.com/guides/reindexing.html

PR #4546, opened 2026-07-09
DRAFT: Add index schema migrations and `ord index upgrade`

dannydeezy:
Every index schema bump currently requires deleting `index.redb` and reindexing from genesis — days of downtime for a mainnet index with `--index-sats`. That's happened four times in the last four schema-bumping releases (0.24.1, 0.25.0, 0.26.0, 0.27.0 — schemas 31 through 34), and users without archived indexes are told to rebuild from scratch (#3623, #4489).

None of those four bumps actually needed a full reindex:

| Schema | Change | Upgrade path |
|--------|--------|--------------|
| 30 → 31 | None — bump marked redb's automatic v2 → v3 file-format upgrade (#4435, #4436) | No-op |
| 31 → 32 | Added `GALLERIES` table (#4405) | Backfill by re-parsing inscription envelopes from reveal transactions, fetched from bitcoin core one block at a time |
| 32 → 33 | Renamed `GALLERIES` → `GALLERY_SEQUENCE_NUMBERS`, added latest-child collection tables (#4475) | Rename the table; derive latest children from `SEQUENCE_NUMBER_TO_CHILDREN`, already in the index |
| 33 → 34 | Added `hidden` flag to inscription entries, excluded hidden inscriptions from galleries and collections (#4488) | Recompute `hidden` from envelopes (same sweep as 31 → 32) and rewrite the entry table via redb's `rename_table` |

This PR adds an index schema migration registry and an `ord index upgrade` subcommand, with migrations for schemas 30 through 34, so an index built by 0.22.x–0.24.0 upgrades straight to current without reindexing:

- Each migration upgrades schema `N` to `N + 1`, producing the same contents a fresh reindex with the new code would produce for the affected tables.
- Each migration runs in a single write transaction that also updates the schema statistic, so upgrades are atomic and resumable at migration granularity.
- Opening an old index whose schema has a complete migration chain errors with "run `ord index upgrade` to upgrade it in place" instead of "consider deleting and rebuilding".
- Schema bumps **without** a registered migration keep the current delete-and-rebuild error — this makes migrations an option per bump, not an obligation. A future bump that changes sat tracking or inscription numbering can still simply require a reindex.

The envelope-sweep migrations skip everything expensive about reindexing (sat ranges, UTXO tracking, rune updates): they fetch each block containing inscriptions once and rewrite only the affected tables — hours on mainnet instead of days, and the 32 → 33 migration is minutes.

Tests include a round trip: build a current-schema index containing hidden, visible, gallery, and parent-child inscriptions, downgrade it in place to the schema-30 format (legacy entry tuple type, derived tables dropped), run the upgrade, and assert the result is identical to the original, table by table.

Two costs worth noting: the 33 → 34 entry-table rewrite holds old and new tables in the file until commit, so the index file grows transiently (recoverable with compaction afterwards), and the envelope sweeps are bounded by bitcoind RPC throughput.

Opening as a draft to get feedback on the direction before polishing — happy to adjust the design, e.g. automatic migration on open instead of an explicit subcommand, or keeping only the cheap in-database migrations.

🤖 And yes this was generated with claude, but its Fable so maybe its not slop!!!!

### DIFF
--- a/docs/src/guides/reindexing.md
+++ b/docs/src/guides/reindexing.md
@@ -8,6 +8,11 @@ database and restarting the indexing process with either `ord index update` or
 1. A new major release of ord, which changes the database scheme
 2. The database got corrupted somehow
 
+Some database schema changes can be applied to an existing index without
+reindexing, which is usually much faster than reindexing. If this is the case,
+`ord` will print an error message saying so when opening the index, and running
+`ord index upgrade` will upgrade the index in place.
+
 The database `ord` uses is called [redb](https://github.com/cberner/redb),
 so we give the index the default file name `index.redb`. By default we store this
 file in different locations depending on your operating system.
--- a/src/index.rs
+++ b/src/index.rs
@@ -45,6 +45,7 @@ pub(crate) mod entry;
 pub mod event;
 mod fetcher;
 mod lot;
+pub(crate) mod migrations;
 mod reorg;
 mod rtx;
 mod updater;
@@ -284,6 +285,10 @@ impl Index {
             .unwrap_or(0);
 
           match schema_version.cmp(&SCHEMA_VERSION) {
+            cmp::Ordering::Less if migrations::upgradable(schema_version) => bail!(
+              "index at `{}` was built with an older version of ord, run `ord index upgrade` to upgrade it in place, without reindexing: index schema {schema_version}, ord schema {SCHEMA_VERSION}",
+              path.display()
+            ),
             cmp::Ordering::Less => bail!(
               "index at `{}` appears to have been built with an older, incompatible version of ord, consider deleting and rebuilding the index: index schema {schema_version}, ord schema {SCHEMA_VERSION}",
               path.display()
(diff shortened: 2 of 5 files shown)


casey:
I'm not sure there's much value in adding migrations for older releases. This is something we could consider going forward, however I'd still be nervous about it. Migration code duplicates the code of the schema-change itself, and adds multiple states that ord could be in, i.e., "schema N" and "schema N migrated from M", which could be a source of bugs and complexity.

dannydeezy:
> I'm not sure there's much value in adding migrations for older releases. This is something we could consider going forward, however I'd still be nervous about it. Migration code duplicates the code of the schema-change itself, and adds multiple states that ord could be in, i.e., "schema N" and "schema N migrated from M", which could be a source of bugs and complexity.

fair point. in future, would you ever program the migration run into the indexing process such that all ords on the newer (even ones indexing from scratch) perform the same migration at the same block height, such that all are "schema N migrated from M at block height H"?. that might be dumb and gross though, idk

casey:
It's very hard to say, I think it depends on the migration, and what the code looked like.

PR #4553, opened 2026-08-03
Forbid offers from specifying input sighashes

casey:
`bitcoind` checks and forbids non-default sighash flags on unsigned inputs, so this is impossible to exploit, but we shouldn't rely on that. Found and flagged by the inimitable @rob1ham 🙏


Rob1Ham:
🫡

perricrruz109090-cloud (approved):

PR #4558, opened 2026-08-08
Warn when exporting addresses without address index

Bortlesboat:
Warn instead of erroring, per your comment on #4494.

`eprintln!` and not `log::warn!` because `env_logger::init()` is called
without a default filter, so warn records don't print unless `RUST_LOG`
is set.

Didn't mark #3917 fixed since it asks to skip addresses entirely, which
isn't what this does.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -707,6 +707,13 @@ impl Index {
   }
 
   pub fn export(&self, filename: &String, include_addresses: bool) -> Result {
+    if include_addresses && !self.index_addresses {
+      eprintln!(
+        "warning: exporting addresses without `--index-addresses` requires fetching every \
+         transaction over RPC, which is slow"
+      );
+    }
+
     let mut writer = BufWriter::new(File::create(filename)?);
     let rtx = self.database.begin_read()?;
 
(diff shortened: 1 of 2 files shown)


casey (changes requested):
Is `log::warn` enabled by default? If so, we should use it instead of `eprintln`. Also, I don't think this needs a test.

Bortlesboat:
No, it isn't. `main` calls bare `env_logger::init()`, so with `RUST_LOG` unset
the filter is `error` and `log::warn!` records are dropped.

Checked it rather than guessing — I swapped the `eprintln!` for `log::warn!` and
ran the export through the test harness, which asserts on the subprocess's
stderr:

- `log::warn!`, no `RUST_LOG`: stderr empty, assertion fails
- `log::warn!`, `RUST_LOG=warn`: warning present, passes
- `eprintln!`: present either way

So `log::warn!` would silence the warning for anyone not already running with
`RUST_LOG` set, which is most people hitting this. `eprintln!` also matches what
`wallet balance` does for its `warning:` line.

Dropped the test — the diff is now just the seven lines in `src/index.rs`.

casey (approved):
LGTM, thanks for checking!

Bortlesboat:
The red Windows job is unrelated: `wallet::send::splitting_merged_inscriptions_is_possible`
failed with "`ord server` 1 blocks behind `bitcoind`". That test never touches
`index export`, and it passes for me on Windows on a00f1d5. Looks like a runner
timing flake — worth a rerun.

